PR #39, opened 2017-01-17
zlib archive URL has moved again!

willmclaren:
I propose you always link to the fossils dir...


chrchang:
gah... ok, will make this fall back on fossils dir if main DL fails.

PR #104, opened 2019-03-19
Fix typo in byte notation

adriaan-vd-graaf:
Hope this is a correct fix.


adriaan-vd-graaf:
Sorry, my byte notation is rusty.

PR #129, opened 2019-10-30
Fix a bug in RcppExports.R

yk-tanigawa:
`ReadList` and `ReadIntList` should take a buffer (`buf`) and pass it to the corresponding C++ function.


chrchang:
Not correct for the current ReadList/ReadIntList implementation.  (There may be an optional preallocated buffer parameter later, but it should be much less important than when processing a single variant at a time.)

PR #146, opened 2020-06-20
enable building on non-x86 systems

mr-c:
Hello again. This is another patch from the Debian package of plink2. It enables building on non-x86 architectures link arm64, ppc64el, s390x, and more: https://buildd.debian.org/status/package.php?p=plink2

This also allows us to compile plink on x86 systems without advanced SIMD usage, as that is required by the Debian x86 64 and 32 bit architecture baselines. On x86 we make several plink2 binaries using the `SFX` shown below. At runtime a dispatcher script selects the most capable binary for the user's CPU.

You can see our compile recipe at https://salsa.debian.org/med-team/plink2/-/blob/master/debian/rules#L36 and the dispatcher script at https://salsa.debian.org/med-team/plink2/-/blob/master/debian/bin/simd-dispatch

If a CPU dispatcher was integrated into plink2 then this external script would not be necessary.

Please let me know what I can do to improve this PR so it gets merged. Thanks!

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "2.0/include/simde"]
+	path = 2.0/include/simde
+	url = https://github.com/nemequ/simde-no-tests
--- a/2.0/Makefile
+++ b/2.0/Makefile
@@ -6,7 +6,7 @@
 
 # Does not currently support -DCPU_CHECK_...
 
-BASEFLAGS=-g -mavx2 -mbmi -mbmi2 -mlzcnt -DZSTD_MULTITHREAD
+BASEFLAGS=-g -DZSTD_MULTITHREAD
 # BASEFLAGS=-g -msse4.2 -DZSTD_MULTITHREAD
 # BASEFLAGS=-g -DZSTD_MULTITHREAD
 
--- a/2.0/build_dynamic/Makefile
+++ b/2.0/build_dynamic/Makefile
@@ -143,17 +143,19 @@ ifdef FORCE_32BIT
   CXXFLAGS += -Wno-sign-compare
 endif
 
-all: plink2 pgen_compress
+SFX ?= ""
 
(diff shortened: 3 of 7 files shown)


chrchang:
Have you tested this on e.g. an ARM machine?  I wouldn't expect this to work at all, because the codebase currently includes some hardcoded unaligned copy operations.  I was planning to clean that up in a year or two (ideally after I could obtain an ARM MacBook to test with).

mr-c:
@chrchang It built on arm64 https://buildd.debian.org/status/package.php?p=plink2&suite=sid and a basic test passed: https://ci.debian.net/packages/p/plink2/

mr-c:
Here's the test script https://salsa.debian.org/med-team/plink2/-/blob/master/debian/tests/run-sample-analysis

chrchang:
Ok, it's possible that I misunderstood the nature of arm64's behavior w.r.t. unaligned copies.  I will try to perform a more thorough test on EC2 within the next few days.

mr-c:
Thanks. If there is a better post-installation test for Debian to run, please let us know

mr-c:
Hello @chrchang ; is there anything I can do to get his PR merged? Thanks!

chrchang:
Sorry about the delay; I am testing this on an EC2 A1 instance now and will merge it today if no problems arise.

chrchang:
I'm currently resolving some system-specific build problems (e.g. "symbol lookup error: ./plink2: undefined symbol: ZSTD_compressStream2").  Will continue working on this over the long weekend.

chrchang:
Ok, confirmed that this basically works, after making the same change to plink 1.9 and running the usual test suite.  I'll make a few tweaks in a subsequent revision so that it's easier for me to build this on a new EC2 A1 instance with Ubuntu 18.04, but merging this as-is first.

PR #203, opened 2021-12-07
[Fix] clang does not support execute yet

choishingwan:


chrchang:
Out of curiosity, what clang version did you run into this problem with?

choishingwan:
This is the clang version information:

clang version 10.0.0
Target: x86_64-apple-darwin20.6.0
Thread model: posix
InstalledDir: /usr/local/opt/llvm/bin


I only got this error when I used -std=c++20 or -std=c++2x when I compile
my software.

Sam

Issue #162, opened 2020-12-21
compile and run cleanly on Mac M1 architecture


jpritikin:
Currently there seem to be assumptions in the code about LP64 and SSE that seem invalid on M1. I don't have M1 hardware myself, but I'd like to know once the 2.0 code compiles and runs correctly on M1.

chrchang:
I probably won't have a Mac to test this on for another year or so, but the current code does compile and run on arm64 EC2 instances due to the simde/ subdirectory, so it's not immediately obvious to me that it would fail to do so on M1 with appropriate compile flags.

mr-c:
@jpritikin In Debian we build plink2 for arm64 without any issues. I also contribute to SIMDe so please let us know if there are any issues on Mac M1, thanks!

DanielEWeeks:
The PLINK code in the 1.9 folder compiles on a Mac M1 after making this one small change to the `Makefile`

```
$ diff Makefile Makefile.org
50c50
<   ZLIB ?= -L. ../zlib-1.2.11/libz.a
---
>   ZLIB ?= -L. ../zlib-1.2.11/libz-64.
```

-------------

The PLINK code in the 2.0 folder does not compile but rather generates this output (on the second try, so it did compile several of the individual files):

```
c++ -c -O2 -std=c++14 -g -mavx2 -mbmi -mbmi2 -mlzcnt -DZSTD_MULTITHREAD -Wall -Wextra -Wshadow -Wformat-security -Wdouble-promotion -Wfloat-conversion -Wold-style-cast -Isimde -I/usr/local/include  -o include/plink2_base.o include/plink2_base.cc
clang: warning: argument unused during compilation: '-mavx2' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-mbmi' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-mbmi2' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-mlzcnt' [-Wunused-command-line-argument]
In file included from include/plink2_base.cc:18:
include/plink2_base.h:1108:23: error: argument to '__builtin_neon_vshrq_n_v' must be a constant integer
  return R_CAST(VecW, _mm_srli_epi64(R_CAST(__m128i, vv), ct));
         ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #207, opened 2022-01-14
Allow Plink 1.9 to compile & run on Apple Silicon 

pettyalex:
Add conditional logic to link to the correct static libz on apple silicon arm64 CPU on Mac OS.

As documented [here](https://github.com/chrchang/plink-ng/issues/162), this is the minimum changeset necessary to allow Plink 1.9 to work with Apple Silicon.

It looks like 2.0 will take a bit more work.

### DIFF
--- a/1.9/Makefile
+++ b/1.9/Makefile
@@ -24,6 +24,8 @@ else
   endif
 endif
 
+ARCH := $(shell uname -p)
+
 # Allow these to be overridden by make arguments or env variables, so people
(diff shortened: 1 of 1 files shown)


chrchang:
At this point, since recent macOS versions don't support 32-bit binaries, it is fine to expect the static library to always be named `libz.a` instead of `libz-64.a`.

pettyalex:
So we can drop the conditional and just always get libz.a on Mac OS? Sounds good.

PR #208, opened 2022-01-14
Fix Plink2 build on Apple Silicon MacOS: Call NEON SIMD instructions with constants.

pettyalex:
Replace the functions that wrap simde with macros, allowing them to work on ARM NEON. NEON requires that bit-shifting instructions are called with constant integers. This lets plink2.0 build on Apple Silicon on MacOS.

I changed the inline functions into macros, which is definitely more fragile and nastier to debug, but otherwise compilation would fail because the underlying NEON intrinsics could only be called with constant integers for the offset, meaning they can't be called from a function that takes runtime parameters. Where there were dynamic calls to these functions, I added a lookup table that will map to the function called with a compile-time constant. This doesn't feel great, but gets Plink2 running on Apple Silicon. I have not yet tested it extensively, but it runs.

I don't love this solution, but it seems to be what other projects are doing:
https://github.com/VectorCamp/vectorscan/pull/81/files#diff-1f738fe3ab986614e926b3ce01fcd42dbf3e8ccb79337931af69450f56319554R178

This would resolve https://github.com/chrchang/plink-ng/issues/162

### DIFF
--- a/2.0/include/plink2_base.h
+++ b/2.0/include/plink2_base.h
@@ -767,17 +767,19 @@ HEADER_INLINE VecW vecw_srli(VecW vv, uint32_t ct) {
   return R_CAST(VecW, _mm256_srli_epi64(R_CAST(__m256i, vv), ct));
 }
 
-HEADER_INLINE VecW vecw_slli(VecW vv, uint32_t ct) {
-  return R_CAST(VecW, _mm256_slli_epi64(R_CAST(__m256i, vv), ct));
-}
+#define vecw_slli(vv, ct) R_CAST(VecW, _mm256_slli_epi64(R_CAST(__m256i, vv), ct))
 
-HEADER_INLINE VecU32 vecu32_srli(VecU32 vv, uint32_t ct) {
-  return R_CAST(VecU32, _mm256_srli_epi32(R_CAST(__m256i, vv), ct));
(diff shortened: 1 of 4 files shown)


chrchang:
I'm fine with the #defines, but the vecw_slli_lookup change would harm performance on non-NEON targets so it should be conditionally compiled.

chrchang:
Well, I'll go ahead and merge this and then add conditional compilation in the immediately following diff.

pettyalex:
No rush to merge, I was planning on revisiting this tomorrow and making it
cleaner.

We could also go in a template direction if you'd like instead?

chrchang:
plink2 code can currently be included in pure C programs, so templates are unlikely to be worth it.

chrchang:
Ok, I'll wait for your change then.

pettyalex:
Corrected.

Unrelated to this changeset, but is there anything that can be done for build_dynamic/Makefile to properly flag NO_SSE42 on Apple Silicon Macs?

chrchang:
> Corrected.
> 
> Unrelated to this changeset, but is there anything that can be done for build_dynamic/Makefile to properly flag NO_SSE42 on Apple Silicon Macs?

That's supposed to be set manually for now.

PR #222, opened 2022-08-26
Update installation instructions for python

KennethEnevoldsen:
Update installation instructions for python to accommodate less experienced users.

It would also be ideal to add a hyperlink to the [documentation](https://www.cog-genomics.org/plink/2.0/), mentioning the python package (I end up spending a bit of time finding my way here).

### DIFF
--- a/2.0/Python/README.md
+++ b/2.0/Python/README.md
@@ -1,11 +1,26 @@
 This provides a basic Python API for pgenlib  (See [python_api.txt](python_api.txt) for details.)
 
 
-##### Build this with this.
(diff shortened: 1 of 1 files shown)


chrchang:
> Update installation instructions for python to accommodate less experienced users.
> 
> It would also be ideal to add a hyperlink to the [documentation](https://www.cog-genomics.org/plink/2.0/), mentioning the python package (I end up spending a bit of time finding my way here).

Okay, it's about time the plink 2.0 "For developers" webpage was filled in; I'll try to take care of it this weekend.

Issue #260, opened 2024-02-27
Pgenlib 0.90.1 is not installable from PyPI with Python 3.12.2 on Mac aarch64


pettyalex:
**OS**: macOS Sonoma 14.2.1 on a 2020 Macbook Air M1
**Python**: Python 3.12.2 from homebrew

While trying to install a Python package that depends on Pgenlib, I encountered this failure from `pip install Pgenlib` in a completely empty venv:

```
      building 'pgenlib' extension
      creating build/temp.macosx-14.0-arm64-cpython-312/src/pgenlib
      creating build/temp.macosx-14.0-arm64-cpython-312/src/plink2/include
      clang -fno-strict-overflow -Wsign-compare -Wunreachable-code -fno-common -dynamic -DNDEBUG -g -O3 -Wall -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX14.sdk -Isrc/pgenlib -I/private/var/folders/ql/rgj8hc250zs662__wyf33y_r0000gn/T/pip-build-env-rljz3v40/overlay/lib/python3.12/site-packages/numpy/core/include -Isrc/plink2/libdeflate -I/Users/alex/scratch/tensorqtl/env/include -I/opt/homebrew/opt/python@3.12/Frameworks/Python.framework/Versions/3.12/include/python3.12 -c src/pgenlib/pgenlib.cpp -o build/temp.macosx-14.0-arm64-cpython-312/src/pgenlib/pgenlib.o -std=c++98 -Wno-unused-function -Wno-macro-redefined -DZSTD_DISABLE_ASM -DLIBDEFLATE_STATIC
      In file included from src/pgenlib/pgenlib.cpp:1258:
      In file included from /private/var/folders/ql/rgj8hc250zs662__wyf33y_r0000gn/T/pip-build-env-rljz3v40/overlay/lib/python3.12/site-packages/numpy/core/include/numpy/arrayobject.h:5:
      In file included from /private/var/folders/ql/rgj8hc250zs662__wyf33y_r0000gn/T/pip-build-env-rljz3v40/overlay/lib/python3.12/site-packages/numpy/core/include/numpy/ndarrayobject.h:12:
      In file included from /private/var/folders/ql/rgj8hc250zs662__wyf33y_r0000gn/T/pip-build-env-rljz3v40/overlay/lib/python3.12/site-packages/numpy/core/include/numpy/ndarraytypes.h:1929:
[...]

PR #261, opened 2024-02-28
fix: Include simde in Pgenlib python package, fixing aarch64 build

pettyalex:
Well, since I noticed a problem with the Pgenlib python package, I might as well fix it.

Add src/plink2/simde directory to the Python package such that it will be available when compiling. This will fix Pgenlib on both Mac aarch64 and Linux aarch64, its broken in both right now.

I noticed that there were old `include` lines in `MANIFEST.in` from when simde headers used to live directly in `include`. I have removed those, they did not match any files and were not doing anything.

I followed your pattern of including specific files rather than using `recursive-include`, but recursive-include would certainly make a simpler, maybe more robust MANIFEST.in.

This _should_ fix https://github.com/chrchang/plink-ng/issues/260. I have tested it on a Raspberry Pi 4 to check `aarch64-linux-gnu`

P.S. I don't know what kind of CI you use to build Pgenlib, but I'd be glad to help you make it produce aarch64 wheels as well. ARM market share is increasing significantly, and it's the best value available on AWS.


pettyalex:
I went ahead and made a workflow to build mac ARM wheels: https://github.com/chrchang/plink-ng/pull/262 as an excuse to learn about cibuildwheel

PR #262, opened 2024-02-28
build: Add a Github Actions workflow that builds wheels on ARM64 Mac OS.

pettyalex:
Add a github actions workflow that will build wheels for Pgenlib python package.

I initially tried to enable Linux wheel building as well with the [cibuildwheel](https://github.com/pypa/cibuildwheel) package, but their docker-based build container doesn't work out of the box with symlinks that link outside of the python package root. I'm sure that I could work around it with some config changes, if you're interested in using Github Actions to build wheels.

Unfortunately, Mac OS runners can be pretty expensive. If you're interested in having an action that checks if Pgenlib is working on Mac, it would be possible to make a much simpler Action that would only check a single Python version.

Let me know what you think. This action would have detected the broken Pgenlib on ARM when it was first broken.


pettyalex:
Oh. That's an interesting outcome. Looking at the build, it is directly building a wheel from the repository without making a source distribution first, which is not what I was expecting.

These wheels should work, but they also will not match the generated sdist, which is potentially a problem. I need to learn a bit more about this.

pettyalex:
@chrchang If opening PRs for this feels like spam and you'd like me to communicate before opening them, just let me know.

I've updated this PR to add two separate github actions:
1. Build the Pgenlib package on a mac OS Apple Silicon runner, just to test that the build is functional in that environment. This is configured to run on every push and PR.
2. Add a manually triggered Github Action to compile wheels for MacOS, Windows, Linux on a large variety of different python versions and architectures. If you're interested, this could make it easy to create tons of wheels to upload to PyPI and simplify many users installs.

chrchang:
> @chrchang If opening PRs for this feels like spam and you'd like me to communicate before opening them, just let me know.
> 
> I've updated this PR to add two separate github actions:
> 
> 1. Build the Pgenlib package on a mac OS Apple Silicon runner, just to test that the build is functional in that environment. This is configured to run on every push and PR.
> 2. Add a manually triggered Github Action to compile wheels for MacOS, Windows, Linux on a large variety of different python versions and architectures. If you're interested, this could make it easy to create tons of wheels to upload to PyPI and simplify many users installs.

This is fine; thank you!

PR #299, opened 2025-05-20
deps: reinstate support for old versions of numpy?

aryarm:
Hi @chrchang ,

I noticed that you recently upgraded pgenlib's minimum version of python from 1.19.0 to 2.0 around the time that you dropped support for py3.8

Cython packages that are compiled with numpy 2.0+ can still be imported and used alongside numpy <2.0

Are you planning to use any features of numpy 2.0+ at run-time (when pgenlib is imported by the user)? If not, I was wondering if I could still request support for old versions of numpy until they become incompatible with pgenlib's minimum python requirement?

This PR allows numpy <2.0 to be installed alongside pgenlib but maintains the requirement that numpy 2.0+ is installed when pgenlib is compiled and built. It also updates the CI workflow to test that things work even with numpy 1.19.3

I also updated the release.yaml workflow to support py3.13 and modified the README to suggest the use of `pip install -e`, which should install the requisite cython/numpy and compile pgenlib automatically

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,6 +10,7 @@ jobs:
   wheel:
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-13, macos-latest]
         python-version: [9, 10, 11, 12, 13]
@@ -44,7 +45,7 @@ jobs:
         working-directory: ./2.0/Python
         if: matrix.python-version <= 9 && matrix.os != 'macos-latest'
         run: |
-          pip install --force-reinstall --no-cache-dir 'numpy==2.0.0'
+          pip install --force-reinstall --no-cache-dir 'numpy==1.19.3'
           pytest tests/
 
       - name: Test against newest numpy version
@@ -59,6 +60,8 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
+        # The minimum and maximum supported versions of python that are not
+        # "end-of-life," "prerelease," or "feature": https://devguide.python.org/versions
         python-version: [9, 13]
 
     steps:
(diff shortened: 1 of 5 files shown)


aryarm:
ah, I think the tests are failing because there have been new commits to master since cc4175bf1840006808806c30f5f1cb9a9b26f0cc

When I forked this repo at cc4175bf1840006808806c30f5f1cb9a9b26f0cc and applied these changes, the tests succeeded:
https://github.com/aryarm/plink-ng/tree/ci/fix

chrchang:
What is the reason you previously specified numpy>=2.0.0 for python >= 3.9?

aryarm:
> What is the reason you previously specified numpy>=2.0.0 for python >= 3.9 in pyproject.toml?

The `install_requires` of `setup.py` tells pip the minimum _run-time_ version of numpy. This is the min version that must be available when pgenlib is used.

https://github.com/chrchang/plink-ng/blob/b522f177a3c495ddbbfcbe40f1fe021a8eaf6ee2/2.0/Python/setup.py#L106-L108

The `requires` section of pyproject.toml tells pip the minimum _build-time_ version of numpy. This is the min version that must be available when pgenlib is compiled and built.

https://github.com/chrchang/plink-ng/blob/b522f177a3c495ddbbfcbe40f1fe021a8eaf6ee2/2.0/Python/pyproject.toml#L2-L7

They don't have to be the same, and actually it's better if they aren't! If pgenlib is built with numpy >2.0, it can still run with numpy <2.0. However, if pgenlib is built with numpy <2.0, it can't be run with numpy >2.0. ([source](https://numpy.org/devdocs/dev/depending_on_numpy.html#numpy-2-0-specific-advice))

So it's best to compile with numpy >2.0 whenever possible. In fact, it isn't possible for python <3.9 since numpy 2.0 doesn't support those versions of python. That's why I had to add a `>= 3.9` version constraint in the pyproject.toml and fallback to `oldest-supported-numpy` otherwise. It's fine that you deleted that line when you dropped support for py3.8.

The min run-time version of numpy can remain unaffected, though. I'm suggesting to allow and test pgenlib with 1.19.3 since that's the first version of numpy to support py3.9

chrchang:
Ah, ok, makes sense.

PR #300, opened 2025-05-21
ci: enable py3.13 wheels and workaround pypy issue

aryarm:
Sorry for the noise, @chrchang

Here's one more change to the release.yaml workflow to 1) upgrade to the latest `cibuildwheel` to allow py3.13 wheels to be built and to 2) workaround the pypy issue that you mentioned here:
https://github.com/chrchang/plink-ng/blob/8671e46090a987a35f100d82cd301eb5a9682a86/2.0/manylinux-build-wheels.sh#L14-L15

I tested this on my fork and was able to build all of the wheels:
https://github.com/aryarm/plink-ng/actions/runs/15167833912

Also, I noticed that the latest 0.92.0 is missing wheels for aarch64 and some others?
https://pypi.org/project/Pgenlib/0.92.0/#files
https://pypi.org/project/Pgenlib/0.91.0/#files
If that's because of something that is wrong with the workflow, please let me know! I'm always happy to fix it

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -30,7 +30,7 @@ jobs:
           platforms: all
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2
+        uses: pypa/cibuildwheel@v2.23
         with:
           package-dir: ./2.0/Python
         env:
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -30,7 +30,7 @@ jobs:
           platforms: all
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.19.1
+        uses: pypa/cibuildwheel@v2
         with:
           package-dir: ./2.0/Python
         env:
--- a/2.0/Python/pyproject.toml
+++ b/2.0/Python/pyproject.toml
@@ -1,7 +1,9 @@
 [build-system]
 requires = [
     "wheel",
-    "setuptools",
+    # pin setuptools on pypy to workaround this bug: https://github.com/pypa/distutils/issues/283 https://github.com/pypa/cibuildwheel/issues/1983
+    "setuptools < 72.2; platform_python_implementation == 'PyPy'",
+    "setuptools ; platform_python_implementation != 'PyPy'",
     "Cython>=3.1.0",
     "numpy>=2.0.0",
 ]


chrchang:
Hmm, best guess is that I ran manylinux-build-wheels.sh on both manylinux2014_x86_64 and manylinux2014_aarch64 Docker images for 0.91.0, and I forgot the latter for 0.92.0.

I will try the GitHub workflow next time.

chrchang:
(Correction, I now see that I did use the GitHub action for 0.91.0.  Anyway, just launched it for 0.92.1, we'll see if it works.)

chrchang:
"Publish distribution to Test PyPI" step succeeded, but then "Publish distribution to PyPI" failed with

```
Checking dist/pgenlib-0.92.1.tar.gz: PASSED
Notice: Generating and uploading digital attestations
Traceback (most recent call last):
  File "/app/attestations.py", line 149, in <module>
    main()
  File "/app/attestations.py", line 129, in main
    dist_to_attestation_map = compose_attestation_mapping(
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/attestations.py", line 103, in compose_attestation_mapping
    assert_attestations_do_not_pre_exist(dist_to_attestation_map)
  File "/app/attestations.py", line 88, in assert_attestations_do_not_pre_exist
    die(error_message)
  File "/app/attestations.py", line 45, in die
    msg = msg.replace('\n', '%0A')
          ^^^^^^^^^^^
AttributeError: 'tuple' object has no attribute 'replace'
```

aryarm:
Thanks for merging this and trying to publish 0.92.1!

> Publish distribution to Test PyPI" step succeeded, but then "Publish distribution to PyPI" failed
I _think_ the AttributeError is actually a red herring. The actual issue seems to be reported in [the actions summary](https://github.com/chrchang/plink-ng/actions/runs/15174821461#summary-42673848502):
```
Attestation generation failure:
('The following distributions already have publish attestations:
```

Folks online seem to say that it isn't good to have the "Test PyPI" step in the same workflow as the "Publish distribution to PyPI" step. My guess is that the "Test PyPI" step is generating attestations which are then conflicting with attestations that may be generated by the "Publish distribution to PyPI" step. To fix this, could we try running "Test PyPI" only on non-master branches?

```diff
      - name: Publish distribution to Test PyPI
        uses: pypa/gh-action-pypi-publish@release/v1
+        if: github.ref != 'refs/heads/master'
        with:
          repository-url: https://test.pypi.org/legacy/

      - name: Publish distribution to PyPI
        if: github.ref == 'refs/heads/master'
        uses: pypa/gh-action-pypi-publish@release/v1
```

PR #318, opened 2026-01-13
Add licensing information

jglamp:
The website and previous versions had this program as being GPLv3 licensed. Added licensing information for posterity and inclusion in repos in the future.


chrchang:
https://www.gnu.org/licenses/gpl-howto.html

"Add a COPYING file with a copy of the GNU GPL or GNU AGPL."

Issue #327, opened 2026-05-16
--vcf: variant metadata table sized to input VCF, not --extract output (OOM on WGS-scale inputs)


carstenerickson:
## Summary

When `--vcf` is paired with `--extract`, plink2 allocates a variant-metadata table sized to the **total record count of the input VCF**, not to the number of records that survive the `--extract` filter. On whole-genome inputs (e.g. ~1B-record VCFs produced by `bcftools mpileup | call -m` without `-v`), this triggers OOM on 64 GB hosts well before any user-visible filter can apply — even with `--memory`, `--max-alleles 2`, and `--snps-only just-acgt` set.

## Reproducer (synthetic, runs in seconds)

```bash
# 1) Generate VCFs of 1M / 5M / 10M / 30M records, alternating genotypes
#    so plink2 doesn't reject as malformed. Constant 100K-ID extract list.
python3 gen_vcf.py 1000000  vcf_1M.vcf
python3 gen_vcf.py 5000000  vcf_5M.vcf
python3 gen_vcf.py 10000000 vcf_10M.vcf
python3 gen_vcf.py 30000000 vcf_30M.vcf
python3 -c "open('extract_100k.txt','w').writelines(f'rs{i}\n' for i in range(1,100001))"

# 2) For each VCF, run plink2 with /usr/bin/time -l to capture peak RSS:
/usr/bin/time -l plink2 --vcf vcf_NM.vcf --extract extract_100k.txt \
    --make-pgen --out out_NM --threads 4 --memory 4096
```

`gen_vcf.py` is a 25-line script that writes biallelic SNPs on `chr1`, alternating `0/0`, `0/1`, `1/1` genotypes (full source at the bottom of this issue).

## Observed scaling (plink2 v2.0.0-a.7.1 M1, macOS 14, 16 GB host)

| Records | VCF size (raw) | VCF size (bgzipped) | Extract IDs | Output `.pvar` records | **Peak RSS** | Wall (s) |
|--------:|---------------:|--------------------:|------------:|-----------------------:|-------------:|---------:|
| 1M  | 36 MB | — | 100K | 100K | **123 MB** | 1.6 |
| 5M  | 189 MB | — | 100K | 100K | **464 MB** | 4.3 |
| 10M | 379 MB | 37 MB | 100K | 100K | **596 MB** (raw) / 996 MB (bgzipped) | 6.0 / 2.6 |
[...]

PR #329, opened 2026-05-16
vcf_subset: add --extract / --exclude variant-ID filters

carstenerickson:
Per your suggestion in #327, this adds ID-list `--extract <fname>` / `--exclude <fname>` to `utils/vcf_subset` rather than pushing the filter into `VcfToPgen`. The two-stage pipeline `vcf_subset --extract → plink2 --vcf` then replaces `bcftools view -R` as the pre-filter for whole-genome `plink2 --vcf --extract` workflows.

Sample output:

```
$ vcf_subset --in big.vcf.gz --extract panel.snps.txt --out filtered.vcf.gz
--extract: 1400000 IDs loaded from panel.snps.txt.
vcf_subset: 1/1 sample and 1400000/1000000000 variants written to filtered.vcf.gz (998600000 excluded by --extract).
```

## Validation

End-to-end: `vcf_subset --extract list.txt → plink2 --vcf` and `plink2 --vcf --extract list.txt` produce **byte-identical .pgen** on a 10M-record synthetic VCF + 100K-ID extract list. Same identity for `--exclude` only, `--extract + --exclude` combined, and extract files containing duplicate IDs. `--keep` / `--indv` / no-filter paths remain byte-identical to baseline.

Peak RSS scaling (100K-ID extract list, 1 sample, M1 build, `/usr/bin/time -l`):

| Records | Input VCF | Peak RSS |
|--------:|----------:|---------:|
| 1M  | 36 MB   | 19.5 MB |
| 5M  | 189 MB  | 13.5 MB |
| 10M | 379 MB  | 13.5 MB |
| 30M | 1.2 GB  | 20.0 MB |

## Notes

- BCF input is rejected at the existing `VcfSubset` entry point, so the new filter never runs there.
- Coordinate-form `--extract bed{0,1}` is out of scope — ID-list only, matching `--keep`'s convention in this tool.

### DIFF
--- a/2.0/utils/vcf_subset/vcf_subset.cc
+++ b/2.0/utils/vcf_subset/vcf_subset.cc
@@ -81,6 +81,8 @@ void DispUsage(FILE* dst_stream) {
 "                       that compression occurs iff the --out path ends in '.gz'\n"
 "                       or '.bgz'.\n"
 "  --erase-info       : Remove INFO header lines, set all INFO entries to '.'.\n"
+"  --exclude <fname>  : Keep only variants whose ID is NOT in the given file.\n"
+"  --extract <fname>  : Keep only variants whose ID is in the given file.\n"
 "  --help             : Just display this help text.\n"
 "  --indv <ID>        : Keep only the sample with the given ID.\n"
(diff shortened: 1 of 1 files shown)


chrchang:
Looks great, thanks!

Issue #330, opened 2026-05-17
--rm-dup: support (CHROM, POS) as the dedup match key (not just variant ID)


carstenerickson:
## Summary

`--rm-dup` matches on variant ID only. PFILE rows sharing
`(CHROM, POS)` but with distinct IDs (canonical case: multi-allelic
loci atomized to biallelic rows upstream of plink2 via `bcftools norm
-m-` or similar) survive all `--rm-dup` actions, then break every
downstream tool that keys on `(CHROM, POS)` — fraposa, ADMIXTURE,
qpAdm, EIGENSOFT, BEAGLE phasing.

Request: let `--rm-dup`'s match key be `(CHROM, POS)` instead of (or
in addition to) ID.

## Concrete case

LD-pruning the HGDP+1kGP v1 panel. After upstream atomization with
`bcftools norm -m-`, the source VCF has biallelic rows at shared
positions with distinct ALTs. plink2's `--max-alleles 2 --snps-only
just-acgt` keeps them, so e.g. `1:2654169:C:G` and `1:2654169:C:T`
both survive `--rm-dup force-first`. ~90 of ~290 K pruned variants
needed manual position-level re-dedupe via a brittle Python pass
that depends on `--set-all-var-ids '@:#:$r:$a'` so the ID is
colon-parseable.

[admix-kit](https://github.com/KangchengHou/admix-kit/blob/main/admix/cli/_utils.py#L201-L219)
(Bogdan-lab admixture toolkit, 30 stars, active 2026-04) ships the
same workaround independently: `plink2 --rm-dup exclude-all` then a
pandas `df_pvar.duplicated(subset=["CHROM", "POS"])` post-pass. Two
tools converging on the same patch.

## Minimal synthetic reproducer

```bash
cat > dup_pos.vcf <<VCF
##fileformat=VCFv4.2
##contig=<ID=1,length=300000000>
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	S1
1	2654169	1:2654169:C:G	C	G	.	PASS	.	GT	0/1
1	2654169	1:2654169:C:T	C	T	.	PASS	.	GT	0/0
VCF

plink2 --vcf dup_pos.vcf --rm-dup force-first --make-pgen --out dedup
wc -l dedup.pvar  # both rows survive; the (chrom, pos) duplicate stays.
```

## Proposed CLI

Extend `--rm-dup` with a `match-mode=<id|pos|id-or-pos>` modifier,
[...]

PR #331, opened 2026-05-17
--rm-dup: add match-mode=<id|pos> for position-keyed dedup

carstenerickson:
Addresses #330.  Adds a `match-mode=<id|pos>` modifier to `--rm-dup`: the default `id` preserves the existing behavior verbatim, while `pos` dispatches to a new `RmDupPos` that dedups by (CHROM, POS) -- the atomized-multi-allelic case from the issue.

Sample output (the #330 reproducer):

```
$ plink2 --vcf dup_pos.vcf --rm-dup force-first match-mode=pos --make-pgen --out dedup
--rm-dup match-mode=pos force-first: 1 duplicate position group (1 variant excluded), 1 variant remaining.
```

## Design

`RmDupPos` is a separate function so the ID-mode `RmDup` doesn't have to grow new mode branches around its allele/INFO/genotype-comparison logic.  Pos-mode skips the ID-htable build, INFO reload, and allele/qual/info/genotype mismatch checks -- none of those are meaningful when dups are detected by (CHROM, POS) alone (atomized-multi-allelic dups have intentionally distinct ALTs by construction).

Single linear pass per chromosome, bounded by `PopcountBitRange` to match the existing plink2 iteration idiom.  Relies on plink2's invariant that `variant_bps` is non-decreasing within a chromosome's vidx range; emits a `--sort-vars`-pointing error if violated.  `retain-mismatch` / `exclude-mismatch` actions are rejected at parse when combined with `match-mode=pos`.

## Validation

9 tests, all pass:

| # | What | Result |
|---|---|---|
| 1 | `--rm-dup force-first` on distinct-ID 2-row VCF | byte-identical .pgen to baseline ✓ |
| 2 | `match-mode=pos force-first` on the #330 reproducer | 1 of 2 survives ✓ |
| 3 | `match-mode=pos exclude-all` on 3-row VCF | dup pair dropped, lone variant kept ✓ |
| 4 | `match-mode=pos error` with dups | exit 7 + diagnostic ✓ |
| 5 | `match-mode=pos retain-mismatch` | parser rejects (exit 8) ✓ |
| 6 | `--rm-dup force-first` on 5M-VCF (no match-mode) | byte-identical to baseline ✓ |
| 7 | 4-way + lone + 2-way mixed | survivors `v100a, v200, v300a` ✓ |
| 8 | Unsorted-within-chromosome `(500, 500, 300)` | sortedness diagnostic fires ✓ |
| 9 | Cross-chromosome (chr1:100 + chr2:100 + dups on each) | `c1a, c1c, c2a, c2b` survive ✓ |

Build clean under `-Wall -Wextra -Wshadow -Wold-style-cast`.

## Notes

- ID-mode default behavior is unchanged; the `Plink2Cmdline` struct gains a `rmdup_match_mode` field that defaults to `kRmDupMatchModeId`.
- `--rm-dup list match-mode=pos` writes the same `<out>.rmdup.list` format as ID-mode.
- `match-mode=id-or-pos` (drop on either key) is mentioned in #330 as the eventual use case, but is deferred -- the dup-group merging semantics for `force-first` need design discussion before code.

### DIFF
--- a/2.0/plink2.cc
+++ b/2.0/plink2.cc
@@ -403,6 +403,7 @@ typedef struct Plink2CmdlineStruct {
   VscoreFlags vscore_flags;
   AlleleAlphanumFlags allele_alphanum_flags;
   RmDupMode rmdup_mode;
+  RmDupMatchMode rmdup_match_mode;
   STD_ARRAY_DECL(FreqFilterMode, 4, filter_modes);
   SelectSidMissingnessMode select_sid_missingness_mode;
   SelectSidTiebreakMode select_sid_tiebreak_mode;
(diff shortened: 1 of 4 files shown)


carstenerickson:
Redundant with the existing `--set-all-var-ids '@:#' --rm-dup --recover-var-ids` recipe (see #330).  Closing.

PR #333, opened 2026-06-05
fix segfault in pvar files containing deletions where ID=.

jeremymcrae:
This tries to fix a problem loading pvar files where deletions have single character IDs e.g. ".". I was getting segfaults when loading these files. Here's some code to reproduce the issue:

```py
from pgenlib import PvarReader

# make synthetic pvar file for deletions where ID=.
pvar_path = 'test.pvar'
with open(pvar_path, 'w') as f:
    f.write('#CHROM\tPOS\tID\tREF\tALT\n')
    for pos in range(1, 100):
        f.write(f'1\t{pos}\t.\tACGTA\tA\n')

pvar = PvarReader(pvar_path.encode())
pvar.close()  # typically segfaults (on linux at least, I didn't check other OSes)
```


chrchang:
Oops, thanks for fixing this!

Issue #341, opened 2026-08-31
--freq counts emits fractional allele counts for het haploid calls, and disagrees with --geno-counts/--hardy on the same file


BenjaminDEMAILLE:
Reproduced with `v2.0.0-a.7.4 M1 (18 Aug 2026)`, built from `master` (22df061) with `2.0/Makefile` on macOS/arm64. Also present in the a.7.x Linux builds I checked.

### Repro

`repro.vcf`:

```
##fileformat=VCFv4.2
##contig=<ID=X,length=200000000>
##FORMAT=<ID=GT,Number=1,Type=String,Description="Genotype">
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	M1	M2	M3	F1	F2
X	5000000	xv1	A	G	.	.	.	GT	0/1	0/0	1/1	0/1	0/0
```

`sex.txt`:

```
#IID	SEX
M1	1
M2	1
M3	1
F1	2
F2	2
```

`M1` is a male with a heterozygous chrX call, i.e. a het haploid call.

```
plink2 --vcf repro.vcf --update-sex sex.txt --make-pgen --out repro
plink2 --pfile repro --freq counts   --out o1
plink2 --pfile repro --geno-counts   --out o2
plink2 --pfile repro --hardy         --out o3
```

### Observed

`o1.acount`:

```
#CHROM	ID	REF	ALT	ALT_CTS	OBS_CT
X	xv1	A	G	2.5	7
```

`o2.gcount`:

```
#CHROM	ID	REF	ALT	HOM_REF_CT	HET_REF_ALT_CTS	TWO_ALT_GENO_CTS	HAP_REF_CT	HAP_ALT_CTS	MISSING_CT
X	xv1	A	G	1	1	0	1	1	1
```

`o3.hardy.x`:

```
#CHROM	ID	A1	AX	FEMALE_HOM_A1_CT	FEMALE_HET_A1_CT	FEMALE_TWO_AX_CT	MALE_A1_CT	MALE_AX_CT	...
X	xv1	A	G	1	1	0	1	1	...
```

`--geno-counts` and `--hardy` both treat M1's call as missing, which matches the documented behaviour ("Heterozygous haploid calls are treated as missing") and gives 2 ALT observations out of 6. `--freq counts` instead splits the call 0.5/0.5 and reports `ALT_CTS 2.5` out of `OBS_CT 7`, so the three commands disagree on the same file. None of the three logs contains a warning.

[...]

PR #342, opened 2026-08-31
chrX: treat het haploid hardcalls as missing in allele-count computation (#341)

BenjaminDEMAILLE:
Addresses the biallelic chrX half of #341.

### Problem

In `LoadAlleleAndGenoCountsThread()` (`2.0/plink2_data.cc`), the biallelic chrX branch starts from

```c
uintptr_t alt1_ct = 4 * genocounts[2] + 2 * genocounts[1] - 2 * sex_specific_genocounts[2] - hethap_ct;  // nonmales count twice
```

`alt1_ct` is in half-allele units. A nonmale het contributes 2, a male hom-alt contributes `4 - 2 = 2`, and a male het contributes `2 - 1 = 1`, i.e. half a REF allele plus half an ALT allele. `weighted_obs_ct` further down still counts that male as a full observation.

So on a file containing het haploid calls, `--freq` and `--freq counts` disagree with `--geno-counts` and `--hardy` on the same variant, and `--freq counts` emits non-integer values. Five-sample repro from #341:

```
--freq counts   ALT_CTS 2.5   OBS_CT 7
--geno-counts   HOM_REF 1  HET 1  TWO_ALT 0  HAP_REF 1  HAP_ALT 1  MISSING 1   -> 2 out of 6
--hardy         MALE_A1_CT 1  MALE_AX_CT 1                                     -> male het excluded
PLINK 1.9       C1 2  C2 4  NCHROBS 6
```

### Change

Subtract the remaining half-allele from `alt1_ct`, and drop the observation from `weighted_obs_ct`.

Male het hardcalls that carry a dosage are excluded from the correction: the existing dosage loop already removes their full hardcall contribution via `alt1_ct -= hardcall_code * sex_multiplier`, so applying the correction to them would double-subtract (and underflow `alt1_ct`, which is a `uintptr_t`). They are counted in the new `hethap_dosage_ct` inside the loop, where `sex_multiplier` and `hardcall_code` are already in hand, so there is no extra pass over the data.

Net effect: het haploid calls are treated as missing for allele-count purposes, matching `--geno-counts`, `--hardy`, the `--set-invalid-haploid-missing` result, and PLINK 1.9. Only variants that actually contain het haploid hardcalls change.

### Verification

macOS/arm64, `2.0/Makefile` dev build, compared against a binary built from `master` (22df061) on the same machine.

- `2.0/Tests/run_tests.sh`: 8/8 pass.
- **No change on valid data.** 300 samples, 400 variants, chr1 + chrX, males hemizygous on chrX, 4% missing. Output byte-identical to master for `--freq`, `--freq counts`, `--geno-counts`, `--missing`, `--hardy`, `--het`, `--sample-counts`, `--maf 0.2 --write-snplist`, `--hwe 0.001 --write-snplist`, `--pca 3 approx`, `--make-bed`, `--export vcf`, `--indep-pairwise 50 5 0.5`, plus `--keep` on a 150-sample subset and `--threads 1`.
- **Intended change on het-haploid data.** 300 samples, 400 chrX variants, 15% of male calls heterozygous:

  | command | vs master | vs `--set-invalid-haploid-missing` |
  |---|---|---|
  | `--freq counts` | differs | identical |
  | `--freq counts --keep` (150 samples) | differs | identical |
  | `--freq counts --threads 1` | differs | identical |
  | `--freq` | differs | identical |
  | `--geno-counts` | identical | identical |
  | `--hardy` | identical | identical |
  | `--maf 0.2 --write-snplist` | identical | identical |

  201 of the 400 variants had a fractional `ALT_CTS` under master; none do now.
- The 5-sample repro now gives `ALT_CTS 2, OBS_CT 6`, matching PLINK 1.9 and `--geno-counts`.
- The dosage case (`GT:DS 0/1:0.7` for a male on chrX) is unchanged by this patch and continues to match `--set-invalid-haploid-missing keep-dosage`.

### Scope

Deliberately limited to the biallelic chrX path. The multiallelic chrX path (`all_dosages[0] -= 2 * ss[0] + hethap_ct;` and the `patch_10` male ALTx/ALTy handling) and the chrY / non-XY-haploid paths half-count het haploid calls the same way; #341 describes them. I left them alone because the right semantics there are less obvious, in particular for chrMT where `--geno-counts` currently reports het haploid calls as `HET_REF_ALT_CTS` rather than missing. Happy to extend this once you say which behaviour you want.

### DIFF
--- a/2.0/Tests/TEST_CHRX_HAPLOID/.gitignore
+++ b/2.0/Tests/TEST_CHRX_HAPLOID/.gitignore
@@ -1,0 +1,5 @@
+*.log
+chrx*
+subset.id
+plink1_*
+plink2_*
--- a/2.0/Tests/TEST_CHRX_HAPLOID/check_counts.py
+++ b/2.0/Tests/TEST_CHRX_HAPLOID/check_counts.py
@@ -1,0 +1,56 @@
+#!/usr/bin/env python3
+"""Verify that --freq counts agrees with --geno-counts on chrX.
+
+Both are computed from the same hardcalls, so for a biallelic variant
+
+  ALT_CTS == HET_REF_ALT_CTS + 2 * TWO_ALT_GENO_CTS + HAP_ALT_CTS
+  OBS_CT  == 2 * (HOM_REF_CT + HET_REF_ALT_CTS + TWO_ALT_GENO_CTS)
+             + HAP_REF_CT + HAP_ALT_CTS
+
+must hold exactly, with het haploid calls landing in MISSING_CT and thus in
+neither side.  ALT_CTS must also be an integer.
+"""
+import sys
+
+
+def load(fname):
+    rows = {}
+    with open(fname) as f:
+        header = f.readline().split()
+        id_idx = header.index('ID')
+        for line in f:
+            fields = line.split()
+            rows[fields[id_idx]] = dict(zip(header, fields))
+    return rows
+
+
+acount = load(sys.argv[1])
+gcount = load(sys.argv[2])
+assert len(acount) == len(gcount), 'variant count mismatch'
+assert acount, 'no variants'
+
+problem_ct = 0
+for variant_id, arow in acount.items():
+    grow = gcount[variant_id]
+    hom_ref = int(grow['HOM_REF_CT'])
+    het = int(grow['HET_REF_ALT_CTS'])
+    two_alt = int(grow['TWO_ALT_GENO_CTS'])
+    hap_ref = int(grow['HAP_REF_CT'])
+    hap_alt = int(grow['HAP_ALT_CTS'])
+    expected_alt = het + 2 * two_alt + hap_alt
+    expected_obs = 2 * (hom_ref + het + two_alt) + hap_ref + hap_alt
+    alt_str = arow['ALT_CTS']
+    obs_str = arow['OBS_CT']
+    if ('.' in alt_str) or ('.' in obs_str):
+        print('%s: non-integer --freq counts output: ALT_CTS %s, OBS_CT %s' %
+              (variant_id, alt_str, obs_str))
+        problem_ct += 1
+        continue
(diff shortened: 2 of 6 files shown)


BenjaminDEMAILLE:
Added `2.0/Tests/TEST_CHRX_HAPLOID`, wired into `2.0/Tests/run_tests.sh`.

It generates a chrX fileset where 20% of the male calls are heterozygous, then asserts that `--freq counts` and `--geno-counts` agree on every variant:

```
ALT_CTS == HET_REF_ALT_CTS + 2 * TWO_ALT_GENO_CTS + HAP_ALT_CTS
OBS_CT  == 2 * (HOM_REF_CT + HET_REF_ALT_CTS + TWO_ALT_GENO_CTS) + HAP_REF_CT + HAP_ALT_CTS
```

plus that `ALT_CTS` is an integer, that `--set-invalid-haploid-missing` does not change either output, and that the same holds on a 99-sample subset (`sample_ct != raw_sample_ct` path) and under `--threads 1`.

It needs no external data and no PLINK 1.9, only `python3`, which the suite already uses. Against a `master` build it fails and reports both symptoms:

```
x0: --freq counts says ALT_CTS 154 / OBS_CT 288, --geno-counts implies 146 / 272
x1: non-integer --freq counts output: ALT_CTS 138.5, OBS_CT 288
```

With the fix, `2.0/Tests/run_tests.sh` passes 9/9. Drop the second commit if you would rather not take the test.

chrchang:
As discussed in https://github.com/chrchang/plink-ng/issues/341 , the current behavior is intentional.  There is one set of commands, containing --freq (with or without 'counts'), which works with dosages, and another set of commands, including --geno-counts, which can only work with hardcalls and has to treat all dosages too far from the nearest integer as missing.  We can discuss how to make this distinction more obvious in the documentation.

Issue #345, opened 2026-08-31
Six error paths set g_logbuf but never print it, so the run fails silently (--mac, --update-map, --tped, --pedmap, .snp cm, --export bcf)


BenjaminDEMAILLE:
Found while adding `--update-cm`. Six places compose an error message into `g_logbuf` and then `goto` a label that only sets `reterr`, so the message is never emitted. The run exits with a nonzero code and **nothing on stderr**.

The worst of these is `--mac`: an unparseable command-line argument kills the run with no explanation at all.

```
$ plink2 --bfile base --mac notanumber --make-bed --out m1 2>err.txt >out.txt
$ echo $?
8
$ cat err.txt          # empty
$ tail -1 out.txt
Start time: Tue Sep  1 00:10:31 2026
```

PLINK 1.9 on the same input:

```
Error: Invalid --mac parameter 'notanumber'.
```

### The six sites

| file:line | trigger | exit code | what should be printed |
|---|---|---:|---|
| `plink2.cc:8818` | `--mac <non-numeric>` | 8 | `Error: Invalid --mac argument '...'.` |
| `plink2_misc.cc:283` | `--update-map` file with a non-numeric bp column | 6 | `Error: Invalid bp coordinate on line N of --update-map file.` |
| `plink2_import_legacy.cc:795` | `--tfile` with an empty .tped | 13 | `Error: <name>.tped is empty.` |
| `plink2_import_legacy.cc:2047` | `--pedmap` with an empty .ped (non-empty .map) | 13 | `Error: <name>.ped is empty.` |
| `plink2_import.cc:17027` | `--eigfile` with a non-numeric cm column in .snp | 6 | `Error: Invalid centimorgan position on line N of .snp file.` |
| `plink2_export.cc:7647` | `--export bcf` where a .pvar `##INFO`/`##FILTER` line does not end with `>` | 6 | `Error: INFO:<id> header line in .pvar file doesn't end with '>'.` |

All six reproduce on `master` (22df061), macOS/arm64, `2.0/Makefile` build. The pattern in each case is:

```c
snprintf(g_logbuf, kLogbufSize, "Error: ...");
goto Foo_ret_MALFORMED_INPUT;
...
Foo_ret_MALFORMED_INPUT:
  reterr = kPglRetMalformedInput;   // g_logbuf never printed
  break;
```

versus the `_WW` labels elsewhere in the same functions, which do `WordWrapB(0); logerrputsb();` first.

[...]

Issue #344, opened 2026-08-31
Feature-parity audit: 266 PLINK 1.9 flags across 18 modules have no PLINK 2.0 equivalent


BenjaminDEMAILLE:
Follow-up to #338. Since the alpha state is expected to persist for a while, the artifact that would help downstream users and packagers most is a precise answer to "what can PLINK 1.9 do that PLINK 2.0 cannot". I could not find one anywhere, so I generated it mechanically.

### Method

Candidate flag names were harvested from both programs' `--help` output and from the help tables in `1.9/plink_help.c` and `2.0/plink2_help.cc`, then **every** name was passed to both binaries. Each binary's own `Unrecognized flag` message is the ground truth, so nothing below depends on documentation being current. Script at the end reproduces the counts exactly.

Binaries: `PLINK v1.9.0-b.7.11 (19 Aug 2025)` and `PLINK v2.0.0-a.7.4 (18 Aug 2026)`, the latter built from `master` (22df061).

### Headline

579 real PLINK 1.9 flags:

| | count |
|---|---:|
| recognized by plink2 | 220 |
| recognized by plink2, retired with a documented replacement | 26 |
| **not recognized by plink2** | **333** |

Of those 333, 46 are renames or modifiers that plink2 does cover (section C), and 21 are the `--hap...` family that 1.9 itself never reimplemented from 1.07 (section F). **The genuine 1.9 → 2.0 capability gap is 266 flags across 18 modules.** Every flag is accounted for exactly once in the partition below (333 = sum of all rows).

---

## A. Whole modules with no plink2 substitute

| module | flags | notes |
|---|---:|---|
| **CNV / segmental** `--cnv-*`, `--cfile` | 34 | relevant to #61, #62, #64. Note 1.9 itself gates `--cnv-list`/`--cfile` behind *"either unfinished or not yet well-tested"*, so this has been effectively unavailable since 1.07 |
| **Set-based tests / annotation / gene reports** `--set*`, `--make-set*`, `--subset`, `--complement-sets`, `--write-set*`, `--set-test*`, `--set-table`, `--set-p`, `--set-r2*`, `--gene*`, `--annotate*`, `--attrib*`, `--range`, `--border` | 33 | no set definition, no set-based tests, no `--annotate`, no `--gene-report` |
[...]

PR #347, opened 2026-08-31
Add --update-cm and --zero-cms

BenjaminDEMAILLE:
From the parity audit in #344: two of the three PLINK 1.9 centimorgan flags with no 2.0 equivalent.

plink2 already reads and writes the CM column (`--make-just-pvar cols=+cm` round-trips it, `--indep-pairphase` has `--ld-window-cm`), but there was no way to populate or clear it.

### `--update-cm <f> [cmcol] [IDcol] [skip]`

Mirrors `--update-map` exactly: column-index and skip-count arguments, "appears multiple times" errors, and the "N values updated, M variant IDs not present" summary. `UpdateVarCms()` is modeled on `UpdateVarBps()`.

When the input has no CM column, `LoadPvar()` leaves `variant_cms` null, so `--update-cm` allocates and zero-fills the array first. That allocation is placed **before** the variant-ID hash table's `bigstack_mark`, because that region is reset in the `--update-name` path.

### `--zero-cms`

Sets `variant_cms` to null, which is the same end state as loading a fileset with no nonzero CM values. The writers already handle that (`if (!variant_cms) { *cswritep++ = '0'; }`), so `--zero-cms --make-just-pvar cols=+cm` writes a column of zeroes rather than dereferencing null.

The two flags are mutually exclusive.

`--cm-map` (SHAPEIT recombination-map interpolation) is the remaining 1.9 CM flag; it is not part of this PR.

### Verification

macOS/arm64, `2.0/Makefile` dev build, against PLINK 1.9 on a 1000-sample / 1800-variant fileset.

| check | result |
|---|---|
| `--update-cm cmfile.txt --make-bed` | .bim **byte-identical** to `plink --update-cm` |
| `--update-cm hdr.txt 2 1 1` (explicit cols + skipped header) | identical, both report 1800 values |
| `--update-cm rev.txt 1 2` (CM column before ID column) | identical |
| `--zero-cms --make-bed` | .bim **byte-identical** to `plink --zero-cms` |
| partial file (5 of 1800 IDs, 1 unknown ID) | `--update-cm: 5 values updated, 1 variant ID not present.`; 1.9 reports the same counts |
| `--extract` + `--update-cm` | update applied before the filter, same as `--update-map` and same as 1.9 |
| duplicate ID in file | `Error: Variant ID 'null_0' appears multiple times in --update-cm file.` |
| non-numeric CM value | `Error: Invalid centimorgan position on line 1 of --update-cm file.` (exit 6) |
| `--update-cm` + `--zero-cms` | rejected, in either order |
| input with no CM column, then `--update-cm --make-just-pvar` | CM column appears with the new values |
| `--zero-cms --make-just-pvar cols=+cm` | column of `0`, no crash |

No behaviour change when neither flag is used: output byte-identical to a `master` build for `--make-bed`, `--freq counts`, `--make-just-pvar cols=+cm`, `--hardy`, `--missing`, `--export vcf`.

`2.0/Tests/run_tests.sh` passes 8/8. Build is warning-free.

Note: the `Error: Invalid centimorgan position...` path prints because `UpdateVarCms_ret_MALFORMED_INPUT` calls `logerrputsb()`. The equivalent label in `UpdateVarBps()` does not, which is one of the six silent failures in #345 / PR #346. This PR does not depend on that one.

### DIFF
--- a/2.0/plink2.cc
+++ b/2.0/plink2.cc
@@ -546,6 +546,7 @@ typedef struct Plink2CmdlineStruct {
   TwoColParams* ref_allele_flag;
   TwoColParams* alt_allele_flag;
   TwoColParams* update_chr_flag;
+  TwoColParams* update_cm_flag;
   TwoColParams* update_map_flag;
   TwoColParams* update_name_flag;
 } Plink2Cmdline;
@@ -1018,6 +1019,11 @@ PglErr Plink2Core(const Plink2Cmdline* pcp, MakePlink2Flags make_plink2_flags, c
       if (unlikely(reterr)) {
         goto Plink2Core_ret_1;
       }
+      if (pcp->misc_flags & kfMiscZeroCms) {
+        // Same end state as loading a .pvar/.bim with no nonzero CM values;
+        // the writers already emit '0' when variant_cms is null.
+        variant_cms = nullptr;
+      }
       LoadFilterLogFlags load_filter_log_flags = pcp->load_filter_log_flags;
       if (load_filter_log_flags & kfLoadFilterLogImportMergeAlreadyApplied) {
         load_filter_log_flags &= ~kfLoadFilterLogImportMergeMask;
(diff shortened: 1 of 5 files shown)


chrchang:
--update-cms and --zero-cms are not mutually exclusive in PLINK 1.9, and using them together makes sense in some contexts.

BenjaminDEMAILLE:
Correct, and the restriction is now gone (`6c66b8b`... force-pushed).

I traced the 1.9 path: `--zero-cms` acts at load time (`plink_data.c:912` suppresses parsing of the CM column, leaving the calloc'd array), and `update_marker_cms()` runs afterwards at `plink.c:806`, so the combination means "zero everything, then set the ones named in the file". That is a useful operation, and I had it wrong.

plink2 now zeroes `variant_cms` in place when `--update-cm` is also present rather than dropping the array, so the update has somewhere to write; `--zero-cms` alone still drops it, and the existing calloc path covers the case where the input had no CM column at all.

Verified against 1.9 on an 1800-variant .bim with nonzero CM values and a `--update-cm` file naming 5 of them. CM columns are identical for `--zero-cms` alone, `--update-cm` alone, and both together, on inputs with and without a CM column, in either flag order. Non-trivially so: an unlisted variant keeps its old value under `--update-cm` alone and becomes 0 when `--zero-cms` is added.

`2.0/Tests/run_tests.sh` passes.

PR #348, opened 2026-08-31
Point PLINK 1.x flag names at their plink2 equivalents

BenjaminDEMAILLE:
From the audit in #344.

plink2 already emits "X is retired, use Y instead" for 26 PLINK 1.9 flags it still recognizes. Another 39 were renamed, or folded into a modifier or column set, so they are not recognized at all and fail with a bare `Unrecognized flag`. That is not much help when migrating a 1.9 command line.

```
$ plink2 --bfile base --recodeA --out x
Error: Unrecognized flag ('--recodeA'). (This is a PLINK 1.x flag; use
"--export A" instead.)
For more info, try "plink2 --help <flag name>" or "plink2 --help | more".
```

### Approach

A static table consulted **only** on the existing unrecognized-flag path, so the parser switch is untouched and nothing that currently parses can change behaviour. `InvalidArg()` moves from a header inline to `plink2_cmdline.cc` next to the table. Matching is case-insensitive, since PLINK 1.9 accepts both `--recodeA` and `--recodea`.

Unknown flags, and 1.9 flags whose functionality genuinely has no plink2 counterpart (`--homozyg`, `--cluster`, `--genome`, …), keep the original message unchanged.

### Entries

| 1.9 flag | hint |
|---|---|
| `--split-x` | use `--split-par` |
| `--update-freq` | use `--read-freq` |
| `--extract-snp` / `--no-snp` | use `--snp` / `--exclude-snp` |
| `--chr-excl` | use `--not-chr` |
| `--chr-output` | use `--output-chr` |
| `--must-have-sex` | use `--remove-nosex` |
| `--reference-allele`, `--update-ref-allele` | use `--ref-allele` |
| `--set-missing-snp-ids`, `--set-missing-nonsnp-ids` | use `--set-missing-var-ids` |
| `--q-score-file` | use `--q-score-range` |
| `--score-no-mean-imputation` | use the `--score 'no-mean-imputation'` modifier |
| `--gc`, `--log10` | use the `--adjust`/`--adjust-file` modifier of the same name |
| `--qq-plot` | use `--adjust cols=+qq` |
| `--beta` | use `--glm cols=+beta` |
| `--dominant`, `--recessive`, `--genotypic`, `--hethom`, `--interaction`, `--hide-covar`, `--no-x-sex` | use the `--glm` modifier of the same name |
| `--all-pheno` | plink2 processes every loaded phenotype by default |
| `--tab` | plink2's `--export` output is tab-delimited by default |
| `--transpose` | use `--export tped` |
| `--with-freqs` | use `--r2-unphased cols=+freq` |
| `--recode12`, `--recodeA`, `--recodeAD`, `--recode-vcf` | use `--export ped 12` / `A` / `AD` / `vcf` |
| `--recodeHV`, `--recode-beagle`, `--recode-bimbam`, `--recode-fastphase`, `--recode-lgen`, `--recode-rlist`, `--recode-structure` | plink2 has no such output yet |

### Verification

Every target was run against the binary, not just read out of the help text. That check caught a case worth flagging: `--export` **accepts** the format names `23, beagle, beagle-nomap, bimbam, bimbam-1chr, fastphase, fastphase-1chr, HV, HV-1chr, lgen, lgen-ref, list, rlist, structure` and then aborts with

```
Error: Only VCF, BCF, oxford, bgen-1.x, haps, hapslegend, A, AD, Av, ped, tped,
compound-genotypes, phylip, phylip-phased, eig, eigt, and ind-major-bed output
have been implemented so far.
```

so the seven `--recode-*` entries for those formats say plink2 has no equivalent yet rather than pointing users at a dead end. (I had this wrong in the first version of #344 and have posted a correction there.)

All 39 hints printed and checked; `--nonsense-flag`, `--homozyg`, `--cluster` still produce the plain message. `2.0/Tests/run_tests.sh` passes 8/8. Build is warning-free.

### DIFF
--- a/2.0/plink2.cc
+++ b/2.0/plink2.cc
@@ -3487,6 +3487,94 @@ static_assert(!kChrOffsetX, "--autosome-num/--chr-set/--cow/etc. assume kChrOffs
 static_assert(kChrOffsetY == 1, "--chr-set/--cow/... assume kChrOffsetY == 1.");
 static_assert(kChrOffsetXY == 2, "--chr-set/--cow/... assume kChrOffsetXY == 2.");
 static_assert(kChrOffsetMT == 3, "--chr-set/--cow/... assume kChrOffsetMT == 3.");
+
+// Sorted by strcmp() so the lookup below can binary-search.  Matching is
+// case-sensitive, like plink2's own flag parser and like PLINK 1.x's; the few
+// 1.x flags with two accepted spellings (--recodeA/--recodea and friends) have
+// an entry each.
+typedef struct Plink1FlagHintStruct {
+  const char* flagname;
+  const char* hint;
+} Plink1FlagHint;
+
+static const Plink1FlagHint kPlink1FlagHints[] = {
+  {"all-pheno", "--glm processes every loaded phenotype by default"},
+  {"beta", "use \"--glm cols=+beta\" instead"},
+  {"chr-excl", "use --not-chr instead"},
+  {"chr-output", "use --output-chr instead"},
+  {"dominant", "use the --glm 'dominant' modifier instead"},
+  {"extract-snp", "use --snp instead"},
+  {"gc", "use the --adjust/--adjust-file 'gc' modifier instead"},
+  {"genotypic", "use the --glm 'genotypic' modifier instead"},
+  {"hethom", "use the --glm 'hethom' modifier instead"},
+  {"hide-covar", "use the --glm 'hide-covar' modifier instead"},
+  {"interaction", "use the --glm 'interaction' modifier instead"},
+  {"log10", "use the --adjust/--adjust-file 'log10' modifier instead"},
+  {"must-have-sex", "no plink2 equivalent; it blanked the phenotypes of unknown-sex samples on output, whereas --remove-nosex drops those samples entirely"},
+  {"no-snp", "no plink2 equivalent; this was a --linear/--logistic modifier that regressed on the covariates alone"},
+  {"no-x-sex", "use the --glm 'no-x-sex' modifier instead"},
+  {"q-score-file", "use --q-score-range instead"},
+  {"qq-plot", "use \"--adjust cols=+qq\" instead"},
+  {"recessive", "use the --glm 'recessive' modifier instead"},
+  {"recode-beagle", "plink2 has no BEAGLE output yet"},
+  {"recode-bimbam", "plink2 has no BIMBAM output yet"},
+  {"recode-fastphase", "plink2 has no fastPHASE output yet"},
+  {"recode-lgen", "plink2 has no .lgen output yet"},
+  {"recode-rlist", "plink2 has no rlist output yet"},
+  {"recode-structure", "plink2 has no STRUCTURE output yet"},
+  {"recode-vcf", "use \"--export vcf\" instead"},
+  {"recode12", "use \"--export ped 12\" instead"},
+  {"recodeA", "use \"--export A\" instead"},
+  {"recodeAD", "use \"--export AD\" instead"},
+  {"recodeHV", "plink2 has no Haploview output yet"},
+  {"recodea", "use \"--export A\" instead"},
+  {"recodead", "use \"--export AD\" instead"},
+  {"recodehv", "plink2 has no Haploview output yet"},
+  {"reference-allele", "use --alt1-allele instead; PLINK 1.x's --reference-allele set A1, which corresponds to ALT1 here, not REF"},
+  {"score-no-mean-imputation", "use the --score 'no-mean-imputation' modifier instead"},
+  {"set-missing-nonsnp-ids", "use --set-missing-var-ids instead"},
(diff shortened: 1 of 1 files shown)


chrchang:
* Several of these hints are inaccurate.
* The MatchUpper{Counted,KLen,K}() or strcaseequal() function in include/plink2_string can perform the case-insensitive string comparison.
* Since the lookup table is sorted, a binary search is more appropriate than brute-force comparison against all keys.

BenjaminDEMAILLE:
Fixed all three; force-pushed as `cc48db5`.

**Inaccurate hints.** You were right that I had not actually checked these. My "verified against the binary" claim covered only that the *target* flag exists, not that the 1.x flag means what I said it means. Auditing all 39 against `plink --help <flag>` and the 1.9 sources turned up four wrong and two imprecise:

- **`--reference-allele` and `--update-ref-allele` were backwards.** `plink_help.c:1790` groups them as aliases of `--a1-allele`, and plink2 rewrites `--a1-allele` to `--alt1-allele` (it is `--a2-allele` that becomes `--ref-allele`). I had pointed both at `--ref-allele`, so anyone following the hint would have assigned the wrong allele. Now `--alt1-allele`, with the reason spelled out in the hint text.
- **`--no-snp` is not a variant filter.** It is a `--linear`/`--logistic` modifier that regresses on the covariates alone (`plink_help.c:1001`). I had it as `--exclude-snp`, a completely different operation. Now reported as having no plink2 equivalent.
- **`--must-have-sex` does not remove samples.** It forces ambiguous-sex *phenotypes* to missing on `--make-bed`/`--recode`/`--write-covar`. `--remove-nosex` drops the samples entirely. Now reported as having no equivalent, with that distinction stated.
- **`--tab`** makes 1.x output "mostly tab-delimited"; a space remains between the two alleles of a genotype. plink2's `--export ped` is tab-delimited throughout, which is 1.x's `tabx`. I checked the actual output of all three. The hint now says so instead of implying they match.
- **`--all-pheno`** is scoped to `--glm` now, since that is where plink2's "all loaded phenotypes by default" behavior applies.

I then re-derived every remaining target by running it, not by reading help text: the `--glm` dominant/recessive/genotypic/hethom/hide-covar/no-x-sex modifiers and `cols=+beta`, `--adjust gc`/`log10`/`cols=+qq`, `--r2-unphased cols=+freq`, `--export ped 12`/`tped`/`A`/`AD`/`vcf`, `--alt1-allele`, `--snp`, `--not-chr`, `--output-chr`, `--split-par`, `--read-freq`, `--q-score-range`, `--score no-mean-imputation`, `--set-missing-var-ids`. All run to completion.

**Case comparison and binary search.** I dropped the case-insensitive comparison entirely rather than swapping in `MatchUpper*`/`strcaseequal`. `MatchUpperCounted()` requires the fixed string to be letters only, which these keys are not, and `strcaseequal()` gives equality but not ordering, so a case-insensitive binary search would still have needed a hand-rolled fold. Since plink2's own flag parser is case-sensitive and so is 1.x's, matching that is the consistent choice: the table is sorted by `strcmp()`, the lookup is a binary search, and the handful of 1.x flags with two accepted spellings (`--recodeA`/`--recodea`, `--recodeAD`/`--recodead`, `--recodeHV`/`--recodehv`) get an entry each. 42 entries now. If you would rather keep case-insensitive matching, say so and I will fold the query and keep the search.

`2.0/Tests/run_tests.sh` passes 8/8, and output is byte-identical to master for `--freq`, `--make-bed`, `--glm` and `--indep-pairwise`.

2.0/plink2_cmdline.cc
```diff
@@ -4506,6 +4506,95 @@ PglErr ParseColDescriptor(const char* col_descriptor_iter, const char* supported
   return reterr;
 }
 
+// PLINK 1.x flags which plink2 renamed, or folded into a modifier or column
```
chrchang: These functions are a better fit for the current plink2.cc .  plink2_cmdline.cc is designed to fit into other programs which want plink-like command-line parsing and memory allocation; it should avoid capturing details of plink2 flags which are irrelevant to all other programs.

(It would of course be reasonable to split plink2.cc into multiple pieces in a future PR.)

chrchang: You could leave the existing InvalidArg() untouched, and define a new InvalidPlink2Arg() function in plink2.cc.

BenjaminDEMAILLE: Moved. The table, `Plink1MigrationHint()` and a new `ReportInvalidFlag()` now live in plink2.cc next to the flag parser, all three file-local; `plink2_cmdline.cc` and `plink2_cmdline.h` are back to their master state, `InvalidArg()` included. `ReportInvalidFlag()` falls back to `InvalidArg()` when there is no hint, so the generic path stays in the reusable layer and only the plink2-flag-specific part moved out of it.

The whole PR is now a 89-line addition to plink2.cc and nothing else. `2.0/Tests/run_tests.sh` passes.

PR #349, opened 2026-08-31
Add --23file (23andMe import)

BenjaminDEMAILLE:
From the parity audit in #344: 23andMe input was one of the read/write asymmetries, and the one with the least ambiguity about intended behaviour, since PLINK 1.9's `bed_from_23()` is a complete specification.

```
--23file <fname> [FID] [IID] [sex] [pheno] [pat. ID] [mat. ID]
```

`TwentythreeToPgen()` follows 1.9 exactly on semantics, but writes `.pgen`/`.pvar`/`.psam` and uses plink2 chromosome codes.

### Semantics carried over from 1.9

- `sex` is `M`/`1`, `F`/`2`, `I` (infer from data, the default), or `0` (force missing).
- A single allele call is accepted only on chrX, where it forces male sex, and on fully-haploid chromosomes; anywhere else it is "fewer allele calls than expected".
- chrXY requires two calls.
- A nonmissing chrY call forces male sex, and is an error when female was forced.
- Chromosome codes must be non-decreasing; unplaced (code `0`) lines do not participate in that check.
- `D`/`I` calls get the complementary indel code as their second allele, so a homozygous `D` or `I` call still has two distinct codes (1.9's `cc ^ 13` trick).

### Two plink2-specific choices

- The observed call becomes REF and the second distinct call becomes ALT. That is exactly the A2/A1 assignment 1.9 writes to its `.bim`, so `--make-bed` reproduces 1.9's `.bed` and `.bim`.
- Variants with negative bp are skipped, matching the other plink2 legacy importers. 1.9 keeps them. Happy to change this if you would rather match 1.9 here.

### Verification

macOS/arm64, `2.0/Makefile` dev build, against `PLINK v1.9.0-b.7.11`.

**Hand-built 14-variant file** covering missing calls, hets, homozygotes, `DD`/`DI`/`II`, chrX haploid, chrY, chrXY, chrMT and unplaced:

```
$ cmp p19.bed b2.bed && echo IDENTICAL
IDENTICAL
```

`.bim`, 1.9 on the left, plink2 on the right:

```
1  rs1  0  100  0  A   |  1   rs1  0  100  .  A
1  rs2  0  200  G  A   |  1   rs2  0  200  G  A
1  rs3  0  300  0  0   |  1   rs3  0  300  .  .
1  rs12 0  500  I  D   |  1   rs12 0  500  I  D
23 rs6  0 1000  0  A   |  X   rs6  0 1000  .  A
25 rs9  0  100  0  T   |  XY  rs9  0  100  .  T
26 rs10 0   50  0  A   |  MT  rs10 0   50  .  A
0  rs11 0    0  0  C   |  0   rs11 0    0  .  C
```

The only differences are plink2's existing conventions (`.` for a missing allele, `X`/`Y`/`XY`/`MT` for 23/24/25/26). `.fam` differs only in tabs vs spaces, which `2.0/Tests` already documents for every importer.

**9280-variant randomized file** across all 26 chromosomes, 5% missing calls, 2% indels: `.bed` byte-identical; all 9280 `.bim` rows identical after mapping `23/24/25/26 → X/Y/XY/MT` and `0 → .`; same indel count (150) and same inferred sex.

**`--chr 1` subset**: `.bed` byte-identical, 7 variants on both sides.

**Error parity**, identical messages on both:

| case | message |
|---|---|
| 3 allele calls | `Line 1 of e1.txt has more allele calls than expected.` |
| out-of-order chromosomes | `Chromosomes in e2.txt are out of order.` |
| missing tokens | `Line 1 of e3.txt has fewer tokens than expected.` |
| single call on an autosome | `Line 1 of e4.txt has fewer allele calls than expected.` |
| single call on chrXY | `Line 1 of e5.txt has fewer allele calls than expected.` |
| invalid chromosome code | `Invalid chromosome code on line 1 of e6.txt.` |
| empty file | `No --23file variants.` |
| comments-only file | `No --23file variants.` |

All four sex arguments behave identically, including `F` erroring on a haploid chrX call. The male-sample sanity warning fires in the same situation. `--keep-autoconv` writes `.pgen` + `.pvar` + `.psam`; otherwise no temporary files are left behind.

`2.0/Tests/run_tests.sh` passes 8/8. Build is warning-free.


chrchang:
I usually implement import and export commands for a file format around the same time, since implementing the other direction is easier when all the file-format details have been freshly reviewed, and it makes round-trip testing easy.  Both of these considerations are relatively unimportant here -- Claude doesn't have the same mental limitations we have, and the lack of PLINK 2-specific functionality here means that round-trip tests don't add much value on top of comparison against PLINK 1.9, so this is your call.

BenjaminDEMAILLE:
Took your suggestion: `--export 23` is now in the PR (`50e3995`).

It turned out not to be optional in the way I assumed. `--export` already accepted `23` as a format name, but it fell through to "have been implemented so far", so before this the format could be read and not written. So the pair really was missing a direction rather than having one for free.

The writer follows 1.9's `--recode 23`: one line per variant, `<ID>\t<chromosome>\t<bp>\t<genotype>` after 1.9's header block; two allele characters with REF first at heterozygous calls, `--` when missing; a single character for a homozygous call on a haploid chromosome; exactly one sample and single-character allele codes required, with the same two errors 1.9 raises.

The haploid rule is the fiddly part, and reviewing the format details while writing the exporter is what surfaced it: 1.9 treats chrY and chrMT as haploid always, but chrX as haploid only for a male sample **with no heterozygous chrX call anywhere in the data** — one het chrX call and the whole chromosome falls back to two characters. I reproduced that, at the cost of a pass over chrX. It shows up immediately in testing: my 9280-variant test file has 39 het chrX calls, so all 60 of its single-character chrX genotypes come back as two characters, in both programs identically.

Verification: output byte-identical to `plink --recode 23` on that 9280-variant file, and on a 730-variant file spanning chr1, chr2, chrX, chrY and chrMT for a male sample, covering the single-character haploid calls and the missing ones. And you were right that round-trip testing is easy once both directions exist: `--23file` then `--export 23` is a fixed point on the latter file, reproducing it exactly. `2.0/Tests/run_tests.sh` passes.

2.0/plink2.cc
```diff
@@ -4092,6 +4095,71 @@ int main(int argc, char** argv) {
...
+                  goto main_ret_1;
+              }
+              if (param_ct > 3) {
+                const char* cur_modif = argvk[arg_idx + 4];
+                if (unlikely(strlen(cur_modif) != 1)) {
+                  goto main_ret_23FILE_INVALID_SEX;
+                }
+                const char cc = cur_modif[0];
+                if ((cc == 'M') || (cc == 'm') || (cc == '1')) {
+                  pc.twenty_three_info.sex_mode = 1;
```
chrchang: Best to define an enum type in plink2_import.h for this.

2.0/plink2.cc
```diff
@@ -4092,6 +4095,71 @@ int main(int argc, char** argv) {
...
+                } else if ((cc == 'F') || (cc == 'f') || (cc == '2')) {
+                  pc.twenty_three_info.sex_mode = 2;
+                } else if (cc == '0') {
+                  pc.twenty_three_info.sex_mode = 3;
+                } else if (unlikely((cc != 'I') && (cc != 'i'))) {
+                  goto main_ret_23FILE_INVALID_SEX;
+                }
+                if (param_ct > 4) {
+                  const char* pheno_str = argvk[arg_idx + 5];
+                  if (unlikely(!ScantokDouble(pheno_str, &pc.twenty_three_info.pheno))) {
```
chrchang: This restriction can be relaxed.  The .psam format supports categorical phenotype values; most strings which don't start with a digit (optionally preceded by '+'/'-' and/or '.') are allowed in that setting.  See IsCategoricalPhenostrNocsv() in plink2_common.cc , which is called during .psam parsing to determine whether to treat a phenotype column as categorical instead of numeric.

2.0/plink2.cc
BenjaminDEMAILLE: Done: TwentythreeSexMode, defined next to TwentythreeInfo. That struct lives in plink2_import_legacy.h rather than plink2_import.h, so I put the enum there; say the word if you'd rather have it in plink2_import.h.

2.0/plink2.cc
BenjaminDEMAILLE: Relaxed. The phenotype parameter is now kept as a string and passed through to PHENO1 unchanged whenever `IsCategoricalPhenostrNocsv()` accepts it; anything else still has to parse as a number, so `1abc` is still rejected. Keeping it as a string also means a number is written exactly as typed rather than round-tripped through `dtoa_g()`.

`--23file f.txt F1 I1 M case` now produces `PHENO1=case`, and reloading that .psam reports "1 categorical phenotype loaded (1 value)."

BenjaminDEMAILLE:
On the import/export pairing: I will leave --23file import-only. PLINK 1.9 has no 23andMe export either, and the format is single-sample with a fixed column layout, so an export would only ever be a round-trip target rather than something a user wants; the value is in the comparison against 1.9, which this PR already does. If you would rather have both directions for the file-format-review reason, I will add it.

Both inline comments are addressed and pushed.

2.0/plink2.cc
chrchang: oops, you're right, I meant plink2_import_legacy.h .

chrchang:
> On the import/export pairing: I will leave --23file import-only. PLINK 1.9 has no 23andMe export either,

This comment confuses me, since it contradicts the correct earlier observation that PLINK 1.9 does have working `--recode 23`.

BenjaminDEMAILLE:
That comment was wrong twice over, and I should not have written it. PLINK 1.9 does have `--recode 23`, as I had said correctly earlier, and this PR already implements the export direction: commit 50e3995a, "Add --export 23 (23andMe output)". I answered a question about work I had already done, from memory, without looking. Disregard it.

For the record, since the claim is what matters rather than the retraction: `--export 23` output on the test fileset is identical to 1.9 `--recode 23` line for line, and `--23file` -> `--export 23` -> `--23file` reproduces the .pgen byte for byte.

Enum location noted, it is in plink2_import_legacy.h.

2.0/plink2_export_legacy.cc
```diff
@@ -423,6 +423,179 @@ PglErr ExportIndMajorBed(const uintptr_t* orig_sample_include, const uintptr_t*
...
+        *chr_name_end++ = '\t';
+        chr_blen = chr_name_end - chr_buf;
+        haploid_out = (chr_idx == y_code) || (chr_idx == mt_code) || (is_male && (chr_idx == x_code) && (!x_hethap));
+      }
+      uintptr_t allele_idx_offset_base = variant_uidx * 2;
+      if (allele_idx_offsets) {
+        allele_idx_offset_base = allele_idx_offsets[variant_uidx];
+        if (unlikely(allele_idx_offsets[variant_uidx + 1] != allele_idx_offset_base + 2)) {
+          logputs("\n");
+          logerrputs("Error: \"--export 23\" cannot be used with multiallelic variants.\n");
```
chrchang: It is straightforward for this to support multiallelic variants.

BenjaminDEMAILLE: Done. `Export23()` now reads through `PgrGetM()`, so the ALT allele codes are available: an ALTx/ALTy call writes both characters, an ALTx/ALTx call writes that one twice, and the "cannot be used with multiallelic variants" bail-out is gone. Multi-character allele codes are still rejected, since the format has no room for them.

The chrX prescan needed the same change and turned out to be the more interesting half. It decides whether a male's chrX genotypes go out as one character or two, and it was only looking for the `0/ALT1` encoding, so an `ALT1/ALT2` heterozygote was invisible and chrX was written as haploid anyway. A patch_10 entry can be either ALTx/ALTy or ALTx/ALTx, so it now compares the two allele codes instead of just checking `patch_10_ct`.

Also added `Tests/TEST_23FILE`, which was missing entirely:

- `--23file` import against PLINK 1.9: identical .bed, and identical .bim/.fam once plink2's chromosome names, `.` missing-allele code and tab delimiters are normalized
- `I` sex inference reaching the same conclusion
- `--export 23` against PLINK 1.9 `--recode 23`: byte-identical body
- an export/re-import round trip
- the multiallelic output, spelled out, since PLINK 1.9 cannot produce it:

```
rm1	1	1	GT     (1/2)
rm2	1	2	TT     (2/2)
rm3	1	3	AT     (0/2)
rmx1	X	3000000	GG     (male chrX, two characters because of rmx2)
rmx2	X	3000001	GT     (1/2)
rmy1	Y	3000000	T      (2/2, one character)
rmy2	Y	3000001	GT     (1/2 het haploid, two characters)
```

BenjaminDEMAILLE:
Rebased onto master (through e78ce815).

2.0/plink2_export_legacy.cc
```diff
@@ -423,6 +423,194 @@ PglErr ExportIndMajorBed(const uintptr_t* orig_sample_include, const uintptr_t*
...
+          genotext[0] = cur_alleles[ac0][0];
+          genotext[1] = cur_alleles[ac1][0];
+          is_het = (ac0 != ac1);
+        }
+        if (haploid_out && (!is_het)) {
+          geno_slen = 1;
+        }
+      }
+      fputs(variant_ids[variant_uidx], outfile);
+      char writebuf[64];
```
chrchang: This is an unlikely but possible buffer overflow.  plink2 allows chromosome/variant/sample ID components to be as long as 16000 characters, so g_textbuf (~262k bytes) is safe but 64 bytes is not.

BenjaminDEMAILLE: Fixed, and you are right that it is reachable rather than theoretical: `--export 23` writes whatever chromosome code the dataset carries, so with `--allow-extra-chr` and a long contig name the row is unbounded. I reproduced it with a 1600-character contig name, which writes a 1608-byte line into what was a 64-byte stack buffer.

Switched to `g_textbuf`. `Tests/TEST_23FILE` now exports that fileset and checks all four fields come out intact.

2.0/plink2.cc
BenjaminDEMAILLE: Good, it is already there — `TwentythreeSexMode` sits next to `TwentythreeInfo` in `plink2_import_legacy.h`. Nothing further needed here.

PR #350, opened 2026-08-31
Add --ibc (GCTA Fhat1/Fhat2/Fhat3)

BenjaminDEMAILLE:
From the parity audit in #344. `--het` reports the method-of-moments F, which is a different quantity (a ratio of sums rather than a mean of per-variant ratios), so `--ibc` adds numbers plink2 did not previously provide.

### Estimators

With `p` the REF allele frequency and `x` the sample's REF allele count at a variant:

| | per-variant term |
|---|---|
| Fhat1 | `(x - 2p)^2 / (2p(1-p))` |
| Fhat2 | `2`, except `2 - 1/(2p(1-p))` for a heterozygous call |
| Fhat3 | `1 + (1-p)/p` for `x=2`, `0` for `x=1`, `1 + p/(1-p)` for `x=0` |

Each reported value is the mean of its terms over the sample's nonmissing calls, minus 1. All three are invariant under swapping the reference allele, so REF/ALT orientation does not matter. I derived these from 1.9's `update_rel_ibc()` weight tables (`ibc_type` 1, 2, 0 respectively) and checked each against the GCTA paper's formulas by hand before implementing.

Like `--het`, this restricts itself to autosomal variants (1.9 does the same, printing `Excluding N variants on non-autosomes from relationship matrix calc`) and rejects haploid genomes with the same message.

### Two deliberate differences from 1.9

- **Monomorphic variants are skipped**, with the count reported. 1.9 substitutes special weights and keeps them in the denominator. Skipping is what plink2's `--het` already does, and the estimators are undefined there. This is the only source of disagreement I found; it is quantified below.
- **Multiallelic variants are skipped**, with a count, since GCTA's estimators are biallelic.

### Verification

macOS/arm64, `2.0/Makefile` dev build, against `PLINK v1.9.0-b.7.11`. "max |d|" is the largest absolute difference over all samples.

| dataset | result |
|---|---|
| 1000 samples x 1800 autosomal variants, 3% missing | `n=1000 max|d| OBS_CT=0 FHAT1=0 FHAT2=0 FHAT3=0` |
| same extended to 2500 variants incl. chrX/Y/MT | `n=1000 max|d| OBS_CT=0 FHAT1=0 FHAT2=0 FHAT3=0` |
| 2504 real 1000 Genomes chr21 samples, polymorphic variants only (597) | `n=2504 max|d| OBS_CT=0 FHAT1=0 FHAT2=0 FHAT3=0` |
| same, with the 3 monomorphic variants left in | `OBS_CT` differs by exactly 3; estimates differ accordingly |
| `--keep` on a 200-sample subset | `n=200 max|d| ... =0` |

Exact agreement on 2504 real samples and 1000 simulated ones.

Hand-checkable case: a 4-sample fileset at `p=0.5` gives `1` for homozygotes and `-1` for heterozygotes on all three estimators, which is what the formulas require, and `OBS_CT 0` with `NA` for a sample whose calls are all missing.

`zs`, `cols=` subsetting, the bad-column-name error, and the haploid-genome rejection all behave like `--het`'s equivalents. Output is byte-identical to a `master` build for `--het`, `--freq`, `--make-bed`, `--hardy`, `--missing` and `--pca approx`. `2.0/Tests/run_tests.sh` passes 8/8; build is warning-free.

### Note

The accumulation loop is single-threaded and unvectorized, unlike `--het`'s kernel. Fine for the sizes `--ibc` is normally used on, but the obvious thing to optimize if you take this.


BenjaminDEMAILLE:
Pushed a performance commit (`7bd86c5`), which addresses the "single-threaded and unvectorized" note at the end of the description. It is now slightly faster than PLINK 1.9's `--ibc`, still single-threaded.

The inner loop updated four separate per-sample arrays (three double sums plus the observation count), so every variant streamed four independent regions of memory and branched three ways on the genotype. Two changes:

- The four arrays become one record per sample, so the loop walks a single stream.
- The per-genotype weights go into 4-entry tables with the missing slot zeroed, removing the branches. Adding 0.0 leaves a finite running sum bit-for-bit unchanged, so the accumulation order is exactly what the previous if/else chain produced. That mattered here: the PR's claim is exact agreement with 1.9, so I did not want any reassociation.

5000 samples x 50000 variants, warm cache, four consecutive runs each:

```
before      0.91 0.64 0.64 0.64
after       0.24 0.24 0.24 0.24
PLINK 1.9   0.27 0.27 0.27 0.27
```

For the record I did consider the larger rewrite: Fhat2 can be reduced to `2*nobs - sum over het variants of 1/(2p(1-p))`, which makes that accumulator sparse, and similar factorings exist for the other two. I did not do it because it changes the summation order and would have cost the exact agreement with 1.9 that the PR is built on. If you would rather have the speed, it is available.

Output is unchanged: byte-identical `.ibc` to the previous version, and the agreement with 1.9 still holds exactly (max absolute difference 0 on all three estimators) on the simulated autosomal fileset, its mixed-chromosome variant, a 200-sample subset, and the 2504 real 1000 Genomes samples. `2.0/Tests/run_tests.sh` passes 8/8.

chrchang:
My main question here is, what inbreeding statistics are researchers most interested in today?  I prioritized implementation of --het over --ibc because I was under the impression that there isn't much interest in the extra --ibc stats.

BenjaminDEMAILLE:
Fair question, and I cannot give you usage numbers, so here is what I can support with measurements plus what I would guess.

**--het's F is not Fhat2.** That is the part I would not have assumed before checking. `--het` reports `F = (ΣO(HOM) - ΣE(HOM)) / (OBS_CT - ΣE(HOM))`, a ratio of sums, whereas GCTA's Fhat2 is a mean of per-variant terms each weighted by `1/(2p(1-p))`, so rare variants carry far more weight. On a 2504-sample 1000 Genomes chr21 subset the two disagree by up to 1.40 in absolute value, and per sample the two rank differently. So `--ibc` is not three columns of which one duplicates `--het`; it is three estimators none of which `--het` currently provides.

**Which of the three still gets used.** My read is that Fhat3 (correlation between uniting gametes) is the one with continuing use, because it is the least sensitive to allele-frequency misspecification of the three, and Fhat1 is the one rarely reported on its own. Fhat2 is the closest in spirit to `--het` and is what people usually mean by "excess homozygosity". But the honest answer to "what are researchers most interested in today" is F_ROH, i.e. `--homozyg`, not any of the Fhat estimators, so if there is a single inbreeding gap worth filling in plink2 I would rank #354 above this PR.

That suggests a smaller alternative, if you want one: add Fhat3 as an optional `--het cols=` column rather than a separate `--ibc` command, which costs one column set and no new command, and drop Fhat1/Fhat2. I am happy to rework the PR that way, leave it as the full GCTA-compatible `--ibc`, or close it. Your call.

**Separately, I found and fixed a real bug in this PR's first version.** It skipped variants with an allele frequency of 0 or 1, on the assumption that 1.9 emitted infinities there. It does not: `update_rel_ibc()` (`plink_calc.c:143`) gives the one homozygous genotype that can actually occur a finite weight, so a monomorphic variant still counts toward NOMISS while contributing 0 to all three estimates, and the infinities sit only in the impossible genotype slots, where they surface if a `--read-freq` file's zero MAFs contradict the data. Skipping the variant shrank OBS_CT and rescaled everything, and additionally shifted Fhat2, because 1.9 deliberately uses weight 1 rather than 2 at those variants "for agreement with GCTA". On the chr21 subset (600 variants, 3 monomorphic) that gave FHAT1 0.200004 against 1.9's 0.194004.

Now fixed, and output is identical to 1.9 on: all 2504 samples of that subset, a 1000-sample 1800-variant set, `--maf` and `--chr` filtered runs, a 60-sample set with both missing calls and monomorphic variants, and the inconsistent `--read-freq` case, where both emit `inf` in the same 46 places. `2.0/Tests/run_tests.sh` passes.

Remaining known gap either way: multiallelic variants are skipped with a log line, since GCTA's estimators are defined for biallelic variants, whereas `--het` handles them.

chrchang:
Implementation of the most relevant multiallelic-variant-supporting statistic as an optional --het column-set is probably the best way to go.  I will look into whether any of these statistics naturally generalizes to multiallelic variants (note that the GRM calculation behind --pca *does* have such a generalization, and that generalization was implemented in PLINK 2 in 2019).

BenjaminDEMAILLE:
On which of the three generalizes: it is **Fhat1**, and the generalization is the one you already have.

Fhat1 is the variance-standardized relationship of a sample with itself, minus 1, i.e. exactly the GRM diagonal minus 1. Checked rather than asserted: on the 2504-sample 1000 Genomes chr21 subset, `--make-rel square bin`'s diagonal minus 1 matches this PR's FHAT1 column to within 4.4e-6 relative across all 2504 samples (mean signed difference 1.5e-8), which is the same statistic up to summation order and the .ibc column's print precision. So its multiallelic generalization is whatever `LoadMultiallelicCenteredVarmaj()` does: one variance-standardized row per non-major allele. Nothing new needs deriving.

The other two do not generalize as cleanly, because both are written in terms of a single ALT dosage x:

- Fhat2 = `1 - (1/N)Σ x(2-x)/(2p(1-p))` is excess homozygosity, and its natural multiallelic form is observed vs expected homozygosity with expected = `1 - Σ p_i²`. That is what `--het` already computes, and `--het` already says it supports multiallelic variants. So this one is not a gap at all.
- Fhat3 = `(1/N)Σ [x²-(1+2p)x+2p²]/[2p(1-p)]` is the correlation between uniting gametes, and x² has no single-allele reading once there are three or more alleles. I do not see a generalization that isn't arbitrary.

So the gap `--het` actually has is Fhat1, and it is the one with a ready multiallelic definition.

One caveat worth stating before it becomes a column: `--het`'s F and Fhat1 are not close on real data. On that chr21 subset the two differ by up to 1.40 in absolute value and rank samples differently, because Fhat1 weights each variant by `1/(2p(1-p))` while `--het`'s F is a ratio of summed observed and expected homozygote counts. Rare variants dominate Fhat1 and barely move F. That is the argument for having both, and also the reason Fhat1 is the more allele-frequency-sensitive of the two.

Happy to rework this PR into `--het cols=` with a single Fhat1 column built on the existing GRM code path, dropping Fhat2 and Fhat3 and the separate `--ibc` command, if that is the shape you want. Say the word on the column name and I will push it.

chrchang:
When trying to generalize these formulas to multiallelic variants, we need to preserve two properties:
* The expected value under Hardy-Weinberg-equilibrium frequencies must always be 1.
* The formula must reduce to the GCTA formula when the number of alleles is reduced to 2.

Let's look at Fhat3 first since that's the most used.  It's 0 for heterozygous genotypes, and some fancier value for homozygous genotypes.

What happens if we write that fancy value in terms of the frequency of the allele that the genotype contains two copies of?  I.e. `1 + (1 - a_i)/a_i`, where i=0 corresponds to REF, i=1 corresponds to ALT1, up to i=n-1 when there are n alleles.  This does correspond to the GCTA formula when there are only two alleles.  And when there are more, the frequency of each homozygous genotype under Hardy-Weinberg equilibrium is `a_i^2`, so the expected value of the statistic is

`\sum_i=0^{n-1} a_i^2 (1 + (1 - a_i)/a_i) = \sum_i=0^{n-1} [a_i^2 + a_i(1 - a_i)] = \sum_i=0^{n-1} [a_i^2 + a_i - a_i^2] = \sum_i=0^{n-1} a_i = 1`

so this generalization actually works.

BenjaminDEMAILLE:
Your Fhat3 generalization is exactly the form this PR already computes for the biallelic case, which is a useful cross-check: the weight table here is

```c
f3_tab[0] = 1.0 + (1.0 - ref_freq) / ref_freq;   // hom REF
f3_tab[1] = 0.0;                                 // het
f3_tab[2] = 1.0 + ref_freq / (1.0 - ref_freq);   // hom ALT
```

i.e. `1/a_i` for a homozygote of allele i and 0 for a heterozygote, reported as `mean - 1`. So `1 + (1 - a_i)/a_i` is not a new formula that needs validating against GCTA; it is the existing one rewritten in terms of the homozygous allele, and it already matches PLINK 1.9 field for field on my test data. Extending it is then just "which allele is the sample homozygous for", with ALTx/ALTy heterozygotes contributing 0 like any other heterozygote.

Applying the same two properties to the other two:

**Fhat2** generalizes the same way. Write E_het for the HWE heterozygosity, `1 - sum_i a_i^2`. Then hom -> 2, het -> `2 - 1/E_het` gives

`E = P(het) * (2 - 1/E_het) + P(hom) * 2 = 2 - E_het/E_het = 1`

and at two alleles `E_het = 2pq`, so the het weight is `2 - 1/(2pq)`, which is the current table. Same shape as Fhat3: only the heterozygous slot needs the allele count, and it needs only the sum of squared frequencies rather than which allele is involved.

**Fhat1** is the one I would not write from scratch. Its biallelic weight is `(y - 2*a_ref)^2 / (2*a_ref*(1 - a_ref))` with y the REF copy count, i.e. a GRM diagonal entry, so the right generalization is whatever the 2019 --make-grm work already settled on rather than a fourth derivation.

That makes Fhat3 the natural single addition, as you suggested. Happy to rework this into a `fhat3` column set on --het and drop --ibc entirely; the only thing I want to check before touching `HetThread()` is whether you want it folded into the existing base-plus-increment scheme (the per-variant base would be `1/a_ref`, with the difflist supplying exactly the samples that need a correction, so it should fit without a second pass over the genotypes) or kept as a separate accumulator for clarity. That is a decision about your hot loop, so I would rather ask than guess.

chrchang:
Regarding Fhat1, let `x_i` be the number of copies of allele i that the sample has.  Then, you can write twice the numerator as `\sum_i=0^{n-1} (x_i - 2a_i)^2`, which can be interpreted as a squared Euclidean distance from the observed mean.  Similarly, you can write twice the denominator as `\sum_i=0^{n-1} 2a_i(1 - a_i)`.  Under multiallelic Hardy-Weinberg equilibrium, the relative frequencies of x_i=0, x_i=1, and x_i=2 satisfy the same relationship they do for biallelic HWE, so the denominator remains the expected value of the numerator under HWE.

So all three --ibc stats actually do generalize cleanly to multiallelic variants, and their behavior on datasets containing many multiallelic variants hasn't really been explored.  In light of this, I think the best call is to implement all three stats as optional --het column-sets.  --het is fast enough that there's no problem with slowing down the base case by <5% to cleanly support these three optional stats.

BenjaminDEMAILLE:
Rebased onto master (through e78ce815); --ibc takes command bit 35, since --ld-score claimed 33 upstream.

BenjaminDEMAILLE:
Rebased onto master (through e78ce815); --ibc takes command bit 35, since --ld-score claimed 33 upstream.

Two follow-ups since:

- The help entry claimed monomorphic variants are skipped because the estimators are undefined there, and that PLINK 1.9 kept them in the denominator. `IbcReport()` does the opposite, and the opposite is right: those variants get PLINK 1.9's `update_rel_ibc()` weights, so they stay in OBS_CT and contribute zero to all three estimates, and a genotype that cannot occur at the given allele frequency yields an infinite estimate, which is how 1.9 flags a --read-freq file that disagrees with the genotype data. Only the text needed fixing.
- Added `Tests/TEST_IBC`, which the command had no coverage for: Fhat1/Fhat2/Fhat3 and OBS_CT against PLINK 1.9 on a 700-variant, 200-sample simulated fileset, whole-sample, on a 120-sample subset (which changes the imputed frequencies), under --read-freq, and on the unfiltered fileset whose rare block supplies the monomorphic variants the note was about. Plus cols= subsetting and the 'zs' output.

chrchang:
To reiterate, the plan is now to remove the separate --ibc command, and instead implement the multiallelic generalizations of the three statistics as optional --het column-sets.  The error message from --ibc should point to "--het cols=+fhat1,+fhat2,+fhat3".

BenjaminDEMAILLE:
Understood: drop `--ibc`, add the three statistics as optional `--het` column sets, and point the retired flag at `--het cols=+fhat1,+fhat2,+fhat3`. Working on it now. Three things I checked first, one of which needs your call.

**Your Fhat1 and Fhat3 generalizations both hold.** I verified numerically rather than only by re-deriving: over 300 random allele-frequency vectors with 2 to 8 alleles, the HWE expectation of each per-variant term is 1 to within 6.7e-16, and over 900 (frequency, genotype) cases the general formula reproduces the biallelic GCTA term to within 1.2e-14.

**Fhat2 generalizes the same way**, which I do not think you spelled out. The biallelic term is 2 for a homozygote and `2 - 1/(2p(1-p))` for a heterozygote; the denominator there is just the expected heterozygosity, so the generalization is

```
term = 2                        for any homozygote
term = 2 - 1 / (1 - sum_i a_i^2)  for any heterozygote
```

Under HWE the expectation is `(1 - E_het) * 2 + E_het * (2 - 1/E_het) = 1`, and at two alleles `1 - sum a_i^2` is `2p(1-p)`, so both of your properties hold. It is in the same numeric check above.

**Fhat2 is not a duplicate of the existing `f` column.** I assumed it would be, since PLINK 1.9's documentation calls Fhat2 "excess homozygosity-based", but they use different denominators: `--het`'s F is a ratio of sums, `(sum obs_hom - sum exp_hom) / (sum nobs - sum exp_hom)`, while Fhat2 is a sum of ratios, `mean_j(1 - O_het_j / E_het_j)`. They agree only when the per-variant expected heterozygosities are equal. So all three column sets carry something the report does not already have.

**The one thing I would like your call on: monomorphic variants.** PLINK 1.9's `--ibc` counts them in OBS_CT and adds 0 to each of the three sums, so each monomorphic variant pulls the reported value toward -1 even though it carries no information about inbreeding. `--het` instead skips them entirely. If the three statistics become `--het` columns they will share `--het`'s OBS_CT, so they will differ from 1.9 and GCTA on any dataset with monomorphic variants retained.

I plan to go with `--het`'s behavior, since one denominator per report is the only thing that makes sense once they share a row, and excluding an uninformative variant seems clearly right on its own terms. But it is a silent numeric difference from GCTA, so say the word if you would rather I match 1.9 and give the three columns their own OBS_CT.

Implementation-wise the terms depend only on the genotype's unordered allele pair, so they fit `--het`'s existing base-plus-increment structure the same way `ehet` does: add the common genotype's value to a per-thread scalar and only the differences for the difflist entries. The sparse path stays sparse, and the whole thing is gated on the columns being requested, so the base case is untouched rather than 5% slower.

chrchang:
Yes, your Fhat2 generalization looks good.  Also agree that monomorphic variants should be skipped and we are ok with breaking from PLINK 1.9 / GCTA here.

BenjaminDEMAILLE:
Done, pushed. `--ibc` is gone; the three estimators are `--het` column sets, and `--ibc` errors with the pointer:

```
Error: --ibc has been retired.  Its three estimators are now optional --het
columns, which also support multiallelic variants; use
"--het cols=+fhat1,+fhat2,+fhat3".
```

**Monomorphic variants excluded**, as agreed.

**Implementation.** Since each term depends only on the genotype's unordered allele pair, they fit `--het`'s existing base-plus-increment structure directly: the common genotype's value goes into a per-thread scalar, and only the differences are written per sample, so the sparse difflist path stays sparse. The whole thing is gated on the columns being requested, so the base case is untouched rather than 5% slower.

For multiallelic variants the terms have closed forms that avoid an O(alleles) loop per sample. With `S = sum_i a_i^2`:

```
Fhat1  hom-i:  (4S + 4 - 8 a_i) / (2(1 - S))
       het:    (4S + 2 - 4(a_i + a_j)) / (2(1 - S))
Fhat2  hom:    2                het: 2 - 1/(1 - S)
Fhat3  hom-i:  1 + (1 - a_i)/a_i      het: 0
```

**Verification.** On a biallelic fileset the new columns are byte-identical to the `--ibc` output this PR previously carried, which was itself checked against PLINK 1.9, across all 400 samples, and identical under `--threads 1`. For the multiallelic case there is nothing to compare against, so `Tests/TEST_HET_IBC` ships an independent implementation of the formulas in Python and checks the report against it: 80 samples, 150 variants with 2 to 4 alleles, agreeing to the report's 6-significant-digit precision.

The suite also pins the two divergences: that the unfiltered fileset differs from `--ibc` in OBS_CT and only there, and that `f` and `fhat2` are not the same number.

One note on that last point, since PLINK 1.9's documentation misled me at first: Fhat2 is described there as "excess homozygosity-based", which reads like `--het`'s F, but they use different denominators (F is a ratio of sums over variants, Fhat2 a mean of per-variant ratios). They coincide only when the per-variant expected heterozygosities are equal, so all three column sets do carry something the report did not already have.

PR #351, opened 2026-08-31
Add --fisher (allelic case/control exact test)

BenjaminDEMAILLE:
From the parity audit in #344.

**Up front:** this reintroduces functionality from `--assoc`, which you deliberately retired in favour of `--glm`. My argument for it is narrow: `--glm` covers the asymptotic and Firth cases, but not the *exact* test, and that is what rare-variant case/control analyses generally want. If you would rather expose Fisher's exact test through a different surface, the `FisherExact2x2()` half of this PR stands on its own and the command half can be dropped.

### Two parts

**1. `FisherExact2x2()`** in `include/plink2_stats.cc`, a port of 1.9's `fisher22()` reusing the existing `kExactTestBias` constant. Plain and mid-p variants.

**2. `--fisher ['zs'] ['midp'] ['cols='<...>]`**, writing `<out>.fisher`. A1 is the minor allele, as in 1.9. Column sets: `chrom`, `pos`, `ref`, `alt`, `a1`, `counts`, `freq`, `nobs`, `or`, `p`.

### Scope decisions

- **Autosomal biallelic variants only.** chrX and the haploid chromosomes need male-haploid allele counting plus a decision about het haploid calls. Rather than guess at 1.9's conventions there, `--fisher` restricts itself the way `--het` and `--ibc` do. On a mixed-chromosome fileset, 1.9 reports 2500 variants and `--fisher` reports the 1800 autosomal ones, with identical counts on all 1800. Happy to extend this if you tell me which chrX model you want.
- **Exactly one case/control phenotype**, since `--fisher` writes a single file. With more than one it asks for `--pheno-name` rather than silently picking one.

### Verification

macOS/arm64, against `PLINK v1.9.0-b.7.11`.

**The statistic itself.** I compiled 1.9's `fisher22()` and my port into one binary and compared them directly over **400,000 random 2x2 tables** (both `midp` settings, entries up to 100,000):

```
tables=400000  max relative diff=0  mismatches=0
```

Bit-identical, not merely close.

**End to end.** Since the p-value function is bit-identical, the remaining question is whether the allele counts feeding it match. Compared against `--assoc fisher counts`, which prints integer counts:

| dataset | result |
|---|---|
| 1000 samples x 1800 autosomal variants | A1 identical, **integer allele counts identical on every variant**, P and OR agree to 1.9's 4-significant-digit print precision (max rel. diff 4.6e-4) |
| same, `midp` vs `--assoc fisher-midp` | identical counts, same agreement |
| 2504 real 1000 Genomes chr21 samples, 600 variants | A1 identical, **allele counts identical**, max rel. dP 3.6e-4, max rel. dOR 4.3e-4 |
| mixed-chromosome fileset | all 1800 autosomal variants match exactly; the 700 non-autosomal ones are absent by design |

The residual p-value differences are entirely 1.9's output rounding: the counts are equal integers and the function is bit-identical, so the underlying doubles must be too.

`--keep` subsets, `zs`, `cols=` subsetting, and the error paths (no case/control phenotype, two case/control phenotypes, a subset with no controls, bad column name) all behave as intended. Output is byte-identical to a `master` build for `--het`, `--freq`, `--hardy`, `--make-bed`, `--glm` and `--indep-pairwise`. `2.0/Tests/run_tests.sh` passes 8/8; build is warning-free.

### DIFF
--- a/2.0/include/plink2_stats.cc
+++ b/2.0/include/plink2_stats.cc
@@ -1968,6 +1968,104 @@ double HweLnP(int32_t obs_hets, int32_t obs_hom1, int32_t obs_hom2, int32_t midp
 // constant to be compatible with them doesn't cost us anything.)
 static const double kExactTestBias = k2m50 / (1LL << 33);
 
+// 2^{-40}; matches PLINK 1.9's FISHER_EPSILON.
+static const double kFisherEpsilon = 0.0000000000009094947017729282379150390625;
+
+double FisherExact2x2(uint32_t m11, uint32_t m12, uint32_t m21, uint32_t m22, uint32_t midp) {
+  // Two-sided Fisher's exact test on a 2x2 contingency table, ported from
+  // PLINK 1.9's fisher22().  Sums the likelihoods of all tables at least as
+  // extreme as the observed one, relative to the observed table's likelihood.
+  // With midp, half the probability mass of the tables tied with the observed
+  // one is subtracted.
+  double tprob = (1 - kFisherEpsilon) * kExactTestBias;
+  double cur_prob = tprob;
+  double cprob = 0.0;
+  int32_t tie_ct = 1;
+  // Ensure we are left of the distribution center, m11 <= m22, and m12 <= m21.
+  if (m12 > m21) {
+    const uint32_t uii = m12;
+    m12 = m21;
+    m21 = uii;
+  }
+  if (m11 > m22) {
+    const uint32_t uii = m11;
+    m11 = m22;
+    m22 = uii;
+  }
+  if ((S_CAST(uint64_t, m11) * m22) > (S_CAST(uint64_t, m12) * m21)) {
+    uint32_t uii = m11;
+    m11 = m12;
+    m12 = uii;
+    uii = m21;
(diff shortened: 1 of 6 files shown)


chrchang:
* A better implementation of Fisher's Exact Test is under https://github.com/chrchang/stats .
* I don't see an appropriate place for that code in PLINK 2 at this time, since there is no dedicated rare-variant analysis module.  So I'm instead working on C/Python/R libraries which the folks developing rare-variant analysis tools can use.

BenjaminDEMAILLE:
Understood, closing.

For anyone who finds this later: the reasoning is that plink2 has no rare-variant analysis module for an exact allelic test to belong to, and the better Fisher implementation lives at https://github.com/chrchang/stats.

One note from the port, in case it is useful for the C library: PLINK 1.9's `fisher22()` is exact within `kExactTestBias` = 2^-83 and terminates on `kFisherEpsilon` = 2^-40, and my port reproduced 1.9's p-values bit-for-bit on the cases I ran, so it can serve as a cross-check for the new implementation if you want one.

PR #352, opened 2026-08-31
Add --indep (variance-inflation-factor pruner)

BenjaminDEMAILLE:
`plink2_ld.h` carries a `// todo: old multicollinearity test` note right below `LdPruneFlags`, and `prune_last_param` is already documented as "VIF or r^2 threshold". This fills that in.

```
--indep <window size>['kb'] [step size (variant ct)] <VIF threshold>
```

### Algorithm

`IndepVif()` follows PLINK 1.9's `ld_prune()` for the non-pairwise case:

1. Remove pairs in near-perfect LD (`|r| > 0.999999`), keeping the higher-MAF member.
2. Repeatedly invert the correlation matrix over the survivors and remove the variant with the largest diagonal entry (its VIF) as long as it exceeds the threshold.
3. On a singular matrix, binary-search for the shortest invertible leading submatrix and drop the row/column just past it, then retry, as 1.9 does.

Correlations use integer accumulators with pairwise-complete handling, matching 1.9's `ld_process_load()` / `ld_dot_prod()` arithmetic. Heterozygous haploid calls are treated as missing, as in 1.9.

### Two implementation notes

- **Each window is recomputed from scratch** rather than carrying correlations across window slides. A pair's correlation does not depend on which window it is seen in, so the results are identical; this trades speed for a much simpler implementation than 1.9's incremental `cov_matrix` / `idx_remap` bookkeeping.
- **Single-threaded.** On 600 variants x 2504 samples: 0.08s versus 1.9's 0.01s. Fine at the sizes `--indep` is normally used on, but the obvious thing to improve if you take this.

### Verification

macOS/arm64 against `PLINK v1.9.0-b.7.11`. "diff" counts differing lines between sorted `.prune.in` files, so 0 means the prune lists are identical.

**Autosomal, 1000 Genomes chr21 (600 variants, 2504 samples):**

| configuration | 1.9 retained | plink2 retained | diff |
|---|---:|---:|---:|
| window 50, step 5, VIF 1.02 | 472 | 472 | 0 |
| ... VIF 1.05 | 500 | 500 | 0 |
| ... VIF 1.1 | 522 | 522 | 0 |
| ... VIF 1.25 | 549 | 549 | 0 |
| ... VIF 1.5 | 559 | 559 | 0 |
| ... VIF 2 | 567 | 567 | 0 |
| ... VIF 3 | 577 | 577 | 0 |
| ... VIF 5 | 578 | 578 | 0 |
| ... VIF 10 | 582 | 582 | 0 |
| window 200, step 50, VIF 1.5 | 510 | 510 | 0 |
| window 10, step 2, VIF 1.1 | 572 | 572 | 0 |
| 100kb window, VIF 1.5 | 191 | 191 | 0 |
| 500kb window, VIF 2 | 198 | 198 | 0 |
| 1000kb window, VIF 3 | 210 | 210 | 0 |

13 of 13 identical, including the kb-window and low-threshold cases where most variants get pruned.

A simulated 1000-sample / 1800-variant autosomal fileset with 3% missingness (which exercises the pairwise-complete correlation path) is also identical.

**chrX:** on a mixed-sex chrX fileset, both retain 396 variants, with 2 of 600 assigned differently. plink2's existing `--indep-pairwise` shows the same-size disagreement with 1.9 on that same fileset, so this looks like plink2's chrX convention rather than a discrepancy introduced here.

Output is byte-identical to a `master` build for `--indep-pairwise`, `--indep-pairphase`, `--freq`, `--make-bed` and `--r2-unphased`. Error paths (`VIF threshold < 1`, combining with another pruning command, `--indep-preferred`) behave sensibly. `2.0/Tests/run_tests.sh` passes 8/8; build is warning-free.


BenjaminDEMAILLE:
Correcting the performance claim in the description above, and fixing what it was hiding.

I benchmarked the original version on the 600-variant 1000 Genomes fileset and reported 0.08s against 1.9's 0.01s. That was a bad benchmark: the cost grows with variant count, so at any realistic size the gap was much worse than 8x. On 5000 variants x 5000 samples it was **5.3s against 1.9's 0.10s**, which is impractical rather than merely slower.

Two causes, both now fixed in `e9f51ba`:

**Correlations were scalar per sample.** With any missingness present, every pair took the pairwise-complete branch, which tested two bits per sample. Each variant is now held as three bitvectors (code +1, code -1, nonmissing), so the dot product, the masked sums and the sums of squares all reduce to popcounts over whole words: `x*y` is +1 when the two codes agree in sign and -1 when they differ, and `x^2` is 1 for every nonzero code. The accumulators stay exact integers, so the resulting doubles are bit-for-bit what the scalar loop produced.

**Every window recomputed all its pairs.** The description said this was a deliberate simplification, and it was, but it cost about a factor of ten at the default step. Slots are now assigned as `position % window_max`, which is collision-free within a window since a window never holds more than `window_max` variants, so a pair's correlation survives window slides and each variant is loaded and correlated once.

Warm cache, four consecutive runs each, 5000 samples, window 50, step 5, VIF 2:

| | 5000 variants | 2000 variants |
|---|---|---|
| before | 5.55 5.29 5.29 5.30 | 2.48 |
| after | 0.15 0.15 0.15 0.15 | 0.06 |
| PLINK 1.9 | 0.10 0.10 0.10 0.10 | 0.04 |

So 35x faster, and now within 1.5x of 1.9 rather than 50x off.

Prune lists are unchanged. I re-ran the whole parity battery: ten autosomal configurations on 1000 Genomes chr21 (VIF 1.02/1.1/2/10, window 200 step 50, window 10 step 2, and 100kb/500kb/1000kb windows) plus the simulated autosomal fileset, all with identical `.prune.in` and `.prune.out`. The two chrX filesets still differ from 1.9 by the same 2 variants as before, which is the convention difference already noted, not a regression. Output stays byte-identical to master for `--indep-pairwise`, `--indep-pairphase`, `--freq`, `--make-bed` and `--r2-unphased`, and `2.0/Tests/run_tests.sh` passes 8/8.

BenjaminDEMAILLE:
For consistency with the triage I just posted on #344, I classified `--indep` there as superseded: the VIF criterion was a workaround for not having a fast pairwise pruner, and `--indep-pairwise` is that pruner. On that reading this PR is the same category as #351, and I am happy to close it.

Leaving it open only because the two criteria are not interchangeable, so anyone reproducing an old pipeline gets a different variant set from `--indep-pairwise`. If that reproducibility argument does not move you, say so and I will close it.

chrchang:
Since it is straightforward to run --make-bed and then call PLINK 1.9 --indep, there is no actual need to "phone in" an implementation that is not obviously better than 1.9's.  Compare PLINK 2.0 --indep-pairwise to 1.9 --indep-pairwise, and then aim for at least the square root of that wall-clock-time reduction.

BenjaminDEMAILLE:
Done, and the bar was the right thing to ask for: it turned up three correctness bugs, not just a slow implementation. Pushed as `634aa72`.

**Measurements**, 5000 samples x 50000 variants over 22 chromosomes, best of 7 warm runs, 16-core M1:

| | PLINK 1.9 | plink2 | |
|---|---:|---:|---|
| `--indep-pairwise 50 5 0.5` | 0.543s | 0.131s | 4.1x |
| `--indep 50 5 2` | 0.975s | 0.080s | **12.2x** |
| `--indep 50 5 2 --threads 1` | 0.975s | 0.469s | 2.1x |

So the target was sqrt(4.1) = 2.0x and the result is 12.2x. Single-threaded it is 2.1x, which clears the bar on its own; the rest is threading. The ratio you get for `--indep-pairwise` depends on the dataset: on a single-chromosome version of the same data it drops to 1.7x, since threading has nothing to distribute, and `--indep` is 1.6x there.

**What changed.** Chromosomes are distributed over threads the way `--indep-pairwise` distributes subcontigs, with `LoadBalance()` doing the assignment. Chromosomes rather than gap-split subcontigs, because 1.9 restarts window start offsets at each chromosome boundary, so splitting further would shift them and change the output. The main thread reads into a per-thread ring buffer sized so it never gets more than a batch ahead of what the compute thread consumed; each thread accumulates removals in its own bitarray, OR'd at the end, so results do not depend on `--threads`. Separately, the -1/0/1 recode was a per-sample `GetNyparrEntry()`/`SetBit()` loop and now goes through `GenoarrSplit12Nm()` plus word masking, which is where the single-threaded 2.1x comes from.

I also tried replacing `InvertMatrixChecked()` with the Cholesky-based `InvertSymmdefMatrixChecked()`. It is ~11% faster and I reverted it: the two disagree about which matrices count as singular, and the singular case is exactly where 1.9's binary-search-and-drop path runs, so windows larger than the founder count prune differently.

**Three bugs the verification found**, all in the version you were looking at:

1. **A `kb` window was 1000x too wide.** plink2's parser already converts kb to bp; 1.9's keeps the kb value. I ported 1.9's `1000 * ld_window_size` literally, so every `--indep <n> kb` run used an `n`-megabase window.
2. **The window tail at the end of a chromosome was skipped.** 1.9 keeps sliding the start to the last variant. For a variant-count window those tails are subsets of the last full window and prune nothing further, which is why the autosomal tests still matched; for a window wide enough to span a chromosome the singular path does drop more, and 1.9 pruned 462 of 1800 variants where I pruned 164.
3. **Per-variant genotype buffers were spaced `NypCtToWordCt()` words apart.** `PgrGet()` writes whole vectors, so with an odd word count each variant ran one word into the next, and the last variant of a batch ran off the end of the buffer. This one only showed up once the buffers became per-thread rings.

**Verification.** `--indep` output is identical to 1.9's, and independent of `--threads`, on 28 configurations: variant-count windows (sizes 20/50/100, steps 1/3/5/10, thresholds 1.5/2/5) and kb windows (50/200/500 kb), with and without `--maf`, on a 2504-sample 1000 Genomes chr21 subset, a 1000-sample 1800-variant set, and 5000-sample 50000-variant sets laid out as 1 and as 22 chromosomes. `2.0/Tests/run_tests.sh` passes; `--freq`, `--make-bed`, `--glm`, `--indep-pairwise`, `--indep-pairphase`, `--r2-unphased` and `--hardy` are byte-identical to a `master` build.

**One deliberate difference, now surfaced rather than silent.** chrX does not match 1.9: its default `--ld-xchr 1` puts males on a 0/1 dosage scale and females on 0/1/2, and this scores both the same way. Restricted to one sex, chrX matches 1.9 exactly (56/56 and 59/59 pruned on my test set); mixed, 1.9 pruned 28 chrX variants where this pruned 0. `--indep` now warns on chrX and the help text says so. If you want the PR at all, I can implement the `--ld-xchr 1` scale instead; it needs a male/female split in the correlation, like `IndepPairwiseUpdateSubcontig()` does.

The "is `--indep` worth having" question from #344 still stands separately, and closing this remains fine by me.

chrchang:
--indep is substantially lower-priority than --homozyg, but I'd still be willing to merge a good implementation.  I will take a closer look at this PR later this week.

2.0/plink2_ld.cc
```diff
@@ -2527,15 +2528,893 @@ PglErr LdPruneWrite(const uintptr_t* variant_include, const uintptr_t* removed_v
   return reterr;
 }
 
+// --indep: PLINK 1.x's variance-inflation-factor pruner.  Within each window,
+// pairs in near-perfect LD are removed first, then the variant with the
+// largest diagonal entry of the inverted correlation matrix (its VIF) is
+// removed as long as that entry exceeds the threshold.
+//
+// Unlike IndepPairwise(), this recomputes each window's correlations from
```
chrchang: This comment should be updated to reflect the excellent improvements that were made since the first draft.

BenjaminDEMAILLE: Rewritten. It still described the first draft: single-threaded, and recomputing each window's correlations from scratch.

Issue #353, opened 2026-08-31
PLINK 1.9: --homozyg reports spurious chrX runs for some male samples (wrong subset_size in copy_bitarr_subset_excl call)


BenjaminDEMAILLE:
Found while implementing `--homozyg` for plink2 and diffing outputs.

On chrX, `--homozyg` is supposed to skip males (`roh_update()`: *"skip males on Xchr"*). It usually does, but for some males it does not, and reports a run of homozygosity for them.

### Repro

200 chrX variants, all samples homozygous, one male and the rest female:

```bash
# n samples, the male at index <male_idx>
plink --file mini --allow-no-sex --homozyg --homozyg-snp 10 --homozyg-kb 1 \
      --homozyg-window-snp 10 --out m19
```

| n | male index | segments reported | male reported? |
|---:|---:|---:|---|
| 65 | 63 | 64 | no (correct) |
| 65 | 64 | **65** | **yes** |
| 128 | 127 | 127 | no (correct) |
| 130 | 129 | **130** | **yes** |
| 200 | 63 / 64 / 130 | 199 | no (correct) |

On real data: 1000 Genomes chr21 relabelled as chrX, 2504 samples with alternating sexes, `--homozyg --homozyg-snp 10 --homozyg-kb 10 --homozyg-window-snp 10` reports **1215 female segments plus 4 male segments**. Dropping to 2496 samples (a multiple of 64) makes the 4 male segments disappear.

### Cause

`plink_homozyg.c:2552`:

```c
copy_bitarr_subset_excl(sex_male, sample_exclude, sample_ct,
                        popcount_longs_exclude(sex_male, sample_exclude, sample_ctl),
                        sample_male);
```

`copy_bitarr_subset_excl()`'s fourth parameter is `subset_size`, the number of **output** bits to produce. Every other call site in the codebase passes the sample count:

```
plink_calc.c:2771:  copy_bitarr_subset_excl(pheno_nm, sample_exclude, unfiltered_sample_ct, sample_ct, g_pheno_nm);
plink_calc.c:2962:  copy_bitarr_subset_excl(pheno_nm, sample_exclude, unfiltered_sample_ct, sample_ct, pheno_nm_local);
plink_dosage.c:1433: copy_bitarr_subset_excl(pheno_nm, sample_exclude, unfiltered_sample_ct, sample_ct, pheno_nm_collapsed);
```

[...]

PR #354, opened 2026-08-31
Add --homozyg (runs of homozygosity)

BenjaminDEMAILLE:
From the parity audit in #344: ROH was the largest module with no plink2 substitute at all.

Ports PLINK 1.9's `calc_homozyg()`/`roh_update()` scanning-window algorithm, with the nine tuning flags and the three report files.

```
--homozyg ['subtract-1-from-lengths']
--homozyg-snp / -kb / -density / -gap / -het
--homozyg-window-snp / -window-het / -window-missing / -window-threshold
```

A window of `--homozyg-window-snp` variants is a hit for a sample when it holds at most `--homozyg-window-het` heterozygous and `--homozyg-window-missing` missing calls; a variant is eligible when at least `--homozyg-window-threshold` of the windows covering it are hits; maximal eligible runs are reported when they satisfy `--homozyg-snp`, `--homozyg-kb`, `--homozyg-density` and `--homozyg-het`, with `--homozyg-gap` forcing a break. Fully-haploid chromosomes and chrMT are skipped, males are skipped on chrX, and chromosomes with fewer variants than the scanning window produce no runs, all as in 1.9.

Writes `.hom`, `.hom.indiv` and `.hom.summary`, in plink2's tab-delimited style rather than 1.9's fixed-width columns.

**Not implemented**, and rejected with a clear message rather than silently ignored: the `group`/`group-verbose` pool reports, `consensus-match`, `extend`, `--homozyg-match` and `--pool-size`. Those are a separate subsystem in 1.9 (pairwise allelic matching, pools, heaps) and I did not want to guess at it.

### Verification

macOS/arm64 against `PLINK v1.9.0-b.7.11`. 1000 Genomes chr21, 600 variants x 2504 samples.

| configuration | segments (both) | only in 1.9 | only in 2.0 | field mismatches |
|---|---:|---:|---:|---:|
| baseline (snp 10, kb 10, window 10) | 2427 | 0 | 0 | 0 |
| `--homozyg-window-het 0` | 16 | 0 | 0 | 0 |
| `--homozyg-window-missing 0` | 2427 | 0 | 0 | 0 |
| `--homozyg-window-threshold 0.5` | 832 | 0 | 0 | 0 |
| `--homozyg-het 0` | 16 | 0 | 0 | 0 |
| `--homozyg-het 2` | 40 | 0 | 0 | 0 |
| `--homozyg-density 1` | 2427 | 0 | 0 | 0 |
| `--homozyg-gap 5` | 2427 | 0 | 0 | 0 |
| window/snp 30 | 3240 | 0 | 0 | 0 |
| `subtract-1-from-lengths` | 2427 | 0 | 0 | 0 |

"field mismatches" compares POS1, POS2, KB and NSNP on every shared segment; KB matches to the last digit. DENSITY/PHOM/PHET agree to 1.9's 3-decimal output precision (max absolute difference 5e-4).

For the baseline run all three files were checked in full:

- `.hom`: 2427/2427 segments identical (sample, chromosome, SNP1, SNP2, POS1, POS2, NSNP, KB).
- `.hom.indiv`: NSEG, KB and KBAVG identical for all 2504 samples.
- `.hom.summary`: AFF and UNAFF identical for all 600 variants.

A simulated 1000-sample autosomal fileset with 3% missingness is also identical (947 segments).

### chrX

On chrX the two disagree, and I am fairly confident the difference is a PLINK 1.9 bug rather than a divergence here: 1.9 fails to skip males whose index falls in the final partially-filled word of its sample bitarray, because `plink_homozyg.c` passes the male count where `copy_bitarr_subset_excl()` expects the sample count. Filed separately as #353, with a minimal repro (65 samples, male at index 64) and the boundary cases.

With a sample count that is a multiple of 64 the bug does not fire, and the two agree exactly: 1211/1211 segments on a 2496-sample chrX fileset, no differences.

Output is byte-identical to a `master` build for `--freq`, `--het`, `--make-bed`, `--hardy` and `--indep-pairwise`. Error paths (unimplemented modifiers, bad arguments) behave sensibly. `2.0/Tests/run_tests.sh` passes 8/8; build is warning-free.


BenjaminDEMAILLE:
Pushed a performance commit (`80ac802`). The scan is now faster than PLINK 1.9's rather than 3x slower, with identical output.

The first version cost three full passes over the sample array per variant: the main scan, a pass to retire the outgoing row's heterozygous/missing counts, and a pass to expand the next row into a byte array. 1.9 gets away with one full pass plus two sparse ones.

- Window rows are now packed 2-bit codes instead of one byte per sample, so a 50-variant window over 5000 samples occupies 62 KB rather than 250 KB and the scan shifts through a word in a register.
- Retiring the outgoing row's heterozygous/missing counts iterates the set bits of the recoded word; homozygous calls encode to zero, so it only touches samples that actually have one.
- Window-hit bits are stored one word per 32 samples, matching the genotype word layout, so retiring them is a set-bit iteration rather than a per-sample `IsSet`.

5000 samples x 50000 variants, `--homozyg-snp 50 --homozyg-kb 100`, warm cache, four consecutive runs each:

```
before      1.67 1.40 1.41 1.40
after       0.42 0.42 0.42 0.42
PLINK 1.9   0.57 0.57 0.57 0.57
```

Worth recording a wrong turn: I first tried fusing all three passes into one heavier loop, which only got to 0.66s. Splitting the sparse work back out, the way 1.9 structures it, is what actually paid. The ablation that pointed there was that `--homozyg-window-het 0` and the default cost the same, so the remaining time was traversal, not the ROH state machine.

Output is unchanged: the full parity check re-ran clean, twelve configurations on 1000 Genomes chr21 plus a simulated autosomal fileset and the 2496-sample chrX fileset, identical segment sets with identical POS1/POS2/KB/NSNP, `.hom.indiv` identical for all 2504 samples, `.hom.summary` identical for all 600 variants. `2.0/Tests/run_tests.sh` passes 8/8.

2.0/plink2_help.cc
```diff
@@ -815,6 +815,38 @@ PglErr DispHelp(const char* const* argvk, uint32_t param_ct) {
...
+"      (default 5).  A variant is eligible for a ROH when at least\n"
+"      --homozyg-window-threshold of the windows covering it are hits\n"
+"      (default 0.05).\n"
+"    * Reported runs must contain at least --homozyg-snp variants (default\n"
+"      100), span at least --homozyg-kb kilobases (default 1000), average at\n"
+"      least one variant per --homozyg-density kb (default 50), and contain at\n"
+"      most --homozyg-het heterozygous calls (unlimited by default).  Two\n"
+"      consecutive variants more than --homozyg-gap kb apart (default 1000)\n"
+"      cannot be in the same run.\n"
+"    * Fully-haploid chromosomes and chrMT are skipped, and males are skipped\n"
```
chrchang: Documentation generally uses 'chrM' instead of 'chrMT'.

2.0/plink2_help.cc
```diff
@@ -815,6 +815,38 @@ PglErr DispHelp(const char* const* argvk, uint32_t param_ct) {
 "    maybefid/fid/maybesid/sid to the .imendel output file; and\n"
 "    chrom/pos/ref/alt to the .lmendel[.zst] output file.\n\n"
               );
+    HelpPrint("homozyg\0homozyg-snp\0homozyg-kb\0homozyg-density\0homozyg-gap\0homozyg-het\0homozyg-window-snp\0homozyg-window-het\0homozyg-window-missing\0homozyg-window-threshold\0", &help_ctrl, 1,
+"  --homozyg ['subtract-1-from-lengths']\n"
```
chrchang: For consistency with the rest of plink2's reports, this should be something like "  --homozyg ['zs'] ['subtract-1-from-lengths'] (column-control options for each output file where that's plausibly useful)", and there should be list(s) of supported column sets.  'zs' should apply to the output files with one entry per variant (currently .hom.summary), and anything else that is routinely very large (I guess this is *not* the case with the .hom file).

2.0/plink2_misc.cc
```diff
@@ -10374,6 +10374,644 @@ PglErr HetCalcMain(const uintptr_t* sample_include, const uintptr_t* variant_sub
...
+      const uint32_t bit_idx = ctzw(cur_word);
+      if (bit_idx & 1) {
+        het_cts[sample_idx_base + (bit_idx / 2)] += 1;
+      } else {
+        missing_cts[sample_idx_base + (bit_idx / 2)] += 1;
+      }
+      cur_word &= cur_word - 1;
+    }
+  }
+  (void)sample_ct;
```
chrchang: What is this line for?

2.0/plink2_misc.cc
```diff
@@ -10374,6 +10374,644 @@ PglErr HetCalcMain(const uintptr_t* sample_include, const uintptr_t* variant_sub
...
+} RohRecord;
+
+static const double kRohEpsilon = 0.000000000931322574615478515625;  // 2^-30
+static const double kRohSmallishEpsilon = 0.00000000000005684341886080801486968994140625;  // 2^-44
+
+// Recodes a genovec into the representation --homozyg scans: 0 = homozygous,
+// 1 = missing, 2 = heterozygous, in place of plink2's 0/1/2/3.  Also adds the
+// row's heterozygous and missing calls to the running window counts, touching
+// only the samples that actually have one.
+static void RecodeHomozygRow(const uintptr_t* genovec, uint32_t sample_ct, uint32_t sample_ctl2, uint32_t* het_cts, uint32_t* missing_cts, uintptr_t* dst) {
```
chrchang: In PLINK 2, this needs to handle multiallelic variants.  patch_01 can be ignored, but when patch_10_ct is nonzero, it is necessary to iterate through the (patch_10_set, patch_10_vals) entries and correct dst[] and het_cts[] for each ALTx/ALTy heterozygous genotype found in that list.

2.0/plink2_misc.cc
```diff
@@ -10374,6 +10374,644 @@ PglErr HetCalcMain(const uintptr_t* sample_include, const uintptr_t* variant_sub
...
+          }
+        }
+        const uint32_t forced_end = (variant_bps[old_uidx] - variant_bps[older_uidx]) > max_gap;
+        uint32_t sample_idx = 0;
+        for (uint32_t geno_widx = 0; sample_idx != sample_ct; ++geno_widx) {
+          uintptr_t geno_word = cur_row[geno_widx];
+          uintptr_t sw_new = 0;
+          const uint32_t idx_stop = MINV(sample_idx + kBitsPerWordD2, sample_ct);
+          for (uint32_t lowbits = 0; sample_idx != idx_stop; ++sample_idx, ++lowbits) {
+          const uintptr_t cur_call = geno_word & 3;
```
chrchang: There is oddly missing indentation here.

chrchang (commented):
While multithreading is not strictly necessary for this PR, the code should be organized with the expectation that multithreading will be implemented relatively soon.

BenjaminDEMAILLE:
Reorganized and multithreaded; pushed as `4cb4fdb`.

The scan turned out to be cleanly parallel over samples rather than over variants: every array it touches is indexed by sample, and the window buffers hold one word per 32 samples, so two calls covering disjoint sample-word ranges touch disjoint memory. That is now the unit of work. `HomozygScanRange()` advances the scan across a range of variants for a range of sample words, and the reader hands one range to each thread.

Three things had to change to get there:

- **The reader now stays a window ahead of the scan**, so a whole block of scan steps runs without touching the file. The recoded-row ring is sized for a block plus a window, plus the one row before the block whose bp position the block's first step still needs. (That last row cost me an hour: without it the first step of every block saw a stale bp and split runs at block boundaries, which showed up as segments of exactly `block_size` SNPs.)
- **The scan is now expressed as `chr_variant_ct + 1` steps, each derived from its variant offset** rather than from loop-carried state. This replaced the separate main and shrinking-window loops, which had duplicated copies of the run-detection logic, with one loop.
- **`RecodeHomozygRow()` only recodes now**; folding a row into the window's het/missing counts is per-sample work, so it moved into the workers.

Each worker accumulates into its own slice of the ROH list, and the slices are concatenated in worker order afterwards. A worker only ever writes records for samples in its own range, so each sample's records stay contiguous and in discovery order, and output does not depend on `--threads`.

Timings, best of 6 warm runs, 16-core M1:

| | PLINK 1.9 | before | after | |
|---|---:|---:|---:|---|
| 5000 x 200000, 22 chr | 1.844s | 1.604s | **0.363s** | 5.1x vs 1.9 |
| 5000 x 50000, 22 chr | 0.463s | 0.395s | **0.090s** | 5.1x |
| 2000 x 100000, ROH-rich | 0.415s | 0.357s | **0.109s** | 3.8x |

The third set has runs of homozygosity injected (12708 segments found), so the run bookkeeping is actually exercised; the other two are unstructured, where the scan finds nothing.

Along the way I also tried a sparse formulation of the inner loop, visiting only the samples whose run state could change (`(hit ^ in_run) | (in_run & non-homozygous)`), on the theory that most samples are a no-op at any given variant. Measured, it was 3-6% *slower* than the straightforward loop: the per-sample het/missing bookkeeping dominates, and that is not sparse. Dropped it rather than carry the complexity.

Verification: output identical to PLINK 1.9's, and independent of `--threads` (1, 3 and 16), on 30 configurations covering default settings plus varied `--homozyg-kb/-snp/-window-snp/-window-het/-window-missing/-window-threshold/-density/-gap/-het`, across a ROH-rich 2000-sample set, a 600-sample set, a 2504-sample chrX set, a 1000 Genomes chr21 subset, and 1000- and 5000-sample simulated sets. `2.0/Tests/run_tests.sh` passes; `--freq`, `--make-bed`, `--glm`, `--het` and `--hardy` are byte-identical to a `master` build.

chrchang:
Generally looks good now (though the transient CI failure was disturbing, may be important to track down the root cause if it comes up again).  The main remaining step at this point is to define an initial group of plink2-style column sets.

2.0/plink2_help.cc
BenjaminDEMAILLE: Fixed.

2.0/plink2_misc.cc
BenjaminDEMAILLE: Leftover from an earlier revision where the function still took sample_ct. Removed.

2.0/plink2_misc.cc
BenjaminDEMAILLE: Fixed. That block has since been pulled out into HomozygScanRange(), so the misaligned lines are gone with it.

2.0/plink2_misc.cc
BenjaminDEMAILLE: Implemented. Variants with more than two alleles now go through `PgrGetM()`, and the patch_10 entries whose two allele codes differ are recoded as heterozygous in `dst[]` before the row is folded into the window counts. The het count is derived from `dst[]` by `HomozygFoldRow()`, so correcting `dst[]` corrects both.

Test: one sample, 60 variants, all homozygous except variant 30. Three versions of that variant, all with the same scan settings (`--homozyg-window-het 0`, so a single heterozygous call splits the run):

| variant 30 | runs found |
| --- | --- |
| `0/0` | one run, v0-v59 |
| `0/1` (biallelic het) | two runs, v0-v29 and v31-v59 |
| `1/2` (ALT1/ALT2 het) | two runs, v0-v29 and v31-v59 |

The last row was one run of v0-v59 before the fix, i.e. the heterozygote was invisible. The `1/2` and `0/1` outputs are now identical field for field.

BenjaminDEMAILLE:
Column sets are in, along with the multiallelic fix and the smaller points from the inline comments.

Following the `--mendel` pattern: one `cols=` descriptor whose members apply to whichever of the three files they are relevant for. IID is always present, as are ID1/ID2 (the first and last variant IDs of the run) in .hom and ID in .hom.summary.

```
Shared between .hom and .hom.indiv:
  maybefid, fid, maybesid, sid, maybepheno, pheno
Shared between .hom and .hom.summary[.zst]:
  chrom, pos
Specific to .hom:
  kb, nsnp, density, phom, phet
Specific to .hom.indiv:
  nseg, kbtot, kbavg
Specific to .hom.summary[.zst]:
  aff, unaff
```

Default is everything except `fid`/`sid`/`pheno`, i.e. what PLINK 1.9 emits.

`zs` applies to .hom.summary only, which is the one file with one line per variant. .hom has one line per run and .hom.indiv one per sample, so neither seemed worth compressing; happy to extend it if you disagree.

Two naming changes worth flagging, since they change the output rather than just adding to it: SNP1/SNP2 are now ID1/ID2, and .hom.summary is ordered `#CHROM POS ID` rather than `#CHROM ID POS`, to match how the rest of plink2 orders those columns.

Values are unchanged; .hom.indiv still matches 1.9 field for field on both test filesets. I have not reproduced the transient CI failure locally across the runs since.

2.0/plink2_help.cc
BenjaminDEMAILLE: Done, following the `--mendel` pattern: one `cols=` descriptor whose members apply to whichever of the three files they are relevant for. The synopsis is now

```
  --homozyg ['zs'] ['subtract-1-from-lengths'] [{'cols='<column set descriptor>}]
```

and the help lists the sets grouped by which file each applies to. `zs` covers .hom.summary only, the one output with a line per variant; .hom has a line per run and .hom.indiv one per sample, so neither looked worth compressing. Say the word if you want it to cover them too.

chrchang:
I see two primary ways for the default --homozyg parameters to go wrong when used on modern datasets:
1. There are many variants with tiny MAF.  I suspect this command would benefit from a required --homozyg-maf parameter with a documented recommendation of 0.05 for the original other defaults.
2. There is too much LD.  As I write this, there is a note about LD pruning at the top of https://www.cog-genomics.org/plink/1.9/ibd , but now that I think about it, that note should be repeated for each of the two commands, and added to the --help entry for --homozyg (looks like I already added it to --genome).

Hopefully that does enough to point this particular gun away from some users' feet.  Some sort of LD-pruning-related sanity check may be worth implementing in the future, but it isn't yet obvious to me how that might work.

BenjaminDEMAILLE:
Both done and pushed.

**`--homozyg-min-af`**, default 0.05. On 20000 variants x 600 samples:

```
default (0.05)   136 variants excluded    1630 runs
0.2             6697 variants excluded    1626 runs
0                  no exclusion           1631 runs
```

The default changes output, which is the point of the change. `--homozyg-min-af 0` is byte-identical to the previous behaviour in both `.hom` and `.hom.summary`, so the PLINK 1.x comparison stays reachable for anyone who needs it, and my own regression check against 1.9 still has something to compare against.

**LD warning** added to the `--homozyg` help entry, alongside a note on what the frequency floor is for:

```
  * These results are only meaningful on an LD-pruned variant set: linkage
    disequilibrium extends runs that are not evidence of autozygosity.  Use
    --indep-pairwise first.
  * Variants with minor allele frequency below --homozyg-min-af (default
    0.05) are excluded, since rare variants are homozygous in nearly
    everyone and pad out runs the same way.  Set it to 0 for PLINK 1.x
    behavior.
```

On the LD sanity check you left for later: I do not have a proposal I would defend either. My one thought is that the check people actually need is not "is this set pruned" but "is this run explained by LD", which is a per-run question rather than a per-dataset one, and the cheap version of it, mean r^2 among the variants spanned by each called run, is already most of the cost of `--indep-pairwise` on the same region. So I suspect the honest answer stays a documentation warning until someone has a specific dataset where a cheaper statistic demonstrably separates real runs from LD artefacts.

Same reasoning applies to the `--blocks` defaults question on #374, where I could not get an experiment to converge either.

BenjaminDEMAILLE:
Rebased onto master (through e78ce815); --homozyg takes command bit 34, since --ld-score claimed 33 upstream.

Also added `Tests/TEST_HOMOZYG`, which the command had no coverage for.  It plants runs of homozygosity in a deterministic 30-sample, 6000-variant fileset and compares .hom, .hom.indiv and .hom.summary against PLINK 1.9 under default thresholds, tighter ones, and --homozyg-het; run boundaries and counts have to match exactly, the floating-point columns within tolerance since 1.9 prints three decimals.  The comparison runs pass --homozyg-min-af 0, which is the PLINK 1.x behavior.  It also checks that --threads 1 reproduces the multithreaded result, that 'zs' compresses .hom.summary and leaves .hom alone, and that cols= drops the columns it doesn't name.

chrchang:
After writing the first version of my previous comment, I edited it in place to change --homozyg-maf (or --homozyg-min-af, may as well support both spellings) to not have a default value.  The reasoning is that, if someone *has* tuned the other --homozyg parameter values to their MAF < 0.05 use case, silently filtering down to MAF >= 0.05 is bad.  This is an "ask rather than guess" scenario.

BenjaminDEMAILLE:
Done, pushed.

`--homozyg-min-af` now has no default and is required; `--homozyg-maf` is accepted as a second spelling. Running `--homozyg` without it is an error:

```
Error: --homozyg requires --homozyg-min-af (equivalently, --homozyg-maf).  0.05
is a reasonable value with the other default parameters; 0 reproduces PLINK
1.x, which applied no frequency floor.
```

The help entry gives the reason rather than just the recommendation, since "why is there no default" is the question a user hits first:

```
  * --homozyg-min-af is required, and has no default.  Variants with minor
    allele frequency below it are excluded, since rare variants are
    homozygous in nearly everyone and pad out runs the same way.  0.05 is
    a reasonable value alongside the other defaults; 0 reproduces PLINK
    1.x, which applied no frequency floor.  There is no default because
    the right floor depends on the other parameters: if you have tuned
    them for a low-MAF variant set, silently filtering to 0.05 would
    throw away exactly what you were looking at.
```

`Tests/TEST_HOMOZYG` now also checks that the two spellings agree and that the command refuses to run without one of them. The PLINK 1.9 comparison runs pass `--homozyg-min-af 0`, which is 1.9's behavior.

Issue #58, opened 2017-12-12
[REQUEST] Flip-scan with D'


JoniColeman:
**tl;dr - it would be useful if there were an option for using D' instead of correlation for --flip-scan.** 

--flip-scan is a useful tool in quality control for identifying flipped ambiguous SNPs when component parts of a GWAS have been genotyped separately. It can also be used to identify variants that have been miscalled during genotyping, as they (may) depart from the expected pattern of LD with neighbouring SNPs. To some extent, this is captured by --flip-scan's current correlation-based measure. However, D' might be a more useful measure here, as it is more sensitive to the presence of unlikely haplotypes. As such, it would be useful if --flip-scan had a D' mode as well. 

Ideally, the output would be a variant of that from --flip-scan verbose

```
CHR_INDX     Chromosome code
SNP_INDX     Index variant identifier
BP_INDX      Index variant base-pair coordinate
A1_INDX      Index variant allele 1
SNP_PAIR     Second variant identifier
BP_PAIR      Second variant base-pair coordinate
A1_PAIR      Second variant allele 1
D_PRIME_A    Case-only D'
D_PRIME_U    Control-only D'

```

BenjaminDEMAILLE:
#355 implements the base `--flip-scan` command in plink2, which this request builds on. It is correlation-based, like 1.9's, so it does not answer the D' part; but it is the piece that has to exist first, and adding a D' mode on top of it is then a matter of swapping the per-pair statistic — the neighborhood walk, the case/control split and the reporting are all in place.

Verified field-for-field against PLINK 1.9 on a simulated set with deliberately flipped variants and on a 1000 Genomes chr21 subset, and 4.4-6.6x faster than 1.9 on 5000 samples x 50000 variants.

PR #355, opened 2026-09-01
Add --flip-scan (LD-based case/control strand check)

BenjaminDEMAILLE:
Closes the base-command half of #58: that issue asks for a D' mode for `--flip-scan`, which presumes the command exists, and it does not in plink2.

For each variant, `--flip-scan` computes the correlation with each nearby variant separately in cases and in controls, and reports the neighbors whose correlation changes sign. That is the signature of a strand-ambiguous variant which was resolved differently in two separately-genotyped batches.

`--flip-scan ['verbose'] ['zs']`, plus `--flip-scan-window`, `--flip-scan-window-kb` and `--flip-scan-threshold`, with 1.9's defaults (10 variants, 1000 kb, r = 0.5). Output is `.flipscan`, and `.flipscan.verbose` with `verbose`; both tab-delimited with a header line, carrying 1.9's columns.

**No new per-ISA code.** The correlation kernel is the one `--indep-pairwise` already uses: `SplitHomRef2het()` plus `DotprodWords()`/`SumSsqWords()` over the {+1, 0, -1} recoding. So this inherits the existing AVX2, SSE4.2, SSE2 and NEON paths as-is. `plink2_ld.cc` compiles clean under all four (x86-64 `-msse2`, `-msse4.2`, `-mavx2 -mbmi -mbmi2 -mlzcnt -mfma`, and native arm64).

**Threading.** Each variant's result depends only on its own window, so the scan parallelizes over variants: the reader loads a block plus its two halos, and the workers run three phases over it — recode into the two subsets, fill the r cache, tally — with a barrier between phases, since each needs the previous one finished for variants another thread owns. Computing each pair once and reading it from both endpoints halves the correlation work. Output does not depend on `--threads`.

5000 samples x 50000 variants over 22 chromosomes, best of 5 warm runs, 16-core M1:

| | PLINK 1.9 | plink2 | |
|---|---:|---:|---|
| default window (10) | 0.279s | **0.063s** | 4.4x |
| `--flip-scan-window 50` | 0.737s | **0.111s** | 6.6x |

**One deliberate difference from 1.9.** A neighbor pair where either group is monomorphic has an undefined correlation. 1.9 lets the resulting `nan` through, and since every comparison against a nan is false, the pair ends up counted as a *sign flip* — so `--flip-scan` reports `NEG=1` with `R_NEG=nan` for it. That is a false positive, so such pairs are skipped here. On a 2504-sample 1000 Genomes chr21 subset it affects 10 of 600 variants, all with MAF below 0.001, and none once `--maf 0.01` is applied. The help text says so.

**Verification** against PLINK 1.9, field by field. On a 4000-variant 800-sample simulated set with LD blocks and 200 deliberately flipped variants (all 200 detected), `POS_CT`, `R_POS`, `NEG_CT`, `R_NEG` and the flipped-neighbor ID lists are identical across seven configurations — default, `--flip-scan-window 5` and `40`, `--flip-scan-window-kb 50`, `--flip-scan-threshold 0.2` and `0.8`, and `--maf 0.05` — as are all 3540 rows of the verbose output. Same on the chr21 subset apart from the monomorphic-pair rows above. The only remaining column difference anywhere is `ALT_FREQ`, where 1.9 prints 3 significant digits.

`2.0/Tests/run_tests.sh` passes; `--freq`, `--make-bed`, `--glm`, `--indep-pairwise`, `--r2-unphased`, `--hardy` and `--het` are byte-identical to a `master` build.


chrchang:
Generally looks good, just need to define plink2-style column sets now.  Thanks for calling out the issue with PLINK 1.9's handling of monomorphic variants, I will probably change PLINK 1.9's behavior there.

BenjaminDEMAILLE:
Column sets added, same `--mendel`-style single descriptor as #354, covering both output files:

```
chrom, pos            (also control the .flipscan.verbose columns)
ref, alt
altfreq, posct, rpos, negct, rneg, negids   (.flipscan only)
```

The variant IDs and the two correlations the verbose file exists to report (R_CASE, R_CTRL) are always present, so `cols=` cannot empty it out. Default is all of them, i.e. what PLINK 1.9 emits.

I also reordered the .flipscan columns to plink2's convention: POS ahead of ID, REF ahead of ALT. That changes the output rather than just adding to it, so flagging it explicitly. Values are unchanged; all seven key fields still match 1.9 row for row across the 4000-variant test fileset (1.9's A1/A2 being plink2's ALT/REF).

On the monomorphic-variant point: glad it is useful. To restate it in one line for whoever picks it up, PLINK 1.9's `flipscan_chrom_thread()` lets a `nan` correlation through when one group is monomorphic, and since `nan < 0` is false but the sign test is written as a comparison, such a pair ends up counted as a sign flip.

chrchang:
I have updated PLINK 1.9's nan-handling in its latest release.

But when reviewing the broader context, I can understand why nobody noticed the unwanted difference for PLINK 1.07 in >10 years: the existing --flip-scan should never be used on variants with very low MAF in the source population, since simply checking for flipped allele frequency between cases and controls gets the job done.  And most of the time, a user really wants to perform a comparison of a single dataset of their own against a reference dataset, rather than two parts of their own dataset; forcing them to merge their dataset with a reference before using this command is not great.

So I now think this command needs a rework.  In addition to the original mode of comparing one subset of the initial dataset against the rest, it should have a mode to compare the full initial dataset against overlapping variants in a second reference dataset (similar to --pgen-diff), and a mode which just compares against a reference allele frequency file (--read-freq infrastructure can be reused here).  The updated algorithm should look something like:

0. If we aren't just performing an allele-frequency comparison, require at least 50 founders in each group we're computing LD in, unless --bad-ld was specified.
1. If major-allele (as defined in the initial dataset) frequency differs by more than a threshold (default >50% if we're following up with the LD scan, >20% if we're only comparing against a reference-allele-frequency file), mark the variant as PROBLEM=Y.
2. Perform the original --flip-scan check on the remaining variants, with an additional filter: only variants with major-allele frequency below a threshold (default 0.9), and which weren't marked as problematic in step 1, participate.  The old --flip-scan-window default value of 10 can be kept... but let's make those 9+9 pairwise comparisons count, instead of wasting them against tiny-MAF variants which yield little information when compared against.

There will need to be additional CASE_MAJ_FREQ, CTRL_MAJ_FREQ, and PROBLEM columns.  There can be a 'ref-allele-based' modifier which causes REF to be used in place of MAJ (including in the output-file header).

BenjaminDEMAILLE:
Rebased onto master (through e78ce815); --flip-scan takes command bit 36, since --ld-score claimed 33 upstream.  Squashed to a single commit: the incremental history stopped applying cleanly once --indep landed in the same plink2_ld.cc region.  Re-verified against PLINK 1.9 after the rebase.

BenjaminDEMAILLE:
Added `Tests/TEST_FLIP_SCAN`. It plants 20 haplotype blocks and swaps the alleles of every 7th variant in the cases only, which is the strand inconsistency the scan exists to find, then compares .flipscan against PLINK 1.9 under the defaults, --flip-scan-window, --flip-scan-window-kb and --flip-scan-threshold, plus .flipscan.verbose under 'verbose'. All 130 flagged variants and all 408 verbose neighbor pairs match, and --threads 1/3/8 agree.

Two conventions had to be normalized in the comparison, both stemming from which allele the report is relative to: 1.9's F is the minor-allele frequency and plink2's is the ALT frequency, and recoding one variant of a pair flips the sign of both correlations, so the verbose file is compared on the magnitudes plus whether the two correlations agree in sign (the quantity the scan is actually about).

One thing the sweep does not cover, which you may want to know about: **small --flip-scan-window-kb values diverge.** On this fixture (variants 2 kb apart within a block, --maf 0.05 applied) PLINK 1.9 reports no neighbors at all at 40 kb and below, while plink2 truncates the window and keeps going:

| --flip-scan-window-kb | 1.9 variants with POS_CT > 0 | plink2 |
|---:|---:|---:|
| 30 | 0 | 107 |
| 40 | 0 | 107 |
| 50 | 107 | 107 |
| 100 and up | 107 | 107 |

The transition sits right about where the kb window stops being able to hold a full --flip-scan-window (10 variants), which makes it look like 1.9 gives up rather than truncating. plink2's behavior seems the more useful of the two, but it is a difference, so the test sweeps only the values where they agree and this is on record.

chrchang:
The small --flip-scan-window-kb behavior sounds like a PLINK 1.9 bug.  I will investigate.

But in the meantime, see the --flip-scan redesign comment above.  PLINK 1.x --flip-scan was useful for a hard QC case but ignored the easier cases, so it isn't used that often and when it is used it is sometimes misinterpreted.  --flip-scan should instead handle both the easy and hard cases in a report that has a straightforward interpretation.

(To improve the false-positive / false-negative balance, PROBLEM=Y could require at least *two* negative-LD-matches in the hard case.)

BenjaminDEMAILLE:
Working through the redesign. The mechanics fit the existing scan better than I expected, so most of the algorithm is cheap; what I would like to settle first is the report's shape, since two of the questions below change how the output layer is built and I would rather not write it twice.

**What falls out for free.** The scan already fills a `VariantAggs` per group per variant (`nm_ct`, `sum`, `ssq`) at load time, and the REF allele count is exactly `nm_ct + sum`. So the per-group major-allele frequencies step 1 needs cost nothing extra: `CASE_MAJ_FREQ` and `CTRL_MAJ_FREQ` come straight out of the aggregates already being computed, and eligibility for step 2 can be decided in the same pass, before any correlation is computed. The `halo` means a variant's partners are loaded before it is used as an index, so an eligibility flag set at load time is available in time to filter both roles.

**Unambiguous, implementing now:** the >=50-founders-per-group gate (skipped when only comparing frequencies, overridable with `--bad-ld`), the step-1 frequency-difference threshold (0.5 with the LD follow-up, 0.2 without), the major-allele-frequency < 0.9 participation filter, requiring at least two negative-LD matches for PROBLEM=Y in the hard case, the `CASE_MAJ_FREQ`/`CTRL_MAJ_FREQ`/`PROBLEM` columns, and `ref-allele-based` swapping MAJ for REF in both the values and the header names.

**Four questions:**

1. **Row policy.** Should the report have one row per scanned variant, including those flagged in step 1 (LD columns then being NA, since they did not participate) and those excluded by the 0.9 filter? Or stay closer to the current report and only emit variants that reached the LD scan? A `PROBLEM` column reads like the former to me, but that makes the output no longer a pure product of the scan.

2. **"Participate" in step 2.** Does a variant excluded by step 1 or by the 0.9 filter also stop being usable as a *comparison partner* for other variants, or only as an index variant? I read your wording as both, which is also what makes the 9+9 comparisons count for something, but it is worth pinning down.

3. **Shape of the reference-frequency mode.** With only a frequency file there is no case/control split and no LD, so I assume it degenerates to one row per variant with the dataset's major-allele frequency, the reference's, and PROBLEM from the 0.2 threshold alone. Are the column names then something like `MAJ_FREQ`/`REF_MAJ_FREQ`, or do you want the CASE/CTRL names reused?

4. **Priority of the second-dataset mode.** This is the largest of the three (it needs the `--pgen-diff`-style reader), and the frequency-file mode already covers "compare my dataset against a reference" for the common case. Would you rather I land the within-dataset and frequency-file modes first and add the second-dataset mode in a follow-up, or hold the whole thing until all three are in?

Happy to just pick defaults on all four and send it if you would rather review code than prose; say the word and I will go with (1) one row per variant, (2) both roles, (3) `MAJ_FREQ`/`REF_MAJ_FREQ`, (4) defer the second-dataset mode.

BenjaminDEMAILLE:
First half pushed, going with the defaults I offered above since I would rather have code to react to than keep asking. Steps 1 and 2 are in, with the new columns, the `ref-allele-based` modifier, the 50-founders-per-group gate (overridable with `--bad-ld`), and the participation filter. The two extra comparison modes are not in yet.

```
#CHROM POS ID REF ALT CASE_MAJ_FREQ CTRL_MAJ_FREQ POS_CT R_POS NEG_CT R_NEG PROBLEM NEG_IDS
```

**Three new parameter flags**, all with the defaults you named: `--flip-scan-freq-diff` (0.5), `--flip-scan-max-maj-freq` (0.9), `--flip-scan-min-neg` (2).

**On what "participate" means**, I went with both roles: a variant flagged in step 1, or above the 0.9 major-allele frequency, is skipped as an index variant *and* as a comparison partner. That is what makes the 9+9 slots count for something, which was the point of the filter.

**One implementation note worth having**: the per-group frequencies turned out to be free. The scan already fills a `VariantAggs` per group per variant, and the REF allele count is exactly `nm_ct + sum`, so `CASE_MAJ_FREQ`/`CTRL_MAJ_FREQ` and the whole step-1 decision fall out of aggregates that were being computed anyway. Eligibility is settled in the load phase, which finishes before any correlation is computed, so filtering both roles needed no extra pass.

Both groups report the frequency of the *same* allele -- the major one across the whole dataset -- so that a difference between the columns is a difference in the data rather than a difference in which allele got picked per group.

**Testing.** The 1.9 comparison is still there and still exact: `--flip-scan-freq-diff 1 --flip-scan-max-maj-freq 1 --flip-scan-min-neg 1` is precisely PLINK 1.x's algorithm, so the LD half is checked against 1.9 as before, and the new behavior is tested separately (that a step-1 flagged variant really does drop out of the scan, that two-negative-matches can only reduce the number of calls, that `ref-allele-based` changes nothing but the two frequency columns, and that the founder gate fires and `--bad-ld` lifts it).

Still open, and I will do these next unless you redirect: the reference-frequency mode, then the second-dataset mode. Question 3 above (column naming when there is no case/control split) is the only one still worth an answer before I start.

BenjaminDEMAILLE:
Second mode pushed: `--flip-scan-ref-freq`.

```
plink2 --pfile mine --flip-scan --flip-scan-ref-freq panel.afreq
#CHROM POS ID REF ALT MAJ_FREQ PANEL_MAJ_FREQ PROBLEM
```

It reads the file with the same `ReadAlleleFreqs()` `--read-freq` uses, so the accepted formats are the ones users already have, and it does the allele reconciliation. Both sides report the frequency of the same allele -- the major one in the loaded dataset -- so a difference is a difference in the data rather than in which allele each side picked. A variant the file does not cover gets `PROBLEM=NA` rather than a call, with a count in the log. Threshold defaults to 0.2 here rather than 0.5, per your spec, and the mode needs no case/control phenotype, no founders and no sorted coordinates.

**On the third mode, I have deliberately stopped rather than kept going**, and I would rather say why than have it look unfinished.

The design is settled and it is not the hard part: the existing scan already handles two groups with different sample counts, so "group 0 = reference dataset, group 1 = mine" is a change to the loader only -- read group 1 from reader 1 and group 0 from reader 2 instead of splitting one read into two subsets. The rest (correlate, tally, output) is untouched. The work is the second-fileset plumbing: a second `ChrInfo` + `LoadPvar`, opening the second .pgen, matching variants by ID, and reconciling alleles including the REF/ALT swap case, which is delicate precisely because a swap is also one of the things being detected.

That is roughly 500 lines, and this PR already has two substantial changes you have not seen yet. Stacking a third on top before you have looked at either makes it harder to review, not easier. So it is going in a follow-up PR once you have had a look at these two, unless you would rather have it here.

The only thing I would still like an answer on is question 3 from earlier: I went with `MAJ_FREQ`/`PANEL_MAJ_FREQ` (and `REF_FREQ`/`PANEL_REF_FREQ` under `ref-allele-based`) for the frequency-file mode, since there is no case/control split to name. Easy to change if you would rather they read differently.

Tests cover the new mode end to end: exact agreement against the dataset's own frequencies, a deterministic 0.3 shift on every fifth variant caught with no false positives on the rest, `PROBLEM=NA` for variants the file omits, that it runs without a phenotype, and that the flag is rejected without `--flip-scan`. The PLINK 1.9 comparison for the LD half is still exact via the compatibility settings.

BenjaminDEMAILLE:
Follow-up with the second-fileset mode is up as #389, stacked on this branch.

chrchang:
To answer your column-header question: either your current PANEL_MAJ_FREQ / PANEL_REF_FREQ, or mapping current dataset -> CASE_ and reference dataset -> CTRL_ is fine with me.

Meanwhile, since multiple binary phenotypes may now be loaded, there should be a way to specify the phenotype name, and the command should error out if no phenotype name is provided and multiple binary phenotypes are defined.

Optional, could be in a separate PR: also support categorical phenotypes, where the user provides a categorical phenotype name and a group of category name(s) to treat as cases.

chrchang:
oops, clicked merge on this one when I intended to merge another PR.

chrchang:
Content restored at #394 .

BenjaminDEMAILLE:
Noted on all three, and no problem about the accidental merge.

- **Column headers**: I will keep the current `PANEL_MAJ_FREQ` / `PANEL_REF_FREQ` spelling then, since it reads better than `CTRL_` for something that is not a control group.
- **Phenotype name**: agreed, and the same request came up on #371 for `--twolocus`, so I will do them the same way -- an optional phenotype name, an error when none is given and more than one binary phenotype is loaded, and no implicit case/control split otherwise. I will send it as a follow-up once #394 is in, to avoid stacking on a branch that is mid-restore.
- **Categorical phenotypes**: happy to add that too, as its own PR after the binary case.

#389 (compare against a second fileset) is still open and stacked on this branch's content, so it will need a rebase once #394 lands; say the word if you would rather I retarget it at that branch now.

BenjaminDEMAILLE:
Both halves of what you left open in #394 are up:

- **#389** rebased on the restored flip-scan. It is now a clean 610-line diff containing only the second-fileset mode, since the four flip-scan commits it was stacked on are in master.
- **#398** implements the phenotype-name parameter, replacing ede5d76d's placeholder error. The name stays optional when exactly one case/control phenotype is loaded.

Both green.

PR #357, opened 2026-09-01
Add --lfile/--lgen (long-format import)

BenjaminDEMAILLE:
plink2 can neither read nor write PLINK 1.x long format. (I listed this in #344 as a write-only asymmetry; that was wrong — `--export lgen` is accepted as a format name but falls through to "have been implemented so far", same as `--export 23` did.) This adds the read direction.

`--lfile [prefix]` and `--lgen <filename>`, plus `--reference` and `--allele-count`, matching 1.9's flags.

Unlike .ped and .tped, .lgen genotypes arrive one (sample, variant) call per line and in no particular order, so the whole matrix is held in memory before it can be written. PLINK 1.9 has the same constraint ("does not support multipass writes"), and the importer says so when it runs out of room. Everything else reuses the existing legacy-import machinery: `LoadMap()` for the .map, `LoadPsam()`/`RewritePsam()` for the .fam, `MapToPvar()` for the .pvar, and the same allele-assignment logic `PedmapToPgen()` uses, since .lgen has the same "alleles are whatever strings the file happens to contain, at most two per variant" problem.

REF/ALT follows PLINK 1.x: the more common allele becomes A2/REF, counting only nonmissing calls, with the first-seen assignment kept on an exact tie. With `--reference` the file decides instead, its optional third column names the other allele, and calls absent from the .lgen are homozygous-reference rather than missing.

**Verification** against PLINK 1.9, on a 1000-sample 1800-variant set (1.8M genotype calls, 3% missing): `.bed` byte-identical and `.bim` identical for

- `--lfile`,
- `--lgen` + `--map` + `--fam` given separately,
- compound genotypes (both alleles in one token),
- `--reference`.

And byte-identical for `--allele-count` on a fully-genotyped 300-sample 800-variant set, including the case where the reference file names the minor allele rather than the major one (which is where a frequency-based fallback would have diverged).

`2.0/Tests/run_tests.sh` passes; `--freq`, `--make-bed`, `--glm` and `--hardy` are byte-identical to a `master` build.

Two small parser changes were needed so the accompanying flags can be combined: `--map` and `--psam`/`--fam` now accept `--lfile`/`--lgen` as the companion input flag. `--reference` and `--allele-count` are validated after the parse loop rather than during it, since plink2 sorts flags alphabetically and they would otherwise be seen before `--lfile`.


BenjaminDEMAILLE:
Added the export direction too (`185d8681`), since it turned out to be missing as well and you have said you prefer implementing both at once.

`--export lgen` writes .lgen + .map + .fam; `lgen-ref` adds the .ref file naming each variant's reference and alternate allele and omits the calls that are homozygous for the reference allele, matching 1.9's `--recode lgen-ref`. Lines are `<FID> <IID> <variant ID> <allele 1> <allele 2>`, allele 1 first as in PLINK 1.x, tab-delimited like plink2's other legacy output.

Verified against 1.9 on the same 1000-sample 1800-variant set: the .lgen has the same 1800000 lines with identical fields, and lgen-ref the same 851496 lines (compared as a set, since the order differs) with an identical .ref.

And the round-trip check you mentioned on #349 is now available here, in all four combinations — plink2's export read back by plink2 and by PLINK 1.9, for both `lgen` and `lgen-ref` — each reproducing the original .bed byte-for-byte.

chrchang:
Since very few people use this flag in PLINK 1.9, there isn't much value in rushing an implementation that is still bound by PLINK 1.9's genotypes-must-fit-in-memory constraint.

PLINK 2.0 already spills data to temporary files in several scenarios where PLINK 1.x does not; this is another case where that's the right tradeoff to make.

BenjaminDEMAILLE:
Understood, and I agree the in-memory constraint is the wrong thing to inherit. I will convert this to spill to temporary files rather than requiring the whole genotype matrix to fit, and reopen when it does; there is no point reviewing the current shape further.

For reference while it sits: the correctness work in it is independent of that change and still stands (the `VariantIdDupflagHtableFind()` and `ZeroTrailingNyps()` fixes, and the allele-order rule), so the rewrite is about the buffering strategy rather than the parsing.

BenjaminDEMAILLE:
Added `Tests/TEST_LGEN`, and it caught a crash.

**Unaligned per-variant genotype slices.** The importer holds one genotype vector per variant in a single allocation and spaced the slices by the packed word count, so on any sample count whose packed and aligned word counts differ, every slice at an odd index was not vector-aligned and `GenovecInvertUnsafe()` tripped its assertion:

```
Assertion failed: (IsVecAligned(genovec)), function GenovecInvertUnsafe, file pgenlib_misc.cc, line 1094.
```

200 samples is enough (7 words packed, 8 or 16 aligned), and `--lfile <prefix> --make-bed` aborted outright. Fixed by spacing the slices with `NypCtToAlignedWordCt()`. The test fixture uses 200 samples so this stays covered.

**What the test checks:** the .bed round trip through `--export lgen`, agreement with PLINK 1.9 reading the same .lgen, `--lgen` with .map/.fam named separately, `--reference` against `--export lgen-ref` (including that the reference-coded file is smaller, which is the point of it), and that a missing .map is an error rather than a silent empty result.

This is the same bug class as the one in #375, so it may be worth a look at any other place that slices one allocation into per-variant genotype vectors.

PR #358, opened 2026-09-01
Add --simulate and --simulate-qt

BenjaminDEMAILLE:
The last data-generation command plink2 was missing. `--simulate` builds a case/control dataset from a PLINK 1.x simulation parameter file, `--simulate-qt` a quantitative-trait one, with `--simulate-ncases`, `--simulate-ncontrols`, `--simulate-prevalence`, `--simulate-n`, `--simulate-label` and `--simulate-missing`, and the `tags`/`haps` and `acgt`/`1234`/`12` modifiers.

Relevant to the test suite: `2.0/Tests/run_tests.sh` calls `plink --simulate` to build the input for three of its eight tests, so plink2 could not generate its own test data. (Those three still compare against 1.9 afterwards, so this does not remove the dependency outright — it just means the *input* no longer has to come from another binary.)

**Same seed, same dataset.** The frequency-table helpers are ported from 1.9's `simulate_init_freqs_{cc,qt}()` and `simulate_cc_get_conditional_probs()`, and the generator issues exactly the same sequence of random-number calls, including the separate 64-bit SFMT seeded from the main one. Genotypes are built in PLINK 1 coding, as 1.9 does, so the "more common allele becomes A2" test matches bit-for-bit, and are converted to plink2's coding when the .pgen is written. Output is .pgen + .pvar + .psam, plus the .simfreq file 1.9 writes.

**Verification**, against PLINK 1.9 with `--seed 11`, on twelve configurations — default, `--simulate-missing`, non-default case/control counts, `--simulate-prevalence`, `--simulate-label`, `acgt`, `1234`, `12`, `tags`, `haps`, and `--simulate-qt` with and without missingness. In every one, the .bed obtained by converting plink2's output is **byte-identical** to 1.9's, and the .bim, .fam and .simfreq match field for field. That includes the quantitative-trait phenotypes, which come from the same normal-deviate stream.

`2.0/Tests/run_tests.sh` passes; `--freq`, `--make-bed`, `--glm` and `--hardy` are byte-identical to a `master` build, as is `--dummy` with a fixed seed, which confirms the shared RNG state is untouched.

If you would rather not carry a bit-compatible port and would prefer a fresh generator with plink2 conventions, say so — the exact-match property is what made it verifiable, but it is not something the feature needs.


chrchang:
Implementation of --simulate{,qt} was deferred since it looked like few people (outside of myself in the bit of test code you called out) used it; there seemed to be some work on more realistic ways to generate this type of data; and --dummy, which is somewhat fancier in PLINK 2.0 than 1.9, takes care of the unrealistic case.  However, I haven't surveyed this subfield recently.

BenjaminDEMAILLE:
That is fair, and matches what I found: outside of PLINK's own test code the flag barely appears, and where it does, the generative model is the unrealistic one that --dummy already covers better in PLINK 2.

I will leave this open but treat it as low priority rather than pressing it. If you would rather not carry the surface area at all, say so and I will close it; the implementation is verified against 1.9 (same SFMT stream, same conditional-probability derivation) so it will keep just as well unmerged.

PR #359, opened 2026-09-01
Add --test-missing

BenjaminDEMAILLE:
Differential missingness: is a variant's missingness rate different between cases and controls? A QC test with no plink2 equivalent, and the last item on the Tier 1 list from #344.

`--test-missing ['midp'] ['zs']` writes a `.missing` report with `F_MISS_A`, `F_MISS_U` and `P`. The 2x2 table is (missing, nonmissing) x (case, control) and the test is Fisher's exact, as in PLINK 1.x. Heterozygous haploid calls count as missing, and variants with no missing calls — or nothing but missing calls — are omitted, both matching 1.9.

**This reintroduces the question from #351, so please read that part before the code.** `FisherExact2x2()` is the `fisher22()` port from that PR, and plink2 still has no other Fisher implementation. Two reasons I am putting it in front of you anyway rather than assuming the earlier answer carries over:

- The objection there was that there is no dedicated rare-variant analysis module for an *allelic exact test* to live in. This is a QC check on missingness counts, in the same family as `--missing` and `--hardy`, not an association test on genotypes.
- If you disagree, the fix is small and local: the p-value computation is the single `FisherExact2x2()` call in `TestMissingReport()`, isolated in `include/plink2_stats.cc`. Swapping in the implementation from https://github.com/chrchang/stats when it lands, or replacing it with a chi-square statistic, changes one line here.

Happy to do either, or to close this, on your word.

**Verification** against PLINK 1.9, on 5000 samples x 50000 variants and on 2000 samples x 100000 variants, both with and without `midp`: identical row sets, and `F_MISS_A`, `F_MISS_U` and `P` all agree to 1.9's printed precision, with p-values matching to better than 1 part in 10^12. Datasets with no missing calls produce a header-only report in both programs.

`2.0/Tests/run_tests.sh` passes; `--freq`, `--make-bed`, `--glm`, `--hardy` and `--missing` are byte-identical to a `master` build.


chrchang:
Oh, I had forgotten about --test-missing's use of the Fisher 2x2 computation.  I will update include/plink2_stats.cc with the implementation from https://github.com/chrchang/stats within the next few days then, and make a comment on this PR when I have done so.

chrchang:
Fisher22TwoSidedP is now available in include/plink2_stats .

BenjaminDEMAILLE:
Switched over, rebased onto 8bfebe8b. `Fisher22TwoSidedP()` replaces the `FisherExact2x2()` this PR was adding, so the diff no longer touches include/plink2_stats at all: 102 lines removed, and the PR is now confined to the four files that implement the flag.

Results are unchanged. Against PLINK 1.9 `--test-missing` on 100000 variants, the maximum relative difference in P is 4.2e-4 two-sided and 3.9e-4 with `midp`, both under the 5e-4 floor imposed by 1.9 printing four significant digits; variant IDs and both missingness frequencies match exactly on every row.

2.0/plink2_misc.cc
```diff
@@ -10763,6 +10763,175 @@ PglErr HetCalcMain(const uintptr_t* sample_include, const uintptr_t* variant_sub
...
+          chr_end = cip->chr_fo_vidx_start[chr_fo_idx + 1];
+        } while (variant_uidx >= chr_end);
+        const uint32_t chr_idx = cip->chr_file_order[chr_fo_idx];
+        char* chr_name_end = chrtoa(cip, chr_idx, chr_buf);
+        *chr_name_end++ = '\t';
+        chr_blen = chr_name_end - chr_buf;
+        is_x = (chr_idx == x_code);
+        is_fully_haploid = IsSet(cip->haploid_mask, chr_idx) && (!is_x);
+      }
+      reterr = PgrGet(sample_include, pssi, sample_ct, variant_uidx, simple_pgrp, genovec);
```
chrchang: See PgrGetMissingnessD().

chrchang:
There should be an option to look at dosage missingness.  --missing already implements something similar with optional column sets.

2.0/plink2_misc.cc
BenjaminDEMAILLE: Switched to it. The loop now calls

```
PgrGetMissingnessD(sample_include, pssi, sample_ct, variant_uidx, simple_pgrp, missing_bv, nullptr, need_hets? het_bv : nullptr, genovec_buf);
```

so the multiallelic/phase/dosage tracks are skipped instead of decoded, and the `PgrGet()` + `ZeroTrailingNyps()` + `SetHetMissing()` + `GenoarrToMissingnessUnsafe()` sequence collapses into it. `hets` is only requested on chrX and the fully-haploid chromosomes, where it is OR'd into the missingness vector (masked to males on chrX).

Comparing against PLINK 1.9 while I was there turned up a real bug: on chrY, PLINK 1.x looks at males only, in the numerators *and* the denominators. I had been counting females as missing, so every chrY row was wrong. Fixed by masking the missingness vector to males and using male case/control counts as the denominators there; chrY variants are skipped outright when there are no male cases or no male controls.

Added `Tests/TEST_TEST_MISSING`, which compares `--test-missing` and `--test-missing midp` against PLINK 1.9 over chr1/chrX/chrY, plus a .pgen round trip and the `zs` output. On a 5000-sample, 200000-variant fileset the two reports agree on all 200000 rows, and plink2 takes 0.38s against PLINK 1.9's 0.90s.

One limitation worth naming: `PgrGetMissingnessD()` derives `hets` from the raw genovec in this path, so an ALT1/ALT2 heterozygous hardcall on a haploid chromosome is not seen as a het-haploid. That matches what `--missing`'s hethap count already does, so I left it alone; say the word if you'd rather it be exact here.

BenjaminDEMAILLE:
Added in the latest push.

**`dosage` modifier.** `PgrGetMissingnessD` already returns both bitvectors, so the switch is just which one the test reads; the default stays hardcall missingness to match PLINK 1.x. On a fileset where every call carries a dosage but `--hard-call-threshold 0.1` erases most hardcalls, the two are far apart:

```
                    F_MISS_A   F_MISS_U
default (hardcall)  0.817391   0.835294
dosage              0.078261   0.047059
```

and the per-variant totals agree exactly with what `--missing vcols=nmiss,nmissdosage` reports for the same file, which the test suite now checks row by row.

**Column sets.** `cols=` accepts `chrom`, `pos`, `ref`, `alt1`, `alt`, `maybeprovref`, `provref`, `nmissa`, `nobsa`, `fmissa`, `nmissu`, `nobsu`, `fmissu`, `p`; the default is `chrom,pos,fmissa,fmissu,p`, which is byte-identical to the previous output.

The counts and denominators are the addition I found most useful: on chrY the denominator is the male case/control count rather than the sample count, and with only the rates printed there is no way to see that from the report. `nobsa`/`nobsu` make it explicit.

One judgment call: `maybeprovref` is in the default set but, as in `--hardy`, it only prints when the REF column is also requested, so the default report is unaffected.

2.0/plink2_misc.cc
chrchang: I just tried creating a single-sample (male), single-variant chrY dataset with GT=1/2, and "--missing vcols=+hethap" did report it as a het-haploid?

BenjaminDEMAILLE: You are right, and my note was wrong. I reproduced your case: `--missing vcols=+hethap` does report an ALT1/ALT2 chrY call as a het haploid. `--test-missing` was the one getting it wrong, so this was a real bug rather than a shared limitation.

The cause is what I described, applied to the wrong program: `PgrGetMissingnessD()` derives its het bitvector from the raw genotype vector, where an ALTx/ALTy call is encoded as a homozygote with a patch_10 entry, so those calls were invisible here.

Fixed. For multiallelic variants on chrX and the fully-haploid chromosomes, the multiallelic track is read and a patch_10 entry whose two allele codes differ is added to the het bitvector. Biallelic variants keep the single-read fast path, so the common case is unaffected.

On a four-male chrY fileset with one ALT1/ALT2 call and one REF/ALT call in the same sample:

```
before:  Y  yv_refalt  1  2  0  2      (yv_alt12 absent entirely --
                                        dropped as having no missing calls)
after:   Y  yv_alt12   1  2  0  2
         Y  yv_refalt  1  2  0  2
```

which is what `--missing` reports for both. Regression test added.

chrchang: See lines 9567-9592 of pgenlib_read.cc .  PgrGetMissingnessD() should already be inspecting the multiallelic-hardcall data track and patching the returned hets[] bitvector.

BenjaminDEMAILLE: It does, but only on one of the two branches, and `--test-missing` takes the other one.

`PgrGetMissingnessD()` splits on

```c
const uint32_t dosage_is_relevant = missingness_dosage && VrtypeDosage(vrtype);
const uint32_t need_to_skip_aux1or2 = dosage_is_relevant && (vrtype & 0x18);
if (!need_to_skip_aux1or2) {
  PglErr reterr = ReadMissingness(..., missingness_base, hets, genovec_buf);
  ...
  if (reterr || (!dosage_is_relevant)) {
    return reterr;
  }
} else {
  // ... the aux1b het patching you pointed me at is in here
}
```

The lines 9567-9592 block is inside the `else`, so it is reached only when `missingness_dosage` is non-null and the variant actually carries dosages. In hardcall mode `--test-missing` passes `missingness_dosage == nullptr`, so it goes through `ReadMissingness()`, which carries its own note:

```c
// Currently assumes no phase or multiallelic hardcalls.
// tried to have more custom code, turned out to not be worth it
...
  if (hets) {
    PgrDetectGenoarrHetsUnsafe(genovec_buf, NypCtToWordCt(sample_ct), hets);
  }
```

i.e. hets straight off the raw genotype vector, no aux1b patch. Removing my block makes `Tests/TEST_TEST_MISSING` fail: the ALT1/ALT2 chrY variant drops out of the report entirely.

Your `--missing vcols=+hethap` check passed because the variant-level hethap count comes from `PgrGetMDCounts()` in `LoadAlleleAndGenoCounts()`, which does handle the multiallelic track. The **sample-level** count goes through `LoadSampleMissingCts()` -> the same hardcall branch above, and it has the gap. On a four-male chrY fileset where m1 carries both a `1/2` and a `0/1` call:

```
$ plink2 --pfile tmp_hh --missing scols=maybefid,nmiss,nmisshh,hethap,nobs vcols=chrom,hethap
#CHROM  ID          HETHAP_CT
Y       yv_alt12    1
Y       yv_refalt   1

#IID  MISSING_CT  MISSING_AND_HETHAP_CT  HETHAP_CT  OBS_CT
m1    0           1                      1          2
```

Two het-haploid calls in the variant report, one attributed to the only sample that has them.

So the fix probably belongs in the hardcall path rather than in `--test-missing`: patching `hets` there would fix `--missing`'s per-sample hethap counts too and let me delete this block. The cost is that every `--missing` run would then read the aux1b track for multiallelic variants, which is what your "turned out to not be worth it" note suggests you already weighed. Tell me which you would rather have and I will do it that way -- happy to move it into pgenlib, or to leave the reader alone and keep the correction local here.

chrchang: Ok, this deserves a pgenlib fix, need_to_skip_aux1or2 needs to be set differently when hets is non-null.  I will address this now.

chrchang: See commit 3881170 .

BenjaminDEMAILLE: Thanks -- that is the right place for it, and I have dropped my local block on this branch.

Heads-up though: 3881170 segfaults on master for an unphased multiallelic fileset. Minimal repro:

```
##fileformat=VCFv4.2
##contig=<ID=Y>
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	m1	m2	m3	m4
Y	1	yv_alt12	A	G,T	.	.	.	GT	1/2	0/0	0/0	0/0
Y	2	yv_refalt	A	G	.	.	.	GT	0/1	0/0	0/0	0/0
```

```
plink2 --vcf tmp_hh.vcf --make-pgen --out tmp_hh
# (males, chrY)
plink2 --pfile tmp_hh --missing scols=maybefid,nmiss,nmisshh,hethap,nobs --out c1
Segmentation fault
```

`--missing variant-only` is fine; it is the per-sample side that goes down. Bisects cleanly: `3881170~1` produces the report, `3881170` crashes.

Crash frame is `PgrDetectGenoarrHets()` -> `PackWordsToHalfwordsInvmatch()` writing to a null `dst`, i.e. `pgrp->workspace_all_hets`. That buffer is only allocated under

```c
if (gflags & kfPgenGlobalHardcallPhasePresent) {
  pgrp->workspace_all_hets = ...;
  pgrp->workspace_subset = ...;
}
```

and the new `(hets && (vrtype & 8))` term routes unphased multiallelic variants into the branch that uses it.

I tried the obvious patch -- adding `|| (max_allele_ct > 2)` to both the sizing and the allocation -- and it still crashed, so there is something else going on that you will spot faster than I will. Not opening a PR against it for that reason; just flagging since it is on master now and `--missing` is a common command.

(Unrelated: this branch is rebased on current master, and the `--test-missing` workaround is gone, so it now relies on your fix. I will re-push once the crash is sorted, since its test suite exercises exactly this case.)

chrchang: Oops, good job covering this with your tests.  Try ce23b38 .

BenjaminDEMAILLE: ce23b38 does it. Confirmed on the repro, and it fixes the per-sample count too, which was the second half of what I reported:

```
#IID  MISSING_CT  MISSING_AND_HETHAP_CT  HETHAP_CT  OBS_CT
m1    0           2                      2          2
```

m1 carries both a `1/2` and a `0/1` chrY call; that was 1 before, against a variant report that said 2.

This branch is rebased on current master with the local multiallelic block removed, so `--test-missing` now leans on `PgrGetMissingnessD()` as intended. `Tests/TEST_TEST_MISSING` passes unchanged with the block gone, which is the check that it is genuinely redundant rather than merely quiet.

Issue #338, opened 2026-08-30
Roadmap: what remains before 2.0 leaves alpha?


BenjaminDEMAILLE:
Asking as a downstream packager: PLINK 2.0 was recently added to Homebrew core as `plink-ng`, and pre-release versions get extra scrutiny there. This is also relevant to anyone pinning versions in a pipeline.

In #94 (2018) you mentioned:

> I was planning to start using real version numbers at the beginning of beta testing.  (About 85% of the way there; still need merge and multiallelic-dosage functions before the program skeleton is complete.)

Real version numbers have since arrived (`v2.0.0-a.7.x`), and the README now says "most basic features other than non-concatenating merge are now in place".

A few questions, no dates expected:

1. Is non-concatenating merge still the main blocker for beta, or has the list grown since #94?
2. Is multiallelic dosage handling from that list still outstanding?
3. Is beta a milestone you still intend to reach, or is the alpha label effectively a steady state at this point?
4. Once 2.0 leaves alpha, do you expect the 1.9 line to keep receiving patch releases in parallel?

Mostly trying to understand whether packagers should treat `a.7.x` as the de facto stable artifact for the foreseeable future.

chrchang:
1. Yes, it is still the main blocker.
2. Yes, multiallelic dosage handling is still outstanding, but support is likely to be mostly absent in the first beta build.
3. It is a milestone I intend to reach, but I suppose it is clear at this point that I am not treating it as urgent; you can expect the alpha state to remain for e.g. another year.
4. I do intend to continue maintaining 1.9 for many years to come.

Yes, `a.7.x` is the stable PLINK 2.0 artifact for now.

BenjaminDEMAILLE:
I can help you with some pr

BenjaminDEMAILLE:
Thanks, that answers the packaging question clearly: we will treat `a.7.x` as the stable artifact and not wait on a beta tag.

Since you said the alpha state is likely to persist for a while, I spent some time on things that make that state easier to depend on rather than on the beta blockers themselves. Three PRs, each verified on macOS/arm64 with `2.0/Tests/run_tests.sh` passing 8/8:

[...]

PR #361, opened 2026-09-01
Implement single-pass non-concatenating --pmerge[-list]

BenjaminDEMAILLE:
Closes the gap flagged in #338: `PmergePass()` was a stub, so `--pmerge`/`--pmerge-list` handled only concatenation, and any job where two filesets have something to say about the same variant was rejected. This implements the single-pass case.

**The merge core needed no changes.** `MergePvariant()` and `MergePgenVariantNoTmpLocked()` already took multiple records per variant from multiple filesets, and `InitPvariantPosMergeContext()` already took the filename and line-offset arrays. What was missing was the driver: open every fileset at once, tag each `.pvar` record with its fileset index, sort globally by (chromosome, position, ID, fileset), group, and hand each group to the existing pair.

**The one design decision worth your eye.** `SpgwInitPhase1()` needs the exact output variant count, and in the general case that isn't the sum of `write_nondoomed_variant_ct`: shared variants collapse, and `MergePvariant()` can drop one on allele count after seeing the union. So the `.pvar` and `.pgen` merges run as two loops over the same sorted array, and the allele remaps from the first are kept rather than re-scraped as the `PvariantMergeContext` comment sketches. Keeping them costs `read_max_allele_ct` bytes per record, negligible next to the records themselves, and it avoids a second implementation of the mapping. If you would rather have the re-scrape, that is a contained change to the second loop.

Multipass jobs and `--variant-inner-join` still error out; the multipass scheduling in `PmergePass()` is untouched, so `next_filesets` stays `nullptr` and nothing guesses at the temp-file counterpart.

**One latent bug fixed along the way.** The phase/dosage presence scan in `MergePgenVariantNoTmpLocked()` read `same_id_records[0]`'s variant index out of every *other* record's `.pgen`. Unreachable while the general path was disabled; with the path live it segfaults as soon as a shared variant sits at different indices in two filesets. There is a regression test for exactly that shape.

## Verification

The oracle is reassembly: split a fileset into overlapping slices, merge them, and the result must reproduce the original. That makes the check exact and independent of 1.9's allele-order conventions.

| case | result |
|---|---|
| self-merge (idempotence) | `.bed`/`.bim`/`.fam` byte-identical |
| split by sample, disjoint | 0 differences over `--pgen-diff`, `.pvar` byte-identical |
| split by sample, overlapping block | 0 differences, `.pvar` byte-identical |
| split by variant, overlapping block (the classic `--bmerge` job) | 0 differences, `.pvar` byte-identical |
| three-way, split both ways at once | 0 differences, `.pvar` byte-identical |
| dosage + phase, sample split | exported VCFs identical |
| dosage + phase, staggered variant indices | exported VCFs identical (crashed before the fix) |

PLINK 1.9's `--bmerge` agrees on the sample-split case. Concatenation output is unchanged byte-for-byte against a `master` build, and the full `Tests/run_tests.sh` suite passes.

## Note on overlap with #360

This branch carries the `TEST_PMERGE` harness from #360 (that PR added the concatenation cases; the second commit here extends it to the general path), so it supersedes #360 as it stands. Happy to close that one, or to merge it first and rebase this on top, whichever you prefer.


chrchang:
The main concern here is memory use on large datasets.  Some users working with biobank-scale data are running into the 255-allele limit; some users work with datasets with hundreds of millions of variants.  I want to ensure that, as long as the user is working on a machine large enough to perform other typical plink2 operations on their dataset, the machine can also expect to support --pmerge.  Speed is of course also desirable, but being able to merge at all is the higher priority.  So I'd prefer to save the `read_max_allele_ct` bytes per record.

In the meantime, several more types of tests are needed:
- Handling of genotype hardcall, dosage, and phase merge conflicts, sometimes within multiallelic variants (though we can ignore the multiallelic + dosage combination for now, beyond confirming that it errors out).
- Handling of .pvar QUAL/FILTER/INFO merge conflicts.  INFO in particular is quite complicated.

Matching PLINK 1.9 --bmerge isn't the main goal (though we do want to ensure that that basic use case does not suffer a significant speed hit); covering all the new cases in a reliable manner is the main challenge.

BenjaminDEMAILLE:
Both points addressed; the test work turned up more than I expected.

### Memory

The per-record allele-remap storage is gone, and so is the two-phase write that needed it. The records are already sorted by (chromosome, bp, ID), so each output variant is a contiguous run and the exact output variant count can simply be counted before anything is merged. The .pvar and .pgen are then written together, with each variant's remap consumed while it is still live in `ppmc.pmc`, exactly as `PmergeConcat()` does.

Per input record, that removes `read_max_allele_ct` bytes of remap (up to 255) plus 12 bytes of group bookkeeping. `write_allele_idx_offsets` and `write_nonref_flags` are also now sized by the output variant count rather than the input record count. Nothing is retained per record beyond the record array itself.

`--merge-max-alleles` is the one thing that breaks the count-ahead identity, since `MergePvariant()` only discovers a doomed variant after taking the union of its alleles. It now errors out explicitly rather than making every other job pay for it.

### Tests

Added: hardcall/dosage/phase conflicts, multiallelic hardcall conflicts, and QUAL/FILTER/INFO conflicts. The assertions use `--merge-mode first`, which never consults a later record, so merging (a, b) must reproduce a byte for byte and (b, a) must reproduce b, independent of how conflicts resolve. QUAL/FILTER/INFO are asserted under both `first` and `nm-first`, which differ on a variant where one fileset has no QUAL or FILTER but does have an INFO value.

Deliberately **not** asserted yet: the `nm-match` and `nm-first` genotype-conflict cases, and INFO conflicts under `nm-match`. Writing them surfaced four defects, all reachable on master today through same-position same-ID duplicates and so independent of this PR:

- #362: `nm-match` lets a missing value overwrite an earlier nonmissing one, contradicting both the documentation and 1.9.
- #363: after a clobber, hardcall and dosage disagree; under `nm-first` a dosage is silently dropped from a sample with no conflict at all.
- #364: an INFO conflict under `nm-match`, the default, writes a corrupt .pvar padded with NUL bytes. Two variants gave 2.2 MB; a three-variant job with QUAL and FILTER conflicts too gave 68 GB.
- #365: under `nm-first`, clobbering one sample collapses other samples' multiallelic genotypes to ALT1, including samples both records agree on.

Baking current behaviour into assertions would freeze those in, so those cases wait until the behaviour is settled; I will add them against the fixed versions.

On the 1.9 comparison: the existing sample-split and variant-split cases already check the `--bmerge` use case exactly (the merged .bed is byte-identical to the unsplit original), so that path stays covered as the new cases go in.

Multiallelic-plus-dosage errors out at import (`--vcf multiallelic dosage import is under development`), so there is no way to construct a fileset to merge; nothing to assert at the merge level beyond that.

BenjaminDEMAILLE:
Rebased onto master (through e78ce815), which now carries the concatenation tests from #360; TEST_PMERGE keeps this branch's version, since it is a superset.

chrchang:
Intended merge behavior should be settled by d0b2879 ; let me know if anything is still ambiguous.

BenjaminDEMAILLE:
Verified d0b2879 against all four reproducers, then wrote the assertions this PR was holding back. Rebased onto master and pushed.

**#362, #364, #365 are fixed** and closed. **#363 is fixed too**, and one half of what I reported there was my own mistake: `--vcf ... dosage=DS` imports `1/1:1.8` as a missing hardcall with a dosage of 1.8, since 1.8 is further than the hard-call threshold from 2, so the second record's `s2` really was `./.:1.8` before any merging and the merge was reproducing it correctly. The real half, a dosage dropped from a sample identical in both records, is gone.

### What the tests now assert

Rather than a transcript of today's output, `check_merge.py` states the rules and applies them to whatever the fixtures import to:

- a sample identical in both inputs comes out unchanged, in every mode, whatever its neighbours do (this is what #363 and #365 broke)
- where only one input has a value, that value wins (#362)
- `nm-match` drops a genuine conflict, `nm-first` takes the first nonmissing value
- a sample's value is its whole call, hardcall and dosage together, since plink treats them as a unit

That runs over the biallelic and multiallelic fixtures in both modes. The INFO case is asserted per key, with the .pvar required to be the size of its contents (#364).

### Two order dependences left out, and reported

`nm-match` is symmetric by definition, so the test also merges each fixture both ways and compares. The biallelic fixture passes and that assertion is in. Two do not, and writing the assertion is what found them:

- **#472**: a multiallelic genotype is dropped when the missing record comes second. `./.` against `2/2` gives `2/2` one way and `./.` the other; the same shape with two alleles gives `1/1` both ways, so d0b2879's fix did not reach the patch track.
- **#473**: a dosage present in only one input survives only when that input comes first. `--export A` confirms it is the stored dosage, not an export artifact.

Both are one-sample, one-variant reproducers on current master through the concatenation path, so neither depends on this PR. I will add the matching symmetry assertions once they are settled. #473 also asks a semantics question at the end that I would rather have your answer on before asserting anything about it.

BenjaminDEMAILLE:
Rebased onto master, and the assertions this PR was holding back are now in.

**#472 and #473 are both closed**, so the three order-dependence cases I left out are asserted:

* the biallelic conflict fixture now compares the whole call rather than only the hardcalls (#473: a dosage present in only one input surviving only when that input came first);
* the multiallelic fixture swaps as well (#472);
* the sample-split fixture from 6f swaps too, which is the case where the merge reconciles overlapping records instead of concatenating them.

`--indiv-sort` defaults to `natural` when merging, so both directions write the same sample order and the exported bodies compare directly; the test asserts that order rather than assuming it.

**One real bug the rebase surfaced.** `MergeWriterStruct` gained a `vrtype_mask` field that `MergePgenVariantNoTmpLocked()` reads for every record. `PmergeConcat()` sets it; `PmergePassSingle()` never did, so a non-concatenating merge read it uninitialized and silently dropped both phase and dosage: a sample-split fileset with phased dosages came out as bare hardcalls, with `--pgen-info` reporting "No dosages present". Fixed in its own commit, which also makes the single-pass path honour `--merge-ignore-phase` and `--merge-ignore-dosage` the way the concatenating one does. `TEST_PMERGE`'s 6f caught it.

Full 2.0 suite passes on a clean build.

Issue #366, opened 2026-09-02
MTAG (multi-trait analysis of GWAS): where, if anywhere, it belongs


BenjaminDEMAILLE:
Someone asked me to implement MTAG (Turley et al., *Nat Genet* 50:229-237, 2018) in PLINK 2. Having worked out what it would take, I think the answer is "not here", but the exercise pointed at something that does belong here, so I am writing both halves up rather than quietly dropping it.

## What MTAG is

Given GWAS summary statistics for several genetically correlated traits, MTAG re-estimates each trait's per-SNP effect by borrowing information from the others. It is a generalized inverse-variance-weighted meta-analysis across *distinct* traits rather than across studies of one trait. Per SNP it is a GLS:

```
beta_mtag[t,j]  proportional to  (w_t / w_tt)' (Omega - w_t w_t'/w_tt + Sigma_j)^-1 beta_hat[j]
```

- `Omega` is the variance-covariance matrix of true SNP effects across traits, **assumed identical for every SNP**. Estimated by method of moments from `E(beta_hat_j beta_hat_j' - Omega - Sigma_j) = 0`.
- `Sigma_j` is the variance-covariance matrix of GWAS estimation error for SNP j, which is where sample overlap between the studies enters.

## Why I don't think it belongs in PLINK 2

**It needs LD Score regression.** `Sigma` is estimated by bivariate LDSC: the diagonal from univariate LDSC intercepts, the off-diagonals from bivariate ones. PLINK 2 has neither LDSC nor LD scores. So "implement MTAG" really means "implement LDSC", and that is much larger than the MTAG estimator itself, which is a few hundred lines.

**It is a method, not a utility.** The summary-statistics surface here is utilities: `--clump`, `--adjust-file`, `--gwas-ssf`, `--score`, `--pheno-svd`. MTAG is different in kind. Its homogeneity assumption is violable in practice, violating it inflates the false discovery rate, and the paper ships the `maxFDR` diagnostic precisely to bound that. A `--mtag` without maxFDR would be actively unsafe to give people; a `--mtag` with it is a multi-trait analysis module, which PLINK 2 does not have.

That is the same reasoning you gave on #351, and I think it applies unchanged here.

## What does belong here: LD scores

[...]

PR #367, opened 2026-09-02
Add --ld-score (LD Score regression input)

BenjaminDEMAILLE:
Sending this off the back of #366; close it if you would rather PLINK 2 stayed out of the LDSC input path, no hard feelings.

Each variant's LD Score is the sum of unbiased r^2 with every variant in a window around it, itself included. The unbiased estimator is `r^2 - (1 - r^2)/(n - 2)`, with n the number of samples the pair actually has in common; that correction is what makes an unlinked pair contribute zero on average rather than 1/n. Default window is a 1 cM radius, matching `ldsc --ld-wind-cm 1`, with `--ld-score-window` and `--ld-score-window-kb` available for the other two conventions.

```
--ld-score ['zs'] [cols=<column set descriptor>]
--ld-score-window <max variant ct radius>
--ld-score-window-kb <max kb radius>
--ld-score-window-cm <max centimorgan radius>
```

Column sets are `chrom`, `pos`, `nobs`, `l2`, default `chrom,pos,l2`; ID is always present.

## Implementation

No new arithmetic: it reuses `SplitHomRef2het()`, `FillVaggs()` and `ComputeIndepPairwiseR2Components()`, so the AVX2/SSE/NEON paths and the exact integer accumulators come along unchanged.

The window is walked once per index variant rather than once per pair. That evaluates each pair twice, but it leaves the output partitioned with no shared accumulators, so threads can take disjoint index ranges without any reduction step. Genotypes are read once, by the main thread, into a ring buffer sized to one block of index variants plus the widest window, so memory is bounded by the window rather than by the chromosome.

Multiallelic variants and haploid chromosomes are skipped: `PgrGet()` collapses a multiallelic variant to REF vs. everything else, which is a different statistic, and chrX would need the male/female split `--r2-unphased` performs. Monomorphic variants are reported as NA. All three are counted in the log.

## Verification

Against an independent NumPy implementation (pairwise-complete r^2 from `--export A`, same correction, same window), on real 1000 Genomes genotypes with real LD (L2 from 0.77 to 6.14, 2496 samples, 600 variants):

- NOBS exact on all 597 defined variants
- all 3 monomorphic variants agree as NA
- L2 max relative difference 4.6e-6, which is the precision `dtoa_g` prints

Same check on a simulated no-LD fileset (400 variants, 2000 samples) gives max relative difference 5.0e-6 with NOBS exact.

Output is byte-identical across `--threads 1/2/4/5/16`. 20000 variants by 600 samples with a 1 Mb radius: 7.32s single-threaded, 1.51s on 16 threads.

## Two open questions from #366

Both are conventions I picked rather than derived, so they are easy to change:

1. **Founders only**, following `--r2-unphased` and `--indep-pairwise`. `ldsc` uses every sample in the reference panel. If you would rather match `ldsc` here, or make it a modifier, say which.
2. **No `.M`/`.M_5_50` output.** `ldsc` writes those variant counts alongside the scores and its regression reads them. They are trivial to add, but they are also trivially recomputable from `--freq`, so I left them out rather than guess.

I also dropped a MAF column I had initially: `.l2.ldscore` has no such column, `--freq` already provides it, and it was the only thing making this command need allele frequencies.


chrchang:
Agree that implementing --ld-score makes sense for the reasons you described; will take a closer look at this PR tomorrow morning.

BenjaminDEMAILLE:
Ahead of your pass, the scaling numbers, since this is the one command in the batch where the LD work genuinely dominates. 50k variants x 5000 samples, 1000 kb window radius:

| threads | 1 | 2 | 4 | 8 | 16 |
|---|---|---|---|---|---|
| wall | 22.6s | 12.3s | 8.0s | 5.9s | 4.6s |

Output is byte-identical across all five thread counts. The main thread reads and the workers take disjoint index ranges off a shared ring buffer, so there is no accumulation order to vary.

chrchang:
Regarding founders-only vs. using all samples, I think it is appropriate to force the user to be explicit when nonfounders are present; a --ld-score-founders flag can parallel PLINK 2.0's existing --ac-founders flag (https://www.cog-genomics.org/plink/2.0/filter#nonfounders ).

BenjaminDEMAILLE:
Done, and pushed.

`--ld-score-founders` parallels `--ac-founders`: with nonfounders present the bare command now errors, and you say which you meant. With no nonfounders in the dataset neither flag is needed, so the common case is unaffected.

```
--ld-score                          Error: ... with neither --ld-score-founders
                                    nor --nonfounders; and nonfounders are present.
--ld-score --ld-score-founders      runs on the 100 founders
--ld-score --nonfounders            runs on all 500 samples
```

I checked that the two produce different LD Scores rather than just different messages, which is the part that would have been easy to get wrong.

Writing that test turned up something worth fixing separately. The optional count column was named `nobs`, which everywhere else in plink2 means a sample or genotype count; here it is the number of variants in the window. Both of the runs above reported `NOBS 101` and neither was wrong, which is precisely the problem: the number is right, the name invites you to read it as a sample count and conclude the flag did nothing. Renamed to `nobsi`, following the `--mendel` precedent for a variant count. Value and meaning unchanged.

The multiallelic change from #372 is on this branch too, and the biallelic output is byte-identical to before across 50/100/250 kb windows.

chrchang:
To clarify my comment about --ld-score and multiallelic variants: a primary consumer of this command's outputs should be programs consuming `ldsc` output, and those programs expect multiallelic variants to be filtered out today, so I think --ld-score does have to default to filtering them out today as well.  But it is reasonable to provide a flag/modifier to disable that filtering.

BenjaminDEMAILLE:
That settles it, and it is the answer I asked for. Reverted and pushed.

Filtering multiallelic variants is the default again, with `multiallelic` as the opt-in to keeping them:

```
--ld-score                 112 of 120 variants, 8 multiallelic skipped
--ld-score multiallelic    120 of 120
```

Biallelic output is byte-identical across 50/100/250 kb windows either way.

The distinction I had missed is that this command's output is an interchange format, not a report: what matters is what the downstream consumer expects, not whether the statistic is defensible on its own terms. `--show-tags`, `--blocks` and `--test-mishap` have no such consumer, so they keep the new default from #372.

One thing this leaves slightly awkward: the three other commands now differ from `--ld-score` in their multiallelic default, which is defensible per command but is a wrinkle for anyone reading them as a family. If you would rather they all default the same way and take a modifier to differ, say so and I will make them consistent; I think per-command is right here but I would not defend it hard.

PR #368, opened 2026-09-02
Add --tdt (transmission disequilibrium test)

BenjaminDEMAILLE:
For each variant, counts the alleles heterozygous parents transmit to their affected children; under the null the two are transmitted equally often.

```
--tdt ['zs'] [{'exact' | 'exact-midp'}] [cols=<column set descriptor>]
```

Column sets: `chrom`, `pos`, `ref`, `alt`, `maybeprovref`, `provref`, `a1`, `t`, `u`, `nobs`, `or`, `chisq`, `p`; ID always present. T and U count the ALT allele, named in the A1 column. CHISQ is dropped when an exact test is requested, matching what 1.9 writes.

This reuses `GetTriosAndFamilies()`, so the pedigree handling is exactly the one `--mendel` uses. Duos are excluded: with one parent unobserved, what the other transmitted is not identifiable.

## Two things worth calling out

**No Mendel pre-pass.** PLINK 1.x runs a Mendel error check first and treats offending genotypes as missing. That turns out to be unnecessary here: a Mendel-inconsistent trio leaves the transmitted allele undetermined, so the transmission table has nowhere to put it and the trio is skipped at that variant, which is the same outcome. Rather than argue it, I tested it (below).

**The exact test's symmetric tie.** The two-sided exact binomial is twice the lower tail while the two tails straddle distinct points, but at `T == U` they meet at a single point, so doubling a half-weighted central term restores it to full weight and yields 1 instead of `1 - P(X = n/2)/2`. My first version had exactly that bug; 1.9 disagreed by up to 2.6e-2 on precisely the `T == U` variants, which is what surfaced it. Handled as a separate case now.

## Verification

Against PLINK 1.9 `--tdt` on 400 simulated trios by 300 variants, with real transmission and 20 variants under a known transmission bias:

| mode | T/U mismatches | max rel diff P | max rel diff CHISQ |
| --- | --- | --- | --- |
| chi-square | 0 / 300 | 3.7e-4 | 4.4e-4 |
| `exact` | 0 / 300 | 3.9e-4 | n/a |
| `exact-midp` | 0 / 300 | 4.1e-4 | n/a |

All under the 5e-4 floor 1.9's four-significant-digit output imposes.

Repeated on the same data with 3% of calls set missing and 857 injected Mendel errors (1.9 counts 857 in `--mendel`): T and U still match exactly in all three modes, max rel diff P 4.9e-4. That is the check that pins the inline Mendel handling to 1.9's pre-pass.

## Not implemented

`perm`/`mperm=`, `set-test`, `poo` and the `parentdt1`/`parentdt2`/`pat`/`mat` statistic selectors all error out explicitly rather than being silently ignored. chrX is skipped along with the haploid chromosomes, since it needs the male-hemizygous handling `--glm` performs; the count is logged.

Happy to fold in chrX or the parent-of-origin analysis if you want them in the same PR rather than a follow-up.


BenjaminDEMAILLE:
Flagging a tension in my own reasoning before you spend review time on this.

In #344 I re-bucketed the modules against your criteria and put family-based association, `--tdt`/`--dfam`/`--qfam`/`--cmh`, in the "obsolete by design" bucket, on the grounds that those tests have no way to adjust for covariates. I then wrote this anyway, which is inconsistent of me.

The case I would make for TDT specifically, and it may not be enough:

- It is not an association test that covariate adjustment would improve. It is a within-family transmission test, and its appeal is exactly that it is immune to population stratification by construction rather than by adjustment. That is a different argument from the one that retires `--model` or `--fisher` in favour of `--glm`.
- `--glm` has no family-based mode, so there is no plink2 path to the same inference.
- The pedigree machinery is already there for `--mendel`, so the marginal surface is one command rather than a module.

Against it: the counter-argument you made about IBD applies here too, in that the field largely moved to family-based mixed models and TDT survives mostly in older study designs.

Your call. If it lands in the same bucket as `--fisher`, say so and I will close this; the implementation is verified either way and costs nothing to leave unmerged. I would rather ask now than have you review 550 lines you did not want.

chrchang:
The main thing that would sway me in favor of keeping TDT is recent use of it by teams that have used modern mixed-model analysis tools in other contexts.

BenjaminDEMAILLE:
That is the right test, and I should be straight that I cannot pass it from where I sit: I have no way to survey what analysis teams are currently running, and I am not going to dress up an impression as evidence.

What I can offer is the structural argument, for whatever it is worth against actual usage data.

TDT's niche is not that it is a better test than a mixed model. It is that it is the one family-based test whose null is immune to population structure by construction rather than by adjustment, because the comparison is within a transmission from a heterozygous parent. That is why it survives in trio designs after the same groups have moved to REGENIE or SAIGE for their unrelated-sample work: it is answering a question about a design the mixed model does not cover, not competing on the same question.

If that is right, the usage you would be looking for is trio and family cohorts, not general biobank analysis, and it would show up as a small steady tail rather than a trend. If it is wrong, and TDT has genuinely been displaced in family designs too, then this PR should be closed and I would not argue.

Either way the code is unaffected by the answer, so there is no work pending on my side. Happy to leave it open while you form a view, or close it now if you would rather not carry an undecided PR; the branch does not go stale in any way that matters.

One thing that is worth keeping regardless of the decision: while porting the exact test I found that 1.9's symmetric-tie case is handled by a special path, and reproducing it needed `log1p(-0.5 * exp(ln_central))` rather than the obvious subtraction, which loses all precision when the central term is near 1. That is recorded in the commit message if TDT ever does come back.

chrchang:
I will postpone a decision on this while we work through the other drafted comments.

If we do return to this, I will provide an updated implementation of the exact test based on https://github.com/chrchang/stats .

BenjaminDEMAILLE:
Understood, and no action pending on my side. Nothing here blocks the other drafts.

If you do come back to it, the swap is contained: the exact test is a single function, `TdtExactLnP(tt, uu, midp)`, with one call site. Dropping in the `chrchang/stats` version means replacing that body and nothing else, so it does not need to wait on any decision about the rest of the command.

The one detail I would not want lost in a rewrite: the symmetric case `tt == uu` needs `log1p(-0.5 * exp(ln_central))` rather than the obvious subtraction, which loses its significant digits when the central term approaches 1. That is a property of the binomial tail at the midpoint, not of my implementation, so it should apply to any replacement.

The branch is stable and rebases cleanly; I will leave it as is until you say otherwise.

BenjaminDEMAILLE:
Rebased onto master (through e78ce815); --tdt takes command bit 38, since --ld-score claimed 33 upstream.  Also dropped two parameters TdtReport() never read, which were producing -Wunused-parameter warnings.

PR #369, opened 2026-09-02
Implement the remaining --export text formats

BenjaminDEMAILLE:
You asked, in effect, why the audit's "text output formats" row mattered. This implements all of it, so the row is now empty except `23`, which is #349.

Thirteen format names that the `--export` parser accepted and then rejected after the whole dataset had been loaded now produce output:

`bimbam`, `bimbam-1chr`, `lgen`, `lgen-ref`, `list`, `rlist`, `beagle`, `beagle-nomap`, `structure`, `fastphase`, `fastphase-1chr`, `HV`, `HV-1chr`.

I sent this initially as bimbam alone, with a note that the other formats mostly serve retired tools and were not worth writing. I was asked to do them anyway, so here they are; the judgement about which ones anyone still needs is yours, and dropping commits from this branch is easy if you only want some.

## Shape

Most are variant-major and follow `ExportTped()`. `structure` and `fastphase` are sample-major, so they transpose a block of samples per pass over the variants rather than holding the whole matrix; with enough memory that is a single pass. `HV` is a per-chromosome `.ped`, which `ExportPed()` already writes, plus a two-column `.info`.

## Verification

Against PLINK 1.9 `--recode <fmt>`, on 1200 samples by 300 variants across two chromosomes, with missing calls and missing phenotypes present. Every field of every line matches:

| format | files compared |
| --- | --- |
| `bimbam`, `bimbam-1chr` | `.geno.txt`, `.pheno.txt`, `.pos.txt`, byte-identical |
| `lgen` | `.lgen` (360000 lines), `.fam`, `.map` |
| `lgen-ref` | `.lgen` (189035 lines, reference genotype omitted), `.ref`, `.fam`, `.map` |
| `list` | `.list` (1200 lines) |
| `rlist` | `.rlist` (900 lines), `.fam`, `.map` |
| `beagle` | `.chr-1/.chr-2` `.dat` and `.map` |
| `beagle-nomap` | `.dat` only, no `.map` written |
| `structure` | `.strct_in` (1202 lines) |
| `fastphase`, `fastphase-1chr` | `.phase.inp` on both chromosomes (3603 lines each) |
| `HV`, `HV-1chr` | `.ped` and `.info` on both chromosomes |

`bimbam-1chr`, `fastphase-1chr` and `HV-1chr` all reject multi-chromosome input, as 1.9 does.

## Three deliberate divergences

1. **`.lgen` delimiters.** 1.9 writes `FID<tab>IID<space><id><space><space><a1><space><a2>`. plink2 uses `exportf_delim` uniformly, so the fields match but the separators do not. Reproducing 1.9's mixed delimiters did not seem worth it.
2. **`HV` missing phenotype.** 1.9 writes `0` there while its own `--recode` writes `-9`. plink2 uses the requested `--output-missing-phenotype` consistently, and matches 1.9's HV output byte-for-byte when passed the same value.
3. **`bimbam` `.pheno.txt`** gets one column per loaded phenotype instead of 1.9's single column, which is what GEMMA's `-n` selects among and reduces to 1.9's output when one phenotype is loaded. Also, plink2 drops the `.recode` infix 1.9 puts in bimbam and fastphase filenames, as it already does elsewhere.

## Also in here

One commit unrelated to any format: unimplemented `--export` format names were only rejected inside `Exportf()`, after the entire dataset had been loaded and filtered. They are now rejected during argument parsing, naming the offending format. With this branch only `23` is left to reject.


BenjaminDEMAILLE:
Second commit added, on the defect I mentioned in #344 rather than another format.

All fourteen accepted-but-unwritable `--export` format names were only rejected inside `Exportf()`, after the entire dataset had been loaded and filtered. `--export structure` on a biobank-scale fileset reads and filters everything before saying it cannot do it. They are now rejected during argument parsing:

```
$ plink2 --bfile big --export structure --out X
Error: "--export structure" is not implemented yet.
```

with no read of the dataset at all. `kfExportfImplemented` sits next to `kfExportfTypemask` so the two are updated together, and the message names the offending format rather than listing every implemented one. The existing check inside `Exportf()` stays as a safety net.

Verified: all eight remaining unimplemented format names now fail before any variant is loaded, and `bimbam`, `ped`, `vcf` and `A` still export normally. Suite still passes.

Happy to split this into its own PR if you would rather keep the bimbam change isolated; I put it here because it edits the same mask.

On the rest of the format module, I posted a per-format breakdown on #344. Short version: of the eight text output formats I had listed as still routine, only `bimbam`/`bimbam-1chr` look worth writing, so I am not planning to send the other six. `structure`, `HV` and `list`/`rlist` are also sample-major, so they would need the load-and-transpose machinery `ExportPed()` carries rather than the variant-major loop this reuses from `ExportTped()`, which is a poor trade for formats whose consumers have moved on.

chrchang:
These were low-priority, but I've always intended to implement the full list eventually rather than dropping them.  I will start reviewing this in detail tomorrow.

BenjaminDEMAILLE:
Before you start the detailed pass, I went looking for problems in this PR myself and found two worth pushing first.

**A real bug: `.rlist` class ordering.** PLINK 1.x emits a variant's lines as HET, then the minor homozygote, then the missing class. I was writing HOM before HET. Same variants and same contents either way, so it survives a casual look, but it is visible to anything reading the file positionally or diffing against 1.9. Confirmed on 36 two-class variants in one dataset and all 300 in another; both now match field-for-field. `.list`, which uses a separate order array, was already correct.

**Performance.** `.lgen`, `.list` and `.rlist` emit one field per (sample, genotype), and I had per-sample work sitting in that innermost loop: a `strlen` per line and, when the delimiter is not a tab, a `memchr` per line to swap the FID/IID separator. On 50k variants x 5000 samples that is 250 million of each in `.lgen` alone. Hoisted into a single pass.

```
--export lgen   4.70s -> 2.89s user CPU
--export list   3.11s -> 2.20s
--export rlist  2.32s -> 1.61s
```

Byte-identical before and after on all three, apart from the ordering fix.

On threading, since it came up on #374: I do not think it is the right lever for these. Past this point they are write-bound, not CPU-bound. `.lgen` alone produces 5.6 GB on that dataset, so parallelising the formatting would run into the write almost immediately. The `bimbam`, `structure` and `fastphase` paths are all under 2s on the same data. If you want these threaded anyway for consistency with `ExportPed`, say so and I will, but I would rather not add thread machinery that buys nothing.

All eight formats still match PLINK 1.9 field-for-field, with the `.lgen` delimiter difference already noted in the PR description. Full suite passes.

chrchang:
The BIMBAM format is dosage-based, and the BEAGLE format is based on genotype likelihoods.  Supporting proper conversion of those is the main way in which PLINK 2.0 can provide significant value over 1.9 here.  --gen, --import-dosage, and --export Av/oxford have the necessary logic.

BenjaminDEMAILLE:
Right on both counts; I had been treating these as text-format ports rather than as conversions. BIMBAM is done and pushed. BEAGLE I want to check with you before writing.

**BIMBAM.** With dosages present the output is now BIMBAM's mean genotype file, which is what GEMMA actually reads: no header, each row being variant ID, the two alleles, then ALT dosage on a 0..2 scale.

```
v0,C,A,0,0.323,0,0.019,0.102,0,0.096,0,1.34,0.256,...
```

Without dosages the output is unchanged and still byte-identical to 1.9's genotype file.

Those are structurally different files rather than variants of one, so the switch is announced in the log rather than made silently. If you would rather it be an explicit format keyword than data-driven, that is a small change and I would probably agree with you; a pipeline that starts getting a different file shape because someone re-imputed upstream is not a nice surprise.

Worth recording that my first attempt wrote a hybrid: the genotype file's `nsample / nvariant / IND,...` header over mean-genotype numeric rows, with the two allele columns missing. Valid as neither format, and it looked plausible enough in a `head` that I nearly kept it.

**BEAGLE.** Here I would rather ask than guess. 1.9's `--recode beagle` writes the marker/genotype pair of files with hardcall genotypes, and the format's own likelihood variant is a different file layout, not a column added to the same one. So "proper conversion" could mean either:

- emit the BEAGLE genotype-likelihood file when the source has dosages, deriving GL the way `--export vcf vcf-dosage=GP` derives GP, mirroring what I did for BIMBAM; or
- leave `beagle` alone and treat the likelihood file as its own output.

I lean towards the first for symmetry with BIMBAM, but BEAGLE 4+ reads VCF and the older formats are mostly of archival interest, so I am not confident the likelihood path has real users to serve. If you think it does not, saying so is a perfectly good answer and I will leave `beagle` as the 1.9-compatible hardcall export.

chrchang:
Okay, I spoke too soon re: BIMBAM and BEAGLE, initial Google summary I read did not call out the important detail that both programs defined several different formats for different data subtypes.  It is probably still best to support more than biallelic unphased hardcall import/export for both format families, but I don’t want to go too deep down that rabbit hole.

BenjaminDEMAILLE:
Understood, and that matches where I landed, so nothing further from me unless you want it.

Where this PR now stands against "more than biallelic unphased hardcall, without going deep":

- **BIMBAM**: done. Dosages present gives the mean genotype file GEMMA reads; hardcalls give the genotype file, still byte-identical to 1.9. That covers the dosage subtype, which is the one that carries information the hardcall path would destroy.
- **BEAGLE**: left as the 1.9-compatible hardcall export. I had offered to derive genotype likelihoods for it; on your answer I will not, since BEAGLE 4+ reads VCF and the likelihood file is a separate layout rather than a richer version of the one being written.
- **Phase**: not handled by either. `--export beagle` on phased data currently writes hardcalls without saying so. If you want one more increment here, a warning when phase or dosage information is being silently dropped on the way out is cheap and would cover the whole legacy-format family rather than one format at a time. Say the word and I will add it; otherwise I will leave the PR as is.

On the data-driven format switch in BIMBAM, my offer stands to make it an explicit keyword instead. I mildly prefer the explicit version for the reason I gave, but it is your call and it is a small diff either way.

chrchang:
Reviewing the confusion that users have had with PLINK 1.9's `--recode bimbam`, and current practice with BIMBAM formats... I think only the Mean Genotype Format should be supported by PLINK 2.0, everything else is more clutter than it's worth.

How about the following design:

`--export mgf` generates a .mgf[.gz] + .pos.txt [+ .pheno.txt] fileset.  (The 'bgz' modifier causes the .mgf file to be block-gzipped; if no numeric phenotype information is loaded, no .pheno.txt is produced; if multiple numeric phenotypes are loaded, the .pheno.txt file has multiple columns.)

Since tab/space delimiters are allowed in .mgf files, we default to tab and support the 'spaces' modifier; no need to support commas on output.  We error out if there's a comma or semicolon in a variant ID or allele code.

`--mgf <full .mgf path> <full .pos.txt path> ['pheno='<full .pheno.txt path>]` imports this format.  Could add `['single-chr='<code>]` later if there is demand.

`--export bimbam` and `--export bimbam-1chr` produce an error message during command-line parsing explaining the new state of affairs, with a note to let us know if they really do need the original functionality.

chrchang:
And how about this design for BEAGLE (targeting v3.3, which didn't support VCF input but did have some functionality absent from later BEAGLE versions):

* `--export beagle-unphased` is equivalent to the old `--recode beagle-nomap` (automatically converted by command-line parser).  `--export beagle-phased` requires all genotypes to be phased, and changes the output file extension to .dat.phased .  `--export beagle` produces an error message directing the user to PLINK 1.9 if they really still need chromosome-split output.
* Multiallelic variants (up to 128 alleles, see https://faculty.washington.edu/browning/beagle/beagle_3.3.2_31Oct11.pdf ) should be allowed.
* Instead of a single phenotype header line of type 'A'/'T', there is a variable (possibly zero) number of phenotype header lines of type 'A'/'T'/'C'.

chrchang:
In the BEAGLE case, since there is now no .map output, I think it's fine to not implement a corresponding import flag.

BenjaminDEMAILLE:
BIMBAM side implemented to your spec and pushed. BEAGLE I have read carefully but not started; scoping note at the end.

**`--export mgf`** writes `.mgf` + `.pos.txt`, plus `.pheno.txt` when at least one quantitative phenotype is loaded, one column per such phenotype. Case/control and categorical phenotypes are left out rather than coerced into a numeric column. Tab-delimited by default, `spaces` switches, `bgz` block-gzips the `.mgf`. No commas on output, and a comma or semicolon in a variant ID or allele code is an error.

Verified:

```
2 quantitative + 1 case/control    .pheno.txt has 2 columns, 2000 rows
no phenotype                       no .pheno.txt written at all
bgz                                real BGZF, decompresses to the plain output byte for byte
planted 'v,comma' variant ID       rejected with a specific message
```

**`--export bimbam` / `bimbam-1chr`** fail at parse time:

```
Error: "--export bimbam" and "--export bimbam-1chr" have been replaced by
"--export mgf", which writes BIMBAM's mean genotype format: dosages rather
than rounded genotypes, in a .mgf file alongside .pos.txt and, when a numeric
phenotype is loaded, .pheno.txt.  Contact us if you need the original BIMBAM
genotype file; PLINK 1.9 still writes it.
```

The 1chr variant is gone rather than kept: `.pos.txt` always carries the chromosome column, which removes the reason it existed.

Worth recording, since it is the kind of thing that survives a quick look: converting the `.mgf` writer to a bgzf stream, my edit ran past the end of that section and left `.pheno.txt` writing through the stream after it had been closed. Segfault, exit 139, with a plausible-looking partial `.mgf` already on disk. `.pheno.txt` and `.pos.txt` stay on `FILE*`.

**Not done yet: `--mgf` import**, and the whole BEAGLE redesign. The BEAGLE piece is bigger than it looks from the spec: variable-count phenotype header lines of type A/T/C, up to 128 alleles per variant, and `beagle-phased` needing a phase check across every variant before it can commit to the output name. I would rather send that as its own commit than bolt it onto this one. Say if you would prefer it in this PR or a separate one.

BenjaminDEMAILLE:
BEAGLE done too, to your spec. Both format families are now in this PR.

```
--export beagle-unphased    single .dat, no chromosome split, no .map
--export beagle-phased      .dat.phased, requires every het call phased
--export beagle             parse-time error pointing at both, and at 1.9
```

Phenotype header lines are now one per loaded phenotype, typed `A` case/control, `T` quantitative, `C` otherwise, and zero when none are loaded. On a set with two quantitative and one case/control phenotype:

```
P FID   ...
I IID   ...
T QT1   ...
T QT2   ...
A CC1   ...
```

Multiallelic variants supported to BEAGLE's 128-allele limit.

**Two bugs, both found by diffing output genotypes against the source VCF rather than reading the file.** Recording them because the first one is the sort that ships:

- My first multiallelic version emitted only REF and ALT1, so a three-allele variant silently dropped its third allele. Every code in the output was a legitimate allele and the file looked entirely plausible; only comparing against the input showed the `G` was never written.
- The phase swap was applied only to genovec value 1, but an ALT1/ALT2 heterozygote is stored as hom-ALT with a patch entry. 91 of 10000 phased genotypes came out with their haplotypes exchanged. The heterozygous-call count driving the phase check had the same blind spot and underflowed, which is what surfaced it.

Verified on 200 samples x 50 phased variants, 4 triallelic: **10000 of 10000 phased genotype pairs match the source VCF exactly, including haplotype order**; the unphased path matches order-insensitively.

I rewrote the function rather than patching it. Three incremental attempts each left it non-compiling, which was a signal about the size of the change rather than about the edits.

Still not done: the `--mgf` import flag. You said BEAGLE needs no import counterpart since there is no .map any more; `--mgf` was in your spec, so say if you want it in this PR or separately and I will do it.

chrchang:
Hmm, there's that flaky CI test again, same failure as last time.  I will try to investigate that after addressing the immediate merge bugs.

`--mgf` can be implemented in a separate PR.

BenjaminDEMAILLE:
Starting on `--mgf` in its own PR as you said. One design question blocks it, and I would rather ask than pick.

**The format carries no sample identifiers.** Neither of the three files has them, which is easy to overlook because the export side never needed to care:

```
.mgf        v0  C  A  0  0.323  0  0.019  ...     variant, alleles, dosages
.pos.txt    v0  1000  1                            variant, bp, chromosome
.pheno.txt  -0.9408  5.4454                        positional, one row per sample
```

Sample identity in BIMBAM is positional across `.mgf` columns and `.pheno.txt` rows, and nothing names them. So `--mgf` can recover genotypes, positions and phenotypes, but not who the samples are. That is a property of the format, not something a better reader gets around.

Three options as I see them, in the order I would rank them:

1. **Synthesize IDs** (`per0`, `per1`, ...), as `--dummy` does, and say so in the log. The resulting fileset is self-consistent and usable for association, but cannot be joined to anything else, and merging two `--mgf` imports would silently pair unrelated samples.
2. **Take an optional `--psam` / `iid-file=`** giving one ID per line in column order, so a user who kept the IDs alongside can restore them. Costs one more parameter and puts the burden on the user to keep the order.
3. **Refuse without an ID source.** Safest against the merge failure mode, most annoying for the common case.

I lean to 1 with a loud log line, plus 2 as an opt-in, and I would avoid 3. But this determines whether the imported data is joinable, so it is your call rather than mine.

Related, and cheaper to answer: should `--mgf` accept comma-delimited input? Our exporter emits tab or space only, but real `.mgf` files in the wild are usually comma-separated, since that is what GEMMA's own documentation shows. I would autodetect on the first line rather than add a flag.

I will hold the implementation until you weigh in; there is no point writing the sample-handling twice.

2.0/bin/plink2
chrchang, line 1: This should not be in the PR.  I'm also guessing that this PR was intended to include fewer new test subdirectories.

BenjaminDEMAILLE:
Force-pushed. Two things:

**Rebased onto master** through c6538797, converting to the self-numbering command-bit enum from #383. All 22 of my open branches are rebased the same way; each one builds warning-free and its test suite passes against the in-repo 1.9.

**Removed an accidental commit payload.** The `Drop the unused ChrInfo parameter from ExportBeagle` commit I pushed earlier today was staged with `git add -A 2.0` and swept in 589 files: a 4 MB `2.0/bin/plink2` plus the test output left in `2.0/Tests/TEST_*/` by suites belonging to other branches (those directories' `.gitignore` files are only present on the branches that add them, so they do not protect a branch that lacks them). The commit now contains only the three source files it describes, 3 insertions and 5 deletions. Sorry for the noise if you fetched it in between.

2.0/bin/plink2
BenjaminDEMAILLE: Both correct, and both fixed — sorry, this was my mistake and I answered it at the PR level rather than here.

The binary and the test subdirectories came from the same error: I staged that commit with `git add -A 2.0`, which swept in 589 files. The test artifacts are only ignored by `.gitignore` files that live inside each `2.0/Tests/TEST_*/` directory, so they are not ignored on a branch that does not itself add that directory, and `2.0/bin/` is not ignored at all. The commit now contains only the three source files it describes, 3 insertions and 5 deletions.

The branch carries one test directory, `TEST_EXPORT_LEGACY`, which is the one this PR is actually for.

chrchang:
Merging now.  Minor changes I'll make afterward:
- plink2_common.h will be updated to mark "--export 23" as implemented; this fixes CI in my testing.
- BEAGLE 3.3 and fastPHASE export will come with a deprecation warning.  Yes, this is a bit weird for the actually-new beagle-phased case, but I'm having trouble finding any examples of people caring about these formats within the last few years; in particular, there seems to be no interest in the BEAGLE 3.3 edge case I called out.
- MGF-format import/export will be moved from plink2_{export,import}_legacy to plink2_{export,import} since they involve non-legacy dosage data.

PR #370, opened 2026-09-02
Add --clump-index-first and --clump-replicate

BenjaminDEMAILLE:
Two of PLINK 1.9's `--clump` options plink2 lacked. Both are filters on which clumps survive, so they are increments on the existing `ClumpReports()` rather than new machinery.

`--clump-index-first` restricts index variants to those reaching `--clump-p1` in the first file. That means knowing which file each significant result came from even when no output column needs it, so it keeps a bitvector of qualifying alleles, allocated alongside `best_ln_pvals` so it survives the compaction step.

`--clump-replicate` keeps a clump only when its below-p2 results come from more than one file. The subtlety is what counts: it has to be exactly the entries the SP2 column would print, which means excluding the index variant's own entry from its own file. It also forces on the per-entry file indices and the clump-entry list that SP2 would otherwise be alone in needing, so it works with `cols=` sets that omit `sp2`.

One bug worth recording, since it produced plausible-looking output rather than a crash: my first version ran the check after the leading columns had been written into the compress-stream buffer, so a rejected clump left a truncated line that spliced onto the next surviving one. The result was the right number of lines with wrong content, which is exactly the failure mode that survives a cursory look. The check now runs before any of the row is written.

## Verification

Against PLINK 1.9 `--clump` on 601 variants of real 1000 Genomes genotypes with two association files, across 36 combinations: `--clump-p2` in {0.1, 0.3, 0.6} by `--clump-r2` in {0.1, 0.5} by with/without `--clump-allow-overlap` by each of `--clump-index-first`, `--clump-replicate`, and both together. The set of surviving index variants matches 1.9 in every combination.

The baseline also matches without either flag, including SP2 contents, which is what makes the comparison meaningful.

## Not in here

The other three 1.9 `--clump` options, `--clump-best`, `--clump-verbose` and `--clump-annotate`, all need per-member r-squared or arbitrary retained input columns. `ClumpEntry` currently carries a p-value bin and a file index and nothing else, and the members are varint-packed with a temp-file spill path, so retaining r-squared per member is a change to memory-sensitive code rather than an increment on it. Given what you said on #361 about memory being the higher priority, I would rather ask before touching that than send it speculatively. Say the word and I will.


BenjaminDEMAILLE:
Following up on my own question with a concrete design, so the answer can be yes or no rather than open-ended.

I read how PLINK 1.9 does it. Annotations are a variable-length trailing array on `Clump_entry`, allocated as `offsetof(Clump_entry, annot) + annot_len - 1` from the end of the bigstack, so they cost nothing unless `--clump-annotate` is given. Per-member r^2 for `--clump-best`/`--clump-verbose` is carried through the same structures the main path uses. The verbose and best code paths run from `plink_ld.c:13281` to `:14101`, interleaved with the `--clump-range` set-definition handling at both ends.

What I would do differently in plink2, to keep `ClumpEntry` and the varint-packed member list untouched:

- **`--clump-annotate`**: same shape as 1.9, a trailing char array on `ClumpEntry` sized from the requested field widths, present only when the flag is given. No cost otherwise.
- **`--clump-best` / `--clump-verbose`**: recompute member r^2 in a second pass rather than storing it. The clumps are fully determined after the first pass, so a second pass can reload each index variant and its members and recompute. That is time for memory, which is the direction you indicated on #361, and it leaves the varint packing and the temp-file spill path exactly as they are.

The cost is that `--clump-verbose` re-reads genotypes for the retained clumps, which for a typical run is a small fraction of the variants. If you would rather have the one-pass version with r^2 stored per member, that is also fine, it is just the version I did not want to send without asking.

Either way it is a large diff in the middle of the code you asked me to shrink, so I will hold until you say which one, or that you would rather leave these three out.

BenjaminDEMAILLE:
Added `Tests/TEST_CLUMP_EXTRAS`, and it caught a bug in --clump-index-first.

**The flag restricted which alleles could be index variants, but not which file's result got reported for them.** The p-value, file index and A1 still came from whichever file had the smallest p-value, so on the test fixture 6 of 14 clumps came out with F=2 where PLINK 1.9 has F=1, and the index variant's own other-file entry then showed up in its SP2 list:

```
1.9:     1  1  b8v6  812001  0.114    11  5  6  0  0  0  ...,b8v6(2),...
plink2:  1  812001  b8v6  2  0.01106  11  6  5  0  0  0  ...,b8v6(1),...
```

`best_ln_pvals[]` and `best_fidx_x2s[]` turn out to be read back only for the index variant's own reported values (secondary variants come from `clump_entries[]`), so holding them to the first file when --clump-index-first is specified is enough. All 14 clumps match 1.9 now.

**What the test checks:** two association reports over one fileset, one on all samples and one on half, compared against 1.9 for plain --clump, --clump-index-first, --clump-replicate and both together (15/14/14/13 clumps). Plus that --clump-index-first leaves no F=2 rows where plain --clump has some, that --clump-replicate can only shrink the report, and that a single input file with --clump-replicate leaves nothing.

PR #371, opened 2026-09-02
Add --twolocus

BenjaminDEMAILLE:
Joint genotype counts for a pair of variants, with the case/control split PLINK 1.x reports when a case/control phenotype is loaded.

```
--twolocus ['zs'] <variant ID> <variant ID>
```

**Format is not 1.x's.** PLINK 1.x writes a fixed-width human-readable report with marginal totals. plink2 has no fixed-width output anywhere else and dropped the width-formatting helpers 1.x used for it (`uint32toa_w10x`, `dtoa_f_w9p6_spaced`, `fw_strcpy`), so reproducing it would mean reintroducing formatting machinery plink2 deliberately does not have, to produce a report style it deliberately does not use. This writes a table instead, one row per (group, genotype, genotype) cell:

```
#GROUP  ID1          GT1  ID2          GT2  CT    FREQ
ALL     rs559462325  G/G  rs181691356  C/C  2491  0.997997
ALL     rs559462325  G/G  rs181691356  A/C  4     0.00160256
```

Every count that produces a marginal is present, so nothing is lost; say the word if you would rather have the marginals as rows.

Two variant IDs is few enough that a linear scan beats standing up a hash table, and it also catches duplicate IDs, which the report cannot resolve.

## Verification

Against PLINK 1.9 `--twolocus`, parsing its fixed-width report and comparing cell by cell: all 48 cells, 16 per group across ALL, CASE and CTRL, match exactly.

That comparison earned its keep: it caught the report emitting two fields where the header declares one, for the missing-genotype rows. The line count was right and the values were right; only the column alignment was wrong, which is the kind of thing that survives reading the output.


BenjaminDEMAILLE:
Rebased onto master (through e78ce815).  --ld-score claimed command bit 33 upstream, so --twolocus moves to bit 39; the branch is a single commit now, since the incremental history no longer applied cleanly across the new plink2_ld.cc additions.  Re-verified against PLINK 1.9 --twolocus after the rebase: the joint genotype counts match cell for cell.

chrchang:
This command should support multiallelic variants, and probably shouldn't have a 'zs' modifier.  Otherwise the idea looks good.

BenjaminDEMAILLE:
Both done, pushed.

**Multiallelic variants.** They were rejected outright; they now get one row per unordered allele pair, sized from the two variants' allele counts rather than a fixed 4x4 table. Genotypes are read with `PgrGetM` and decoded to allele pairs with `PglMultiallelicSparseToDenseMiss`, so ALT1/ALT1 and ALT1/ALT2 land in separate cells instead of being collapsed into one non-REF category. On a triallelic variant against a biallelic one:

```
#GROUP  ID1  GT1  ID2  GT2  CT  FREQ
ALL     v1   A/A  v2   C/C  1   0.166667
ALL     v1   A/G  v2   C/T  1   0.166667
ALL     v1   G/G  v2   C/C  1   0.166667
ALL     v1   A/T  v2   T/T  1   0.166667
ALL     v1   G/T  v2   C/T  1   0.166667
ALL     v1   T/T  v2   .    1   0.166667
```

(zero cells omitted here; the report writes all 28.)

**'zs' removed.** Agreed — the report is one row per joint genotype cell, so it was never large enough to be worth a modifier.

`Tests/TEST_TWOLOCUS` covers the multiallelic case (every genotype present, each sample in exactly one cell, 28 rows for 3x2 alleles) and checks that `zs` is now rejected.

2.0/plink2_ld.cc
```diff
@@ -13423,6 +13423,230 @@ PglErr LdScore(const uintptr_t* orig_variant_include, const ChrInfo* cip, const
...
+          cur_idxs[sample_idx] = missing_idx;
+          continue;
+        }
+        const uint32_t lo = MINV(ac0, ac1);
+        const uint32_t hi = MAXV(ac0, ac1);
+        cur_idxs[sample_idx] = (hi * (hi + 1)) / 2 + lo;
+      }
+    }
+
+    // Case/control split, when a case/control phenotype is available; PLINK
```
chrchang: Since multiple phenotypes may be loaded, the command should take an optional binary/categorical phenotype name, and if no phenotype name is provided I don't think it should report a case/control split even if exactly one case/control phenotype is loaded.

It is ok to implement the phenotype-name parameter in a separate PR.

BenjaminDEMAILLE: Makes sense, and the two halves split cleanly:

- **No implicit case/control split.** I will do this here: without a phenotype name, `--twolocus` reports one table over all samples, even when exactly one case/control phenotype happens to be loaded. That is a behavior change to what is already in this PR, so it belongs in it.
- **`--twolocus-pheno <name>`** (or whatever spelling you prefer) for the binary/categorical case: separate PR, as you suggested.

I will take the same approach for `--flip-scan` once #394 lands, since you asked for the equivalent there.

One question on the categorical case, so the follow-up matches what you have in mind: for a categorical phenotype, do you want the same shape as `--flip-scan` -- a phenotype name plus a set of category names to treat as "cases" -- or one table per category?

chrchang: One table per category here.

BenjaminDEMAILLE: One table per category it is: #395, stacked on this branch.

A binary phenotype is recoded as a two-category one the way `FstReport()` does it, so CASE/CONTROL comes out of the same code path rather than a separate one. Samples with no value for the named phenotype land in ALL only, so the category tables partition it.

The name is a third positional argument, matching how `--cmh` takes its phenotype name. Easy to change to a separate `--twolocus-pheno` flag if you would rather.

(The anchor on this thread is the block #371 already removed -- that PR now reports a single ALL table when no name is given, which was the other half of your comment.)

chrchang:
oops, garbled the merge.  I'll fix this up after landing.

PR #372, opened 2026-09-02
Add --show-tags, --tag-kb, --tag-r2, --list-all

BenjaminDEMAILLE:
Variant A tags variant B when they are within `--tag-kb` of each other and their unphased r-squared is at least `--tag-r2`. Both PLINK 1.x modes:

```
--show-tags ['zs'] {<filename> | 'all'}
--tag-kb <kbs>   (default 250)
--tag-r2 <val>   (default 0.8)
--list-all
```

A filename produces `.tags`, every variant tagging at least one named variant (a superset of the input, since a variant tags itself). `all` produces `.tags.list`, each variant's tags one row per variant. `--list-all` adds that per-variant file to filename mode, restricted to the named variants.

The window-and-r-squared shape is `--ld-score`'s (#367): genotypes held as the hom/ref2het bitvector pair `ComputeIndepPairwiseR2Components()` consumes, in a ring sized to the widest window, so memory is bounded by the window rather than the chromosome. Founders only, multiallelic variants and haploid chromosomes skipped, as in `--r2-unphased`.

## Two bugs the comparison caught

`VariantIdDupHtableFind()` dereferences its `llidx_ptr` output parameter unconditionally on a match. Its comment says llidx is "currently unset" on failure, which reads as though the pointer is optional, so I passed `nullptr`; it segfaults the moment a named variant is found. Not a bug in that function, but worth knowing the parameter is mandatory.

`--list-all` in filename mode reports only the named variants, not every variant. My first version wrote all 600 rows where 1.9 writes 5. Because plink2 sorts flags alphabetically, the "must be used with --show-tags" check also has to happen after the parse loop rather than during it, the same trap `--allele-count` has.

## Verification

Against PLINK 1.9 `--show-tags`, on 601 variants of real 1000 Genomes genotypes:

| mode | combinations | result |
| --- | --- | --- |
| `all` | `--tag-r2` in {0.3, 0.5, 0.8} by `--tag-kb` in {100, 250} | identical, 600 rows each |
| filename | same six | identical |
| filename + `--list-all` | `--tag-r2` in {0.3, 0.8} | both `.tags` and `.tags.list` identical |

## Not in here

`--tag-mode2` is an I/O format variant of filename mode, and `--ld-xchr` is a chrX coding model shared with `--indep`, `--r2` and `--flip-scan`, so it belongs with whichever of those lands rather than here. `--blocks` is the remaining piece of that module and is a much larger job: PLINK 1.9's `haploview_blocks_classify()` is ~250 lines of hand-optimized early exits over a 101-point D-prime likelihood grid, and a transcription slip there produces silently wrong block boundaries rather than a crash. I would rather do it separately and carefully.

### DIFF
--- a/2.0/plink2_help.cc
+++ b/2.0/plink2_help.cc
@@ -1140,6 +1140,24 @@ PglErr DispHelp(const char* const* argvk, uint32_t param_ct) {
 "      all-pairs computation on more than 400k variants.\n"
 "    With either output type, the computation can be subdivided with --parallel.\n\n"
               );
+    HelpPrint("show-tags\0list-all\0tag-kb\0tag-r2\0", &help_ctrl, 1,
+"  --show-tags ['zs'] {<filename> | 'all'}\n"
+"    Report which variants tag which, where variant A tags variant B when the\n"
+"    two are within --tag-kb of each other and their unphased r^2 is at least\n"
+"    --tag-r2.\n"
+"    * With a filename, <output prefix>.tags lists every variant tagging at\n"
+"      least one variant named in the file.  A variant tags itself, so this is\n"
+"      normally a superset of the input list.\n"
+"    * With 'all', <output prefix>.tags.list reports each variant's tags, one\n"
+"      row per variant.  --list-all adds that file to the filename mode,\n"
+"      restricted to the named variants.\n"
+"    * Only founders are considered, as with --r2-unphased.  Multiallelic\n"
+"      variants and haploid chromosomes are skipped.\n\n"
+              );
+    HelpPrint("tag-kb\0tag-r2\0show-tags\0", &help_ctrl, 0,
+"  --tag-kb <kbs>  : Set --show-tags max tag kb distance (default 250).\n"
+"  --tag-r2 <val>  : Set --show-tags min tag r^2 (default 0.8).\n"
+              );
     HelpPrint("ld-score\0ld-score-founders\0ld-score-window\0ld-score-window-kb\0ld-score-window-cm\0", &help_ctrl, 1,
 "  --ld-score ['zs'] ['multiallelic'] [{'cols='<column set descriptor>}]\n"
 "  --ld-score-founders\n"
--- a/2.0/plink2_help.cc
+++ b/2.0/plink2_help.cc
@@ -1152,7 +1152,11 @@ PglErr DispHelp(const char* const* argvk, uint32_t param_ct) {
 "      row per variant.  --list-all adds that file to the filename mode,\n"
 "      restricted to the named variants.\n"
 "    * Only founders are considered, as with --r2-unphased.  Multiallelic\n"
-"      variants and haploid chromosomes are skipped.\n\n"
+"      variants are kept, with the major allele taken against the rest.\n"
+"    * Unique variant IDs are required, since the report and the input list\n"
+"      are both keyed on them.\n"
+"    * Haploid chromosomes are not supported yet, and are an error rather\n"
+"      than a silent omission.\n\n"
               );
     HelpPrint("tag-kb\0tag-r2\0show-tags\0", &help_ctrl, 0,
 "  --tag-kb <kbs>  : Set --show-tags max tag kb distance (default 250).\n"
--- a/2.0/plink2_ld.cc
+++ b/2.0/plink2_ld.cc
@@ -13423,6 +13423,421 @@ PglErr LdScore(const uintptr_t* orig_variant_include, const ChrInfo* cip, const 
   return reterr;
 }
 
+void InitTag(TagInfo* tip) {
+  tip->tag_fname = nullptr;
+  tip->list_all = 0;
+  tip->bp_radius = 250000;
+  tip->output_zst = 0;
+  tip->r2_thresh = 0.8;
+}
+
+void CleanupTag(TagInfo* tip) {
+  free_cond(tip->tag_fname);
+}
+
+// --show-tags: for each variant, the nearby variants it is in strong LD with.
+// Same window-and-r^2 shape as --ld-score: genotypes are held as the
+// hom/ref2het bitvector pair the pairwise kernels consume, in a ring sized to
+// the widest window.
+typedef struct TagCtxStruct {
+  const uint32_t* variant_bps;
+  const uint32_t* chr_variant_uidxs;
+  uint32_t chr_variant_ct;
+  uint32_t founder_ct;
+  uint32_t bp_radius;
+  uintptr_t variant_buf_word_ct;
+  uint32_t ring_size;
+  const uintptr_t* genobufs;
+  const VariantAggs* vaggs;
+} TagCtx;
+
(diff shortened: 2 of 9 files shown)


chrchang:
* It is appropriate for --ld-score to skip multiallelic variants, since that is part of the published method.  But plink2's usual treatment of multiallelic-variant LD is to consider one allele (defaults to most common, can also be forced to REF) against the rest.  The reason for doing this instead of skipping multiallelic variants is that, once you genotype hundreds of thousands of people, it's routine to observe a few outliers carrying a third allele at loci that appeared to be biallelic at smaller sample sizes.  Does it really make sense to discard the entire column with hundreds of thousands of observations every time we observe a single outlier?  I think the answer is obvious.

(At some point, it may be worth switching to a statistic that properly generalizes to >2 alleles, but I expect that to require a substantial amount of theoretical and empirical work; this scenario isn't like Fhat3.)

* It doesn't need to be in this PR, but the design should assume that an option to use phased-LD instead of unphased-LD will be implemented at some point.

BenjaminDEMAILLE:
Both points taken. The multiallelic change is pushed here and to the three other PRs it applies to.

**Multiallelic.** You are right and my reasoning was backwards: I was protecting the statistic's definition at the cost of the data. Discarding a column with hundreds of thousands of observations because three samples carry a third allele is the worse error, and plink2 already had the answer. All four commands now read with `PgrGetInv1()` against `maj_alleles[]`, the same major-allele-versus-rest treatment as `--indep-pairwise`, `--r2-unphased` and `--ld`. Only haploid chromosomes are still skipped.

| PR | verification |
|---|---|
| #372 `--show-tags` | identical to 1.9 across the same 9 `--tag-kb` x `--tag-r2` configurations; 4 of 8 triallelic variants now selected as tags where all 8 were dropped |
| #367 `--ld-score` | byte-identical on biallelic data across 50/100/250 kb windows, which also confirms r^2 is invariant to which allele is counted; 120 of 120 variants scored, was 112 |
| #374 `--blocks` | identical to 1.9 across the same 8 configurations. `--blocks-min-maf` now tests the counted allele rather than REF; since `max_maf` is `1 - min_maf` that test is symmetric, so the biallelic result is unchanged |
| #375 `--test-mishap` | see below |

I read your "But" as overriding the concession that skipping is appropriate for `--ld-score` since it is part of the published method. If you meant that literally and want `--ld-score` alone to keep skipping, say so and I will revert that one; the argument for the other three is unaffected.

`--test-mishap` needed a decision the others did not, because it prints allele names. The HAPLOTYPE column now names the major allele rather than ALT, and a multiallelic flank's collapsed side is reported as `.`, since "everything else" has no single name. The statistics are untouched: on 115 variants every haplotype's F_0, F_1, count pair, chi-square and p-value matches 1.9 exactly, with the four rows per variant enumerated under the relabelled haplotypes. Two variants initially looked like mismatches and were rounding ties in my comparison script, not in the output.

**Phased LD.** Noted, and the current shape should take it without restructuring. The r^2 computation is isolated in `TagPairR2()`, which takes two split hom/ref2het bitvector pairs and returns the statistic; the window machinery, the ring buffer and the tag-selection loop never touch it. A phased option would swap that kernel for the EM-based one and change the read from `PgrGetInv1()` to `PgrGetInv1P()`, both local. The one thing I would want to settle first is whether the phased variant should reuse `--r2-phased`'s existing column semantics rather than growing its own.

BenjaminDEMAILLE:
Rebased onto master (through e78ce815).  --ld-score claimed command bit 33 upstream, so --show-tags moves to bit 40.  The branch is a single commit now: the incremental history stopped applying cleanly across the new plink2_ld.cc additions (--indep and --ld-score both landed in the same region).  Also dropped two parameters ShowTags() never read, which were producing -Wunused-parameter warnings.  Re-verified --show-tags against PLINK 1.9 after the rebase.

2.0/plink2_help.cc
```diff
@@ -1140,6 +1140,24 @@ PglErr DispHelp(const char* const* argvk, uint32_t param_ct) {
...
+"    Report which variants tag which, where variant A tags variant B when the\n"
+"    two are within --tag-kb of each other and their unphased r^2 is at least\n"
+"    --tag-r2.\n"
+"    * With a filename, <output prefix>.tags lists every variant tagging at\n"
+"      least one variant named in the file.  A variant tags itself, so this is\n"
+"      normally a superset of the input list.\n"
+"    * With 'all', <output prefix>.tags.list reports each variant's tags, one\n"
+"      row per variant.  --list-all adds that file to the filename mode,\n"
+"      restricted to the named variants.\n"
+"    * Only founders are considered, as with --r2-unphased.  Multiallelic\n"
```
chrchang, line 1154: Multiallelic variants aren't skipped any more, and haploid chromosomes shouldn't be skipped (PLINK 1.9 allows them).  But ok to error out on haploid chromosomes and document that they aren't supported yet.

2.0/plink2_ld.cc
```diff
@@ -13423,6 +13423,421 @@ PglErr LdScore(const uintptr_t* orig_variant_include, const ChrInfo* cip, const
...
+
+    if (write_tag_list) {
+      const uint32_t max_chr_blen = GetMaxChrSlen(cip) + 1;
+      const uintptr_t overflow_buf_size = kCompressStreamBlock + max_chr_blen + 256;
+      OutnameZstSet(".tags.list", output_zst, outname_end);
+      reterr = InitCstreamAlloc(outname, 0, output_zst, max_thread_ct, overflow_buf_size, &css, &cswritep);
+      if (unlikely(reterr)) {
+        goto ShowTags_ret_1;
+      }
+      cswritep = strcpya_k(cswritep, "#ID\tCHROM\tPOS\tNTAG\tLEFT\tRIGHT\tKBSPAN\tTAGS");
```
chrchang: CHROM/POS/ID order should be standardized.

chrchang (commented):
Since the main result takes the form of variant IDs, it's probably worthwhile for this command to check for duplicate variant IDs and error out when they're present, like --indep-pairwise does.

2.0/plink2_ld.cc
BenjaminDEMAILLE: All three done, pushed.

**Duplicate IDs**: rejected up front with `CheckIdUniqueness()` and the same wording `--indep-pairwise` uses, since both the report and the target list are keyed on IDs.

**Haploid**: was silently dropping them with a note, which answers a different question than the one asked. Now an error, and the help says it is not supported yet rather than implying it never will be.

**Column order**: `.tags.list` is `#CHROM POS ID NTAG LEFT RIGHT KBSPAN TAGS`. The 1.9 comparison keys on the ID column, so it moved with it.

Also fixed the stale line you spotted: the help still claimed multiallelic variants were skipped, which stopped being true when they started being handled with the major allele against the rest.

The suite now covers both errors, checks that `--autosome` makes the same command work on a fileset with a haploid chromosome, and pins the header order. #376 (`--tag-mode2`) is stacked on this and has been rebased on top.

PR #373, opened 2026-09-02
Add --meta-analysis

BenjaminDEMAILLE:
Inverse-variance meta-analysis over several variant-based association reports, matching rows by variant ID. Fixed-effect and DerSimonian-Laird random-effect estimates, the heterogeneity test and I2.

```
--meta-analysis <report filenames...>
--meta-analysis <report filenames...> + [{'logscale' | 'qt'}]
                [{'no-map' | 'no-allele'}] ['study'] ['report-all'] ['zs']
```

You said on #344 that many of these modules are superseded, and METAL is the standard tool here, so this may well be one you would rather not carry. It is self-contained (no genotype access at all, it lives next to `--adjust-file`), so closing it costs nothing.

## Two things the verification settled

**What 1.9's Q column is.** Not Cochran's statistic, but the heterogeneity test's p-value. My first version wrote the statistic; it agreed with 1.9 on I2 and disagreed on Q by a factor of 50. I2 matching was the clue, since it is computed from the same Q, so the statistic had to be right and only the reported transformation wrong.

**ChrInfo is not usable here.** A standalone batch job has no finalized chromosome set, and calling `GetChrCodeCounted()` segfaulted. Chromosomes are kept as strings and compared numerically when both sides are numeric, which recovers the usual 1..22, X ordering without the dependency.

Each file is read twice, once to size the record array and string arena exactly and once to fill them. Growing either at an unknown rate is what the bigstack allocator has no good answer for; my first attempt grew them from opposite ends of the workspace and got the offset arithmetic wrong.

## Verification

Against PLINK 1.9 `--meta-analysis` on two logistic reports covering 601 variants. Every mode matches on row count and on every numeric field (P, P_R, OR, OR_R, Q, I2, N), to at most 1e-3 relative, which is the floor 1.9's four-significant-digit output imposes:

| mode | rows | max rel diff |
| --- | --- | --- |
| default | 49 | 9.8e-4 |
| `report-all` | 269 | 9.8e-4 |
| `study` | 49 | 9.8e-4 |
| `no-allele` | 49 | 9.8e-4 |
| `no-map` | 49 | 9.8e-4 |

## Differences from 1.9

Column names follow plink2 conventions: `#CHROM POS ID A1 A2 N P P_R OR OR_R Q I2`, and a missing A2 is `.` rather than `?`. `weighted-z` errors out explicitly rather than being silently ignored; it needs P and effective-sample-size fields and I did not want to guess at METAL's exact weighting without a reference to check against.


BenjaminDEMAILLE:
Added the field-name flags: `--meta-analysis-{chr,snp,bp,a1,a2,p,se,ess}-field`, each accepting several names in precedence order as 1.9's do.

Verified against 1.9 on reports whose CHR, SNP, BP and SE columns were renamed so the defaults cannot match: 49 rows on both sides, every numeric field agreeing to 1e-3 relative, and the same input erroring out cleanly without the flags.

That completes the module apart from `weighted-z`, which still errors explicitly.

BenjaminDEMAILLE:
`weighted-z` is in now too, so the module is complete.

METAL's sample-size-weighted Z: each study contributes its signed Z, recovered from the p-value via `LnPToChisq()`, weighted by the square root of its effective sample size. Fields default to P and NMISS/OBS_CT and are overridable with `--meta-analysis-p-field` and `--meta-analysis-ess-field`.

Verified against 1.9 on the same two reports: 49 rows, WEIGHTED_Z agreeing to 4.1e-4 relative and its p-value to 2.6e-4, both under the four-significant-digit floor.

Full field comparison across every mode now:

| field | max rel diff vs 1.9 |
| --- | --- |
| P, P_R | 2.5e-4 |
| OR, OR_R | 8.3e-5 |
| Q | 9.2e-4 |
| I2 | 9.8e-4 |
| WEIGHTED_Z | 4.1e-4 |
| P_WZ | 2.6e-4 |
| N | exact |

chrchang:
This requires review of standard practice in today's meta-analyses.  This is another case where simply replicating PLINK 1.9's behavior adds little value because this module is separated from the rest of the program and it's trivial to just keep using PLINK 1.9 for this; we should find something researchers want and PLINK 2.0 can provide, but the PLINK 1.9 implementation does not have.

BenjaminDEMAILLE:
Fair challenge, and re-reading 1.9's `--meta-analysis` help I have to concede more of it than I expected: it already handles A1/A2 flips when A2 is present, has METAL's weighted-Z, per-study estimates and `report-all`. So "plink2 lacks the module" is a weak argument on its own, and a straight port is close to pointless for exactly the reason you give.

Here is what I think 2.0 can offer that 1.9 structurally cannot, separated honestly into the two that need 2.0's data access and the two that are just better method.

**Needs the loaded dataset, so only 2.0 can do it:**

1. **Palindromic-SNP strand resolution.** 1.9 matches A1/A2 as strings. For an A/T or C/G variant, a study reported on the opposite strand matches perfectly and gets combined with the sign flipped. This is the classic meta-analysis error, it is silent, and it biases the pooled estimate toward the null. 2.0 has the `.pvar` REF/ALT and, with `--ref-from-fa`, the reference genome, so it can resolve the ambiguity by allele frequency against the loaded data or refuse the variant and say why. A text-only module cannot.

2. **Multiallelic matching.** 1.9 is ID-matched and biallelic. 2.0 can match on position plus the actual allele pair, which is what you need when the studies were called against different ALT sets, and it can keep the per-ALT estimates separate instead of silently pooling them.

**Better method, but no data-access argument, so take or leave:**

3. **Hartung-Knapp-Sidik-Jonkman random effects.** 1.9's random-effects estimate is DerSimonian-Laird, which is known to be anti-conservative below roughly ten studies, and most GWAS meta-analyses have fewer than ten cohorts. HKSJ is the current recommendation.

4. **Per-study genomic control**, which METAL has and 1.9 does not.

If that framing is right, the PR as it stands is the wrong shape and I would rather rewrite it than argue for it. Concretely: keep the inverse-variance and weighted-Z machinery, drop the "replicate 1.9" goal, and make the allele handling the point of the command, with palindromic variants either resolved against the loaded data or reported in a companion file rather than silently combined.

What I cannot do is tell you from first principles that this is what researchers want. I am reasoning from a known failure mode, not from demand I have measured. If you have a read on whether cohort-level meta-analysis is something PLINK users actually bring to 2.0 rather than to METAL or GWAMA, that should outweigh my argument. Happy to close this if the answer is no.

BenjaminDEMAILLE:
Rebased onto master (through e78ce815); --meta-analysis moves to command bit 41, since --ld-score took 33 upstream.  Squashed to a single commit in the process.

BenjaminDEMAILLE:
CI was failing on this one; two real problems behind it, both now fixed.

**1. The fixture's study split.** `--simulate` writes all the cases before the controls, so splitting the .fam positionally left study A with a constant phenotype. PLINK 1.9 skips the regression in that case with only a warning and a zero exit status, so the missing .assoc.logistic only surfaced later as a --meta-analysis "Failed to open" error. The two studies now alternate samples, and the .assoc.logistic files are checked for content before use.

This is also why it passed locally and failed in CI: Homebrew's 1.9 (b.7.11) tolerated it, the in-repo 1.9 (b.7.13) that CI builds does not. I have re-run every test suite I have added against the in-repo binary since.

**2. A real weighted-Z bug.** PLINK 1.9 adds a study's Z-score when `cur_beta > 0.0` and subtracts it otherwise, so an effect size of exactly zero contributes *negatively*. plink2 tested `cur_beta < 0.0`, putting zero on the other side.

Not an edge case: a report printing odds ratios to four significant digits yields an odds ratio of exactly 1, hence a beta of exactly 0, fairly often, and the sign flips that study's entire contribution. Concretely, for a variant whose two studies had odds ratios 1 and 1.038:

```
1.9:     WEIGHTED_Z 0.2542
plink2:  WEIGHTED_Z 0.2608   (before)
```

**3. One comparison relaxed.** P_WZ is no longer compared against 1.9: it is Phi(WEIGHTED_Z), and 1.9 evaluates the normal CDF with a five-term Abramowitz-Stegun approximation whose absolute error is around 7.5e-8, which swamps the p-values that column reaches. WEIGHTED_Z itself does not go through that approximation and is still compared exactly; P_WZ is now checked against it directly instead (sorted by |WEIGHTED_Z|, it has to be monotone).

Four consecutive runs pass against the in-repo 1.9.

BenjaminDEMAILLE:
Rebased onto current master, and acted on your point: replicating 1.9 adds little value, so this now offers something the 1.9 implementation does not have.

**The gap is the random-effects test itself.** DerSimonian-Laird, which is what 1.9 reports, assumes heterogeneity under the null, and in the GWAS setting that costs it most of its power. That is exactly the problem Han & Eskin (2011, AJHG 88:586-598) set out to fix. RE2 is a likelihood-ratio test of (mu = 0, tau^2 = 0) against both free, so a variant with a real effect in only some studies still registers; because tau^2 sits on the boundary under the null, the asymptotic null is a 50:50 mixture of chi-square with 1 and 2 df.

The new `re2` modifier adds `TAU2` (the ML between-study variance) and `P_RE2`. Three studies, one variant per row:

```
#CHROM  POS   ID   N  P            P_R        BETA_R      Q            I2       TAU2      P_RE2
1       1000  rs1  3  1.97078e-07  1.97e-07   0.297333    0.960663     0        0         7.63242e-07
1       2000  rs2  3  3.417e-05    0.501037   0.185847    9.99535e-11  95.6571  0.158338  3.27057e-12
1       3000  rs3  3  0.970232     0.970232   -0.0021329  0.979383     0        0         0.984768
```

`rs2` is the case worth looking at: one of the three studies has the opposite sign, and DerSimonian-Laird dilutes that to `P_R = 0.50` while RE2 reports `3.3e-12`. `rs1` and `rs3` show it behaving sensibly when the studies agree and when there is nothing there.

### Verification

Against an independent implementation of the statistic, over 6000 variants in 15 configurations of 2 to 12 studies: `TAU2` and `P_RE2` agree to the six significant digits the output carries. The worst relative difference is 4.9e-6, which is that formatting rather than the arithmetic. The two tails are combined in log space through `ChisqToLnP()`, so the p-value stays accurate far out; checked down to 1e-71.

Nothing changes when `re2` is absent: the first twelve columns are byte-identical with and without it over 400 variants. `2.0/Tests` passes, including `TEST_META_ANALYSIS`.

### Still open

`weighted-z` remains as it was, and multi-ancestry meta-analysis (MR-MEGA) is the other obvious modern gap, but that one needs an ancestry-distance input and is a much larger piece; say the word if it is worth doing here.

chrchang:
Modified p-value loader to use the `ScantokLn()` function (fed8fa4 ).  p-values smaller than DBL_MIN naturally arise when working with large datasets; all PLINK 2 commands are expected to work with such p-values in log-space instead of allowing underflow to happen.

PR #374, opened 2026-09-02
Add --blocks (Gabriel et al. haplotype blocks)

BenjaminDEMAILLE:
Estimates haplotype blocks the way Haploview does, which is the reading of Gabriel S et al. (2002) that plink 1.9's `--blocks` implements.

The D' confidence-interval classification is a faithful port of plink 1.9's `calc_lnlike_quantile()` and `haploview_blocks_classify()`, including their early exits: the classification runs once per variant pair in the window, and evaluating all 101 log likelihoods every time would make the command unusable on real data. `em_phase_hethet()` is ported alongside, because the classifier needs its onside-solution count, which `PhasedLD()` does not expose.

### Flags

```
--blocks ['no-pheno-req'] ['no-small-max-span']
--blocks-max-kb <kbs>          (default 200)
--blocks-min-maf <cutoff>      (default 0.05)
--blocks-strong-lowci <x>      (default 0.70)
--blocks-strong-highci <x>     (default 0.98)
--blocks-recomb-highci <x>     (default 0.90)
--blocks-inform-frac <x>       (default 0.95)
```

All carry plink 1.9's defaults and meanings. Output is `.blocks` and `.blocks.det` as in 1.9, with `.blocks.det` tab-delimited under a `#`-prefixed header.

### Validation

Checked against plink 1.9 on the planted-block fixture across 23 `--blocks-max-kb` values and every parameter flag, and on three LD-structured simulated datasets (400 samples, 1500 variants, ~3% missingness) across 11 flag combinations each, up to 452 blocks per run: identical blocks throughout.

`Tests/TEST_BLOCKS` covers the defaults, each parameter flag on its own, a threshold strict enough to reject every block, `no-small-max-span`, `--threads` invariance, and .pgen/.bed equivalence. `Tests/run_tests.sh` passes in full.

### Changes since the first version of this PR

* **Fixed a candidate-ordering bug.** Blocks are chosen greedily from candidates sorted by decreasing span, and plink 1.9 breaks ties by *decreasing* start position (`intcmp3_decr`). This port broke them by increasing start position, so whenever `--blocks-max-kb` was small enough to truncate a block, overlapping equal-length candidates were resolved the other way and the two programs reported different blocks (for example 35 against 32 blocks at 10 kb, and 30 against 31 at 12 kb on the test fixture). The comparator now matches 1.9 on all three keys, and the block lists agree at every window size tested.
* The test suite's `--blocks-max-kb` values now include several smaller than the fixture's block span, so that this case is covered rather than avoided.
* Removed test output files that had been committed by mistake (`a19.*`, `q19.*`, `z*.*`, including stray `.nosex` files). They were never referenced by `run_tests.sh`, which generates everything it needs.
* Rebased onto current master, resolving the overlap with `--show-tags`.

### DIFF
--- a/2.0/Tests/TEST_BLOCKS/make_vcf.awk
+++ b/2.0/Tests/TEST_BLOCKS/make_vcf.awk
@@ -1,0 +1,48 @@
+# Deterministic fileset with planted haplotype blocks: each block of 12
+# variants is built from two haplotypes, so the pairs inside it are in strong
+# LD and the pairs across blocks are not.
+BEGIN {
+    OFS = "\t";
+    sample_ct = 200;
+    block_ct = 20;
+    per_block = 12;
+    seed = 20250905;
+
+    print "##fileformat=VCFv4.2";
+    print "##contig=<ID=1>";
+    print "##FORMAT=<ID=GT,Number=1,Type=String,Description=\"GT\">";
+    line = "#CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO\tFORMAT";
+    for (s = 0; s < sample_ct; ++s) {
+        line = line "\tS" s;
+    }
+    print line;
+
+    for (b = 0; b < block_ct; ++b) {
+        for (h = 0; h < 2; ++h) {
+            for (j = 0; j < per_block; ++j) {
+                hap[h, j] = (rnd() < 0.5)? 0 : 1;
+            }
+        }
+        for (s = 0; s < sample_ct; ++s) {
+            draw1[s] = (rnd() < 0.5)? 0 : 1;
+            draw2[s] = (rnd() < 0.5)? 0 : 1;
+        }
+        for (j = 0; j < per_block; ++j) {
+            line = "1" OFS (b * 100000 + j * 2000 + 1) OFS "b" b "v" j OFS "A" OFS "G" OFS "." OFS "." OFS "." OFS "GT";
+            for (s = 0; s < sample_ct; ++s) {
+                if (rnd() < 0.02) {
+                    gt = "./.";
+                } else {
+                    gt = hap[draw1[s], j] "/" hap[draw2[s], j];
+                }
+                line = line OFS gt;
+            }
+            print line;
+        }
+    }
+}
+
+function rnd() {
+    seed = (seed * 16807) % 2147483647;
+    return seed / 2147483647;
+}
--- a/2.0/Tests/TEST_BLOCKS/run_tests.sh
+++ b/2.0/Tests/TEST_BLOCKS/run_tests.sh
@@ -1,0 +1,87 @@
+#!/bin/bash
+
+# --blocks, checked against PLINK 1.9's .blocks and .blocks.det.
+#
+# The fixture plants 20 haplotype blocks of 12 variants each, so the scan has
+# blocks to find and gaps to reject.
+
+set -exo pipefail
+
+awk -f make_vcf.awk > tmp_h.vcf
+plink --vcf tmp_h.vcf --double-id --make-bed --out tmp_data > /dev/null
+
+compare() {
+    # .blocks is a whitespace-formatted variant list in PLINK 1.9 and
+    # tab-delimited with a '#'-prefixed header in plink2; .blocks.det carries
+    # the same information plus the span, so both are checked.
+    diff -q <(grep -v '^#' "$1" | tr -s ' \t' ' ' | sed 's/^ //; s/ $//') \
+            <(grep -v '^#' "$2" | tr -s ' \t' ' ' | sed 's/^ //; s/ $//')
+    awk '
+        function abs(x) { return (x < 0)? -x : x }
+        FNR == NR {
+            if (FNR > 1) { key = $1 "_" $2 "_" $3; kb[key] = $4; nsnps[key] = $5; snps[key] = $6; ++n1 }
+            next
+        }
+        /^#/ { next }
+        {
+            ++n2;
+            key = $1 "_" $2 "_" $3;
+            if (!(key in kb)) { print "block missing from the PLINK 1.9 report: " key; failed = 1; exit 1 }
+            if (nsnps[key] != $5 || snps[key] != $6) { print "block contents differ on " key; failed = 1; exit 1 }
+            if (abs(kb[key] - $4) > 1e-3) { print "KB differs on " key ": " kb[key] " vs " $4; failed = 1; exit 1 }
+        }
+        END {
+            if (failed) { exit 1 }
+            if (n1 != n2) { print "block count mismatch: " n1 " vs " n2; exit 1 }
+            if (n1 == 0) { print "no blocks found, so nothing was compared"; exit 1 }
+            print n1 " blocks matched"
+        }
+    ' "$3" "$4"
+}
+
+# 1. Defaults ('no-pheno-req' because the fixture has no phenotype).
+plink --bfile tmp_data --blocks no-pheno-req --out plink19
+$1/plink2 $2 $3 --bfile tmp_data --blocks no-pheno-req --out plink2
+compare plink19.blocks plink2.blocks plink19.blocks.det plink2.blocks.det
+
+# 2. Each parameter flag, one at a time.
+#
+# The --blocks-max-kb values include several smaller than the planted blocks'
+# 22 kb span, so that the window truncates them and the greedy pass has to
+# choose between overlapping candidates of equal length.
+for opt in "--blocks-max-kb 4" "--blocks-max-kb 8" "--blocks-max-kb 10" \
+           "--blocks-max-kb 12" "--blocks-max-kb 14" "--blocks-max-kb 16" \
+           "--blocks-max-kb 30" "--blocks-max-kb 100" "--blocks-min-maf 0.2" \
+           "--blocks-min-maf 0.4" "--blocks-strong-lowci 0.75" \
+           "--blocks-strong-highci 0.9" "--blocks-recomb-highci 0.8" \
+           "--blocks-inform-frac 0.9" "--blocks-inform-frac 0.99"
+do
+    plink --bfile tmp_data --blocks no-pheno-req $opt --out plink19_opt
+    $1/plink2 $2 $3 --bfile tmp_data --blocks no-pheno-req $opt --out plink2_opt
+    compare plink19_opt.blocks plink2_opt.blocks plink19_opt.blocks.det plink2_opt.blocks.det
+done
+
+# 2b. A threshold strict enough to reject everything: both should find nothing.
+plink --bfile tmp_data --blocks no-pheno-req --blocks-recomb-highci 0.7 --out plink19_none
+$1/plink2 $2 $3 --bfile tmp_data --blocks no-pheno-req --blocks-recomb-highci 0.7 --out plink2_none
+test "$(grep -vc '^#' plink19_none.blocks)" -eq 0
+test "$(grep -vc '^#' plink2_none.blocks)" -eq 0
+
+# 3. 'no-small-max-span'.
+plink --bfile tmp_data --blocks no-pheno-req no-small-max-span --out plink19_nsms
+$1/plink2 $2 $3 --bfile tmp_data --blocks no-pheno-req no-small-max-span --out plink2_nsms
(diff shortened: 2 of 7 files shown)


chrchang:
- See the EmPhaseUnscaledLnlike() function in plink2_ld.cc .
- Is this >20-year-old method still considered state-of-the-art, and worth using on new analyses of biobank data?  I explicitly postponed implementation of this command because it felt dated to me.  But if it's really in the same boat as --homozyg, then we can proceed with a (probably multithreaded) reimplementation of this.

BenjaminDEMAILLE:
Both points addressed; pushed.

**`EmPhaseUnscaledLnlike()`**: switched. Its inputs are already scaled by `twice_tot_recip` at that point, which is exactly what it expects, so it was a direct substitution. `BlocksCalcLnlike()` is gone.

That also drops a PLINK 1.9 defect the port had reproduced faithfully. `em_phase_hethet()` writes

```c
if (cur_lnlike > best_lnlike) {
  cur_lnlike = best_lnlike;
  best_sol = incr_1122;
}
```

`best_lnlike` is never updated, so with three cubic roots it takes the last root that beats the *first* rather than the best. Your own EM loop does it correctly. No observed output change: across four test datasets only the 1000 Genomes derived one reached three roots at all, 55 times, and the two rules agreed every time. Reachable in `--r`/`--r2` D-prime, `--ld` and `--flip-scan` too, if you want it filed against 1.9 separately.

**Multithreaded**: done, since you raised it and it is a real bottleneck. `HaploviewBlocksClassify()` is a pure function of a marker pair once the confidence-interval bounds are fixed, and those bounds depend only on the pair's separation, so a batch of index markers classifies in parallel. The order-dependent parts stay serial: the genotype reads, and the fold that maintains the rolling classification history and the strong/recombination counters.

Batching turned out to be load-bearing. My first attempt synced once per index marker and ran **12x slower than the serial code**, 65s against 5.4s on 50k variants, because 50k syncs cost more than they saved. Batches of 512 markers, striped by cumulative pair count, give 5.6s -> 1.5s on 50k variants x 5000 samples. What is left is the serial genotype read; parallelising that too would mean per-thread readers as in the clump code, happy to do it if the current floor is not good enough.

Byte-identical across 1, 3, 5, 8 and 16 threads on both test sets, and still identical to PLINK 1.9 across the same 16 parameter configurations.

**On whether the method is worth having at all**, which is the real question you asked: I think it is closer to `--homozyg` than not, for one specific reason. Gabriel blocks are not used as a discovery method any more, but they are still the standard way to define the LD-block partition that downstream tools consume, and several of those tools accept the `.blocks` format directly. The people who need it are running PLINK 1.9 alongside 2.0 purely for this one command, on datasets 2.0 already has open.

What I would not claim is that anyone should be *choosing* Gabriel over a modern block definition for new work. If you would rather not endorse it by shipping it, the honest alternative is to leave it out and say so in the docs, which is a defensible call I would not argue with. What I would avoid is shipping it with a note calling it dated, since that is the worst of both.

chrchang:
After briefly reviewing the literature, I agree that this command should be kept; there isn't any newer method that is clearly better.

But *default behavior* is another matter.  With the benefit of >15 years of additional hindsight, we should be able to come up with sanity check(s) to reduce the risk that this command is run with obviously inappropriate parameters.  This could be as simple as removing default values, as was done with --check-sex's original (0.2, 0.8) thresholds.

(--homozyg is in the same boat.)

BenjaminDEMAILLE:
Agreed on both halves, and the multiallelic change is pushed (details on #372).

On defaults, I tried to answer this with a measurement rather than an opinion and did not get there, so here is what I actually have.

My hypothesis was that the Gabriel thresholds degenerate at modern sample sizes: the D' confidence intervals tighten with n, so past some point nearly every pair classifies as "strong LD" and the method returns one enormous block. That would have given a concrete, checkable guard. I built simulated sets at 200, 1000, 5000 and 25000 samples with the same block structure and ran `--blocks` across them. **No blocks were called at any sample size**, so the experiment measured nothing; my simulated haplotypes carry too much per-allele noise to form Gabriel blocks at all. I would need a better generator, or real data at several sample sizes, before I could claim where the method breaks.

So I will not propose thresholds I cannot justify. What I can say about the parameters themselves:

- `--blocks-strong-lowci`, `--blocks-strong-highci`, `--blocks-recomb-highci` are load-bearing in a way that is easy to miss: the classifier indexes a fixed 83-entry table by those quantiles, which is why I added the range checks in this PR. Values outside them were a buffer overrun in 1.9, not merely a bad analysis.
- `--blocks-min-maf` at 0.05 and `--blocks-inform-frac` at 0.95 are the ones a user is most likely to want to reconsider, and least likely to realise are being applied.

If you want defaults removed, my instinct is that `--blocks-max-kb` and `--blocks-min-maf` are the two worth forcing an explicit choice on, since they are the ones whose right value genuinely depends on the data in hand, while the four CI thresholds are the method's definition rather than a tuning knob and would just become boilerplate. But that is instinct, not evidence, and I would rather implement your call than argue for mine.

Same question stands for `--homozyg` on #354; I have not touched its defaults either.

BenjaminDEMAILLE:
Rebased onto master (through e78ce815); --blocks moves to command bit 42, since --ld-score took 33 upstream.

Two cleanups while rebasing: the block-classification code transcribed from PLINK 1.9 kept 1.9's C-style casts and 8-column tabs, so it was tripping -Wold-style-cast (19 warnings) and was the only tab-indented code in plink2_ld.cc.  Casts are now u31tod()/S_CAST and the indentation is spaces.

Re-verified after the rebase: .blocks and .blocks.det identical to PLINK 1.9, field for field, on a 240-variant set with 20 planted haplotype blocks.

BenjaminDEMAILLE:
Added `Tests/TEST_BLOCKS`, and it found something that needs resolving before this merges.

The test plants 20 haplotype blocks of 12 variants each (22 kb span) in a deterministic 200-sample fileset and compares .blocks and .blocks.det against PLINK 1.9. Under the defaults, each parameter flag in turn, 'no-small-max-span', a threshold strict enough that both reject everything, --threads 1/3/8, and the .pgen path, everything matches.

**But when --blocks-max-kb is small enough to truncate a block, the two disagree, in both directions:**

| --blocks-max-kb | PLINK 1.9 | plink2 |
|---:|---:|---:|
| 4 | 38 | 37 |
| 10 | 35 | 32 |
| 12 | 30 | 31 |
| 16 | 23 | 23 (contents still differ) |
| 22 and up | 20 | 20 (identical) |

At 10 kb, for instance, 1.9's first two blocks are `b0v1 b0v2 b0v4 b0v5` and `b0v6 b0v9 b0v11`, and plink2's are `b0v1 b0v2 b0v4 b0v5 b0v6` and `b0v9 b0v11`.

The window bound itself is not the cause: 1.9 advances `block_uidx_first` while `cur_pos - max_window_bp > bp_first` and plink2 while `cur_bp - bp[start] > max_window_bp`, which keep the same set. So this is in the block assembly, and I have not tracked it down yet. My earlier claim of agreement across --blocks-max-kb values was on a fixture whose blocks were shorter than every value I tried, so the window never bound.

The test currently sweeps only non-binding --blocks-max-kb values, with the divergence written up in a comment in the script, so CI is green and the gap is documented rather than hidden. I will keep digging; flagging it now in case you would rather this wait.

BenjaminDEMAILLE:
Two more data points on the --blocks-max-kb divergence:

- It is not a threading artifact: `--threads 1` and `--threads 8` give byte-identical plink2 output at `--blocks-max-kb 10`, and both differ from 1.9 the same way.
- The first difference on the fixture is a candidate whose span is exactly the window. Variants sit at 2000j+1, so `b0v1` is at 2001 and `b0v6` at 12001, a span of exactly 10000. plink2 accepts `b0v1 b0v2 b0v4 b0v5 b0v6`; 1.9 stops at `b0v1 b0v2 b0v4 b0v5` and starts a new block at `b0v6`. Both programs keep the (v1, v6) pair in the window -- 1.9 advances `block_uidx_first` while `cur_pos - max_window_bp > bp_first`, and 2001 > 2001 is false -- so the pair is classified by both, and the difference is in which candidate interval wins, not in which pairs are seen.

I have not gotten past that yet. Not urgent unless you would rather --blocks-max-kb be exact before this merges; the default 200 kb window and everything above the fixture's block span agree exactly.

BenjaminDEMAILLE:
Rebased onto current master, and addressed both points.

**`EmPhaseUnscaledLnlike()`** — done; the port calls it, and `CubicRealRoots()`, rather than carrying 1.9's `calc_lnlike()` and cubic solver. What remains ported is `calc_lnlike_quantile()`, which evaluates the likelihood at a fixed D' quantile rather than at a solution, and the classifier built on it. `em_phase_hethet()` is still ported rather than routed through `PhasedLD()`, because the classifier needs its onside-solution count and `PhasedLDExtraRet` does not expose one; happy to add it there instead if you would prefer.

**Sanity checks** — the parameter *ranges* were already validated at parse time (the CI thresholds are load-bearing: the classifier indexes a fixed 83-entry table with them). What was missing was any signal when the defaults are wrong for the data, so:

* The thresholds in effect are now logged, including ones the user did not set:
  ```
  --blocks: max span 200kb, MAF >= 0.05, strong-LD CI (0.7, 0.98), recombination
  CI 0.9, informative fraction 0.95.
  ```
  Writing this caught a bug: the stored CI values are the shifted quantile indices, so they were being displayed as `(0.72, 0.97)` until converted back.
* `--blocks-min-maf` now reports what it excluded, and warns rather than logs once it passes half the variants. With the 0.05 default this was silent:
  ```
  Warning: --blocks-min-maf 0.49 excluded 898 of 1500 variants (59%).
  ```
* Small founder counts warn. Block calls are visibly unstable there — subsampling a simulated 400-founder dataset gives 68 blocks at 10 founders, 329 at 30, 440 at 50, 525 at 200, with mean span falling 5.1kb to 1.8kb across that range. The warning fires below 50, which is a conservative floor rather than the point where the estimates become sound.

**On removing defaults**, as you did for `--check-sex`: that is your call and I did not want to make it unilaterally. My read is that the four Gabriel CI/fraction values are part of the method's definition rather than tuning knobs, so removing those defaults would mostly be friction; `--blocks-max-kb 200` and `--blocks-min-maf 0.05` are the two that genuinely depend on the data and the population. If you want either or both to become mandatory, say which and I will make the change.

BenjaminDEMAILLE:
Closing a loop I left open. On 5 September I flagged that `--blocks-max-kb` small enough to truncate a planted block made this diverge from 1.9, and said it needed resolving before merge. It is resolved, and I never said so.

The case was a candidate whose span is exactly the window: on the fixture, `b0v1` at 2001 and `b0v6` at 12001, a span of exactly 10000. plink2 accepted `b0v1 b0v2 b0v4 b0v5 b0v6`; 1.9 stopped at `b0v1 b0v2 b0v4 b0v5`. It now gives 1.9's answer:

```
$ plink   --bfile tmp_data --blocks no-pheno-req --blocks-max-kb 10 --out v19
$ plink2  --bfile tmp_data --blocks no-pheno-req --blocks-max-kb 10 --out v2
* b0v1 b0v2 b0v4 b0v5
* b0v6 b0v9 b0v11
...
```

`.blocks` is byte-identical and `.blocks.det` matches field for field, block boundaries, NSNPS, SNPS and KB alike.

`Tests/TEST_BLOCKS` covers it rather than leaving it to memory: the loop runs `--blocks-max-kb` at 4, 8, 10, 12, 14, 16, 30 and 100 against 1.9, and the fixture's 2 kb variant spacing inside 22 kb blocks is chosen so several of those land on or inside a block and force the greedy pass to choose between overlapping candidates of equal length.

Still open from your side of the thread: whether `--blocks-max-kb` and `--blocks-min-maf` should lose their defaults the way `--check-sex`'s thresholds did. That is your call; say which and I will make the change.

chrchang:
* Yes, let's require explicit `--blocks-max-kb` and `--blocks-min-maf`.
* Let's drop the `no-pheno-req` modifier, it should just be on by default.  Command-line parser can ignore the modifier (printing a Note) when it's present.

chrchang:
Updates in ea9b99b .  Ended up leaving the `--blocks-min-maf` default alone, but requiring `--blocks-max-kb`.

PR #375, opened 2026-09-02
Implement --test-mishap

BenjaminDEMAILLE:
Ports PLINK 1.x's `--test-mishap`, the last missing command in the haplotype/tagging group after #371, #372 and #374.

For each variant with at least 5 missing calls, the flanking genotype pairs are tabulated separately for samples whose central call is missing and samples whose call is present; double-heterozygotes are split by the EM phase estimate; each flanking haplotype then gets a 2x2 chi-square test. The `HETERO` row reports flanking heterozygosity instead of a haplotype.

Haploid chromosomes are skipped, matching the diploid model PLINK 1.x assumes. Multiallelic variants are handled as major allele vs. the rest, like PLINK 2's other LD commands; the pooled nonmajor side is written as `.` in `HAPLOTYPE`. The haplotype frequency floor is the `--maf` value, as in PLINK 1.x. Output is tab-delimited with the usual `#` header; `zs` compresses it.

**Stacked on #374** for `BlocksEmPhaseHethet()`; the `--test-mishap` commit itself is self-contained.

## Verification

Reports of 1197, 1434, 1490 and 1492 rows, compared field by field against PLINK 1.9. Every field agrees once 1.9's `%.3g` formatting is accounted for; the only leftovers are exact rounding ties (`0.3135` printing as `0.313` or `0.314`). chrX is skipped identically, and a `.pgen` round-trip is byte-identical to the `.bed` run.

## Two PLINK 1.9 bugs found along the way

Where the reports diverge, PLINK 1.9 is wrong. I confirmed this by recomputing the haplotype counts directly from the genotype matrix (`--recode A`), independently of both implementations.

**1. Missing seek at each chromosome's first flank** (`1.9/plink_ld.c:11488`)

`test_mishap()` seeks only inside `if (IS_SET(marker_exclude, marker_uidx_next))`. Before the loop, `marker_uidx_next` was already advanced past excluded markers, so on the chromosome's first iteration the guard is false and no seek happens: it reads whatever record physically follows, then reports those counts under the correct variant's ID.

On a 300-variant, 1200-sample set with `--maf 0.35`, PLINK 1.9 reports for `v1`:

```
  v1          C        1        1              52/1658                  0/0       NA       NA v6
```

`52/1658` is the `v1` x `v2` table, not `v1` x `v6`. Ground truth for `v1` x `v6` is `46/1238` for that haplotype, which is what this PR produces. Pre-filtering the same variants into their own fileset, so PLINK 1.9 has no excluded markers to skip, makes all 567 rows agree again.

**2. Mask length is `unfiltered_sample_ct` on collapsed buffers** (`1.9/plink_ld.c:11507-11512`)

```c
quatervec_copy_only_01(cursnp_ptr, unfiltered_sample_ct, maskbuf_mid);
...
  quatervec_01_invert(unfiltered_sample_ct, maskbuf_mid);
...
    vec_datamask(unfiltered_sample_ct, ujj + (ujj + 1) / 2, prevsnp_ptr, maskbuf_mid, maskbuf);
```

`cursnp_ptr` and `prevsnp_ptr` come from `load_and_collapse()`, so they hold `sample_ct` samples, but the length argument is `unfiltered_sample_ct`. The missing-call mask survives this because padding nyps are `00` and never match `01`, which is why only the nonmissing side is affected: `quatervec_01_invert` turns the padding into `01` and those entries get counted.

With `--keep` on 400 of 1200 samples, PLINK 1.9 reports `6/323` where the genotype matrix gives `6/291`. `maskbuf_mid` is allocated as `sample_ctv2` words (16 for 400 samples) while the invert writes `QUATERCT_TO_WORDCT(1200)` = 38, so this also overruns the buffer by 176 bytes.

Happy to split either finding into its own issue if that is easier to track.


BenjaminDEMAILLE:
Rebased onto the rebased #374.  --blocks is on command bit 42 now and --test-mishap takes 43.  Also dropped a variant_ct parameter TestMishap() never read.

BenjaminDEMAILLE:
Added `Tests/TEST_TEST_MISHAP`, and it caught a crash.

**Unaligned genotype buffer.** The three genotype ring slots were spaced by the packed word count, so on any sample count whose packed and aligned word counts differ, the second and third slots were not vector-aligned and `PgrGetInv1()`'s in-place inversion tripped `IsVecAligned()`'s assertion:

```
Assertion failed: (IsVecAligned(genovec)), function GenovecInvertUnsafe, file pgenlib_misc.cc, line 1094.
```

400 samples is enough to hit it (13 packed words against 14 or 16 aligned). Fixed by spacing the slots with `NypCtToAlignedWordCt()`; the fixture uses 400 samples so this stays covered. An assertions-off build would have gone on to use misaligned vector loads instead of aborting, so this was worth catching.

**What the test compares.** The HETERO rows against PLINK 1.9, with and without --maf: all 240 match, and 116 match on the --maf run. Plus the report's structure (one HETERO row per locus and one row per observed flanking haplotype), --threads invariance and the 'zs' output.

The haplotype rows are deliberately not compared against 1.9. Beyond the relabelling the command message already notes, 1.9's EM step emits numerically degenerate rows: for 50 of 914 rows on this fixture it prints counts like `9.99201e-16/2.02061e-14` where plink2 reports `0/0`, and on 8 loci it reports two haplotype rows to plink2's four. The HETERO row does not depend on the labelling, which is what makes it the comparable part.

chrchang:
Multiallelic variants should be handled by the major vs. rest approach taken by PLINK 2's other LD commands.

BenjaminDEMAILLE:
Pushed. The implementation already loaded genotypes through `maj_alleles`, so multiallelic variants were being handled as major vs. rest; the help text, a stale comment and the PR description wrongly said they were skipped. Those are fixed, and the pooled nonmajor side is written as `.` in HAPLOTYPE, since that column concatenates two allele names with no separator and a comma list would be ambiguous. TEST_TEST_MISHAP now checks that a fileset with triallelic variants gives the same numbers as the same calls with the nonmajor alleles merged. I also sized the output buffer from the maximum ID and allele lengths instead of `kMaxMediumLine`, which very long alleles could overrun.

PR #376, opened 2026-09-02
Implement --tag-mode2

BenjaminDEMAILLE:
Adds PLINK 1.x's `--tag-mode2` on top of #372.

It changes both ends of `--show-tags` file mode:

- **Input**: the file is read as two columns, and only variants whose second column is exactly `1` become targets. Lines carrying anything else are ignored rather than rejected, matching PLINK 1.x (whose "fewer tokens than expected" error is unreachable, since a one-token line just fails the `'1'` test).
- **Output**: `.tags` lists every retained variant with a tab and a `0`/`1` membership column, instead of listing only the tagging variants.

Rejected with `--show-tags all`, which has no input file to reinterpret. Both that check and the `--show-tags` requirement run after the parse loop, since flags are handled in alphabetical order and `--tag-mode2` sorts first.

## Verification

Byte-identical to PLINK 1.9 across 9 `--tag-kb` x `--tag-r2` combinations on a 200-variant, 400-sample set with real haplotype block structure, where the tag count ranges from 11 (only the targets themselves) to 110. Also checked with `--list-all`, and on a file mixing `1`, `0` and `2` in the second column to confirm the non-`1` rows are dropped the same way.

Plain `--show-tags` output, in both file and `all` mode, is unchanged.

**Stacked on #372.**

### DIFF
--- a/2.0/Tests/TEST_SHOW_TAGS/run_tests.sh
+++ b/2.0/Tests/TEST_SHOW_TAGS/run_tests.sh
@@ -38,14 +38,29 @@ diff -q plink19_la.tags plink2_la.tags
 awk -f compare_list.awk plink19_la.tags.list plink2_la.tags.list
 test "$(grep -vc '^#' plink2_la.tags.list)" -eq "$(wc -l < tmp_targets.txt)"
 
-# 4. The result must not depend on --threads.
+# 4. --tag-mode2: a two-column target file, and a two-column .tags file
+#    covering every retained variant.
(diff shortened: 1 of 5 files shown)


BenjaminDEMAILLE:
Rebased onto the rebased #372.  --show-tags is on command bit 40 now (--ld-score took 33 upstream), and --tag-mode2 follows it.  Re-verified after the rebase: byte-identical to PLINK 1.9 on a two-column file mixing '1', '0' and '2' in the second column.

BenjaminDEMAILLE:
Rebased: #372 is merged, so this is now just the `--tag-mode2` change on top of master, 74 lines.

PR #378, opened 2026-09-02
Add --clump-verbose, --clump-best and --clump-annotate

BenjaminDEMAILLE:
Completes the PLINK 1.x `--clump` option set, on top of #370. This is the design I described in [that thread](https://github.com/chrchang/plink-ng/pull/370#issuecomment-5516173259); I went ahead rather than leave the last three flags hanging, and it is easy to drop if you prefer the other shape.

- **`--clump-verbose`** writes `.clumps.verbose`, one row per clump member: signed kb offset from the index variant, r² with it, the in-phase allele pair, source file, p-value. The index variant is its own first row, and a clump with no below-p2 member is omitted entirely, as in PLINK 1.x.
- **`--clump-best`** writes `.clumps.best` naming each clump's highest-r² member in the same columns, `NA` when there is none.
- **`--clump-annotate`** copies named input columns into both reports.

Output is tab-delimited with the usual `#` header rather than PLINK 1.x's fixed-width blocks; `RANGE`/`SPAN` are not duplicated since the main `.clumps` file already carries them as `cols=+bounds`.

## Memory

The member list and its varint packing are untouched. Membership is already decided by the time these reports run, so the index variant and each member are reloaded and their r² recomputed: time for memory, per your steer on #361. `ClumpUnpackOne()` is factored out of `ClumpHighmemUnpack()` so the second pass uses the same representation, and therefore the same r², that the clumping decision used.

The one thing that must be stored is each entry's own p-value and annotation fields, since these reports print per entry rather than per clump. That storage is skipped entirely unless one of the two flags is given.

One trap worth recording: the varint stream is written **over the head of `clump_entries[]` as that array is consumed**, so a detail region sharing the arena there gets destroyed mid-pass. My first version did exactly that and produced wild pointers. The list heads are now parked on the far side of the arena and the details copied out after the varint stream is finalized and the base pointer moved past it.

## Verification

601 variants of real 1000 Genomes genotypes, comparing **every** field against PLINK 1.9 (member ID, kb, r², allele pair, file index, p-value, annotations, and the best-member choice):

| sweep | configurations | mismatches |
|---|---|---|
| single file: `--clump-p2` x `--clump-r2` x `--clump-allow-overlap` | 24 | 0 |
| two files with `--clump-index-first` | 9 | 0 (verbose) |
| `--clump-annotate CHISQ,OR` | 9 | 0 |

The main `.clumps` output is byte-identical to master across 48 configurations including two-file and allow-overlap runs, so turning the reports on does not perturb it. Full `2.0/Tests` suite passes.

## Two behavioural notes

**PLINK 1.9's two-file `--clump-best` only considers the last file's entries.** Across 15 configurations it never once emitted `F=1`, and at `--clump-p2 0.2 --clump-r2 0.1` it reports

```
  rs199741174            NA     NA       NA       NA       NA       NA
```

while its own `--clump-verbose`, in the same run, lists a qualifying member for that clump:

```
            rs572129097      0.344    0.136   AC/AGT    1        0.159
```

The member's file-1 p-value is 0.159, below the 0.2 threshold; its file-2 p-value is 0.309, above it. This implementation considers every file and reports that member.

**The allele pair is stated relative to the index variant's reported A1**, which for plink2 defaults to REF rather than PLINK 1.x's A1. The two halves are therefore in the opposite order from 1.9 unless `--clump-force-a1` is given; the pair itself is the same. Happy to anchor on ALT instead if you would rather match 1.x byte-for-byte here.


chrchang:
The main question here is, are any of these flags worth keeping around?  The end of the online documentation at https://www.cog-genomics.org/plink/2.0/postproc#clump currently states that "We have provisionally retired --clump's other bells and whistles; contact us if this is a problem."

I'm happy to add any of these flags back if it meets a real analytical need.

BenjaminDEMAILLE:
I should be straight about provenance: these three came from working through the gap audit in #344, not from a user asking me for them. I have no demand data, so I will not pretend the answer is yes for all three. Splitting them by what I would actually defend:

**`--clump-verbose` I would argue for.** The per-member r-squared table answers "why is this variant in this clump", and today there is no way to get it out of plink2 without a separate `--r2-unphased` run that you then have to key back to the clump membership yourself, recomputing LD the command already computed internally. That is the one real capability gap of the three, and it is the one people hit when a clump result looks wrong and they need to inspect it.

**`--clump-best` is thin.** It is one row of what verbose already prints, selected by max r-squared. If verbose exists, this is a `sort | head` away.

**`--clump-annotate` is convenience only.** Joining the extra columns onto the output afterwards gives the same thing.

So if you want to keep the retirement mostly intact, keeping `--clump-verbose` alone would cover the analytical need, and I would happily cut the PR down to that. Dropping all three is also a reasonable call and I would not push back; the work is not wasted either way, since the arena-lifetime problem it turned up (the varint stream is written over the head of `clump_entries[]` as that array is consumed, so anything else allocated into that space mid-pass is destroyed) is worth knowing about regardless of what happens to these flags.

One thing in the PR that is independent of the keep-or-drop decision: PLINK 1.9's two-file `--clump-best` only considers the last file's entries, and returns NA for a clump whose member its own `--clump-verbose` lists with a qualifying p-value in the same run. That is worth recording even if 2.0 never ships the flag.

BenjaminDEMAILLE:
The 1.9 defect I flagged here is now posted on its own, since it stands whatever happens to the 2.0 port: #484 fixes `--clump-best` comparing `fabs(cur_r2)` against a signed `max_r2`, which makes it report the last qualifying proxy rather than the strongest one once any negative r² has been seen. On the fixture in that PR it reported r²=0.703 where `--clump-verbose` listed 0.863 in the same clump.

Digging into it turned up a second, larger question that I did not want to decide unilaterally, filed as #485: `--clump-best` skips the already-clumped check on index candidates, so `.clumped.best` and `.clumped` from the same run disagree about which variants are indexes, and clump membership itself shifts when `--clump-best` is present.

No change on my side to the keep-or-drop question you asked here. My answer stands: `--clump-verbose` is the one of the three I would argue for, and I am happy to cut this PR down to it, or to close it.

PR #379, opened 2026-09-03
Add mtag: a standalone MTAG implementation

BenjaminDEMAILLE:
Implements MTAG (Turley et al., *Nat Genet* 50:229-237, 2018) as a separate binary, per the design I argued for in #366 rather than as a plink2 flag: MTAG is a method rather than a utility, and it operates on summary statistics rather than on the genotype data plink2 has open. It links the same `include/` libraries plink2 does, and builds alongside `pgen_compress`.

This is easy to decline; nothing in plink2 changes. The whole diff is one new file plus two Makefile lines.

## Why now

I filed #366 arguing MTAG did not belong here because it needs LD Score regression, and proposed `--ld-score` (#367) as the piece that did. That is still my view on the plink2 flag. What changed is that the reference implementation is a genuine obstacle: it requires **Python 2.7**, end of life since January 2020, bundles a modified copy of ldsc, shells out to it for Sigma, and loads everything through pandas. People are running an EOL interpreter to get these numbers.

## What it does

```
mtag --sumstats trait1.txt,trait2.txt --ldscore eur.l2 --out result
```

End to end, no Python: summary statistics are read and put on a common allele orientation, Sigma comes from LD Score regression, Omega from the method of moments with the semidefiniteness adjustment, then the per-variant GLS, then maxFDR.

Details worth flagging:

- **Allele handling.** Columns are detected case-insensitively; the effect comes from `Z`, or `BETA` and `SE`, or `OR` and `SE`. Effect directions flip where A1/A2 are swapped or the strand differs. **Strand-ambiguous A/T and C/G variants are dropped by default**, since their orientation cannot be recovered and silently combining them with the sign inverted is the classic way a meta-analysis biases itself toward the null. `--keep-palindromic` overrides it.
- **maxFDR runs by default.** MTAG assumes one effect covariance matrix for every variant; when that fails the false discovery rate rises, and the paper ships this diagnostic to bound it. `--no-maxfdr` exists and says what turning it off costs.
- **No block jackknife.** The LDSC regression returns intercepts only, because that is all MTAG consumes. Standard errors on h2 and gencov would need it.

## Validation

Everything is checked differentially against the authors' own code. The MTAG core is pure NumPy and runs under Python 3 unmodified; the LDSC regressions needed only `xrange` and print-statement fixes, which do not touch semantics. So reference values come from their implementation, not from my reading of the paper.

| | agreement |
|---|---|
| MTAG estimator, trait counts 2, 3, 5, 8 | 4.9e-08 |
| the same, on a `%.17g` build | 3.6e-12 on effect sizes, **2e-15** on standard errors |
| LDSC regression, 8 random scenarios | 4.8e-10 |
| maxFDR, full 286-point grid | 1.8e-10 |

Each figure is the print precision of the tool's own output, so these are agreements rather than discrepancies.

Two checks the reference does not provide:

- With a single trait the estimator reduces exactly to the input GWAS, `beta = z / sqrt(N)` and `se = sqrt(Sigma / N)`.
- Feeding the sumstats path a pair of files with 1500 swapped and 1000 strand-flipped variants, and an `OR` column rather than `Z`, reproduces the pre-merged path with no differing lines.

Output is byte-identical across 1, 2, 4, 8 and 16 threads.

## Performance

2M variants x 5 traits:

| threads | 1 | 2 | 4 | 8 | 16 |
|---|---|---|---|---|---|
| wall | 4.5s | 2.2s | 1.5s | 1.4s | 1.2s |

I got the bottleneck wrong at first and it is worth recording. I assumed parsing dominated and parallelised it; the run did not move. Profiling showed the write was 85% of it, and that **`snprintf` consults the locale, which serializes it** so thoroughly that parallel formatting only reached 2.3x on 16 threads. Switching to `dtoa_g_p8()` took that phase from 6.97s to 0.19s. The cost is 8 significant digits in the output rather than 12, which is well beyond what GWAS effect sizes are known to; the 8- and 12-digit outputs agree to 5e-8 over all 2M rows, exactly the rounding.

## Not here

Partitioned LD Scores, the numerical-optimization variant of Omega, `--perfect_gencov` and `--equal_h2`, and the spike-slab fitting the reference offers behind `--fit_ss`. All are additions rather than corrections to what is here.

### DIFF
--- a/2.0/utils/mtag/mtag.cc
+++ b/2.0/utils/mtag/mtag.cc
@@ -35,10 +35,10 @@
 #include <string>
 #include <unordered_map>
 
-#include "include/plink2_base.h"
-#include "include/plink2_string.h"
-#include "include/plink2_text.h"
-#include "include/plink2_thread.h"
+#include "../../include/plink2_base.h"
+#include "../../include/plink2_string.h"
+#include "../../include/plink2_text.h"
+#include "../../include/plink2_thread.h"
 
 #ifdef __cplusplus
 namespace plink2 {
--- a/2.0/Makefile
+++ b/2.0/Makefile
@@ -57,7 +57,7 @@ endif
 %.o: %.cc
 	$(CXX) -c $(CXXFLAGS) $(ARCH32) -o $@ $<
 
-all: plink2 pgen_compress
+all: plink2 pgen_compress mtag
 
 # for clean build, "make clean" first
 # Run mkdir for both plink2 and pgen_compress as we don't know which
@@ -71,6 +71,11 @@ pgen_compress: $(PGCOBJ)
 	$(MKDIR) -p bin
 	$(CXX) $(PGCOBJ) \
 		-o bin/pgen_compress
+
+# multi-trait analysis of GWAS summary statistics
+mtag: $(MTAGOBJ)
+	$(MKDIR) -p bin
+	$(CXX) $(ARCH32) $(MTAGOBJ) -o bin/mtag $(LDFLAGS)
 
 .PHONY: install-strip install clean
 
--- a/2.0/Makefile
+++ b/2.0/Makefile
@@ -57,7 +57,7 @@ endif
 %.o: %.cc
 	$(CXX) -c $(CXXFLAGS) $(ARCH32) -o $@ $<
 
-all: plink2 pgen_compress mtag
+all: plink2 pgen_compress
 
 # for clean build, "make clean" first
 # Run mkdir for both plink2 and pgen_compress as we don't know which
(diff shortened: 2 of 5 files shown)


2.0/utils/mtag/mtag.cc
```diff
@@ -0,0 +1,1910 @@
+// This file is part of PLINK 2.0, copyright (C) 2005-2026 Shaun Purcell,
```
chrchang, line 1: This should live under 2.0/utils/mtag/ , in the same way that vcf_subset.cc lives under 2.0/utils/vcf_subset/ .

BenjaminDEMAILLE: Moved to `2.0/utils/mtag/`, laid out the same way as `vcf_subset`:

- `2.0/utils/mtag/mtag.cc`, with `../../include/...` include paths.
- `2.0/utils/mtag/Makefile`, which is `vcf_subset`'s with the target renamed. mtag needs the same source set (`TEXTLIB_CCSRC`), so nothing else had to change.
- `2.0/Makefile` and `2.0/Makefile.src` are back to what they were; the PR no longer touches the main build at all.

The branch now adds exactly two files. Built it standalone from `2.0/utils/mtag/` after rebasing on current master: no warnings, and the binary runs.

Also rebased on current master while I was in here, since it was five commits behind.

PR #382, opened 2026-09-05
Add --genome (method-of-moments IBD report)

BenjaminDEMAILLE:
The IBD module is the largest remaining PLINK 1.9 command with no plink2 equivalent (issue #344, section A): `--make-king-table` reports kinship, not IBD sharing proportions, so pipelines that filter on PI_HAT still have to keep a 1.9 binary around.

```
  --genome ['zs'] [{unbounded | nudge}] ['cols='<column set descriptor>]
  --genome-min-pi-hat <val>
  --genome-max-pi-hat <val>
  --ppc-gap <kbs>
```

Columns are RT, Z0, Z1, Z2, PI_HAT, PHE, DST, PPC and RATIO by default, with NSNP, IBS0/IBS1/IBS2, HOMHOM and HETHET available. `--min`, `--max`, `--nudge`, `--unbounded` and `--genome-full` now point at their replacements rather than being unrecognized.

### Implementation

The per-pair counts this needs are the ones the KING-robust scan already produces, so the pass structure is `CalcKing()`'s dense path. Three changes make it cheaper than either program's:

- **Three counts per pair instead of five.** In the hom/ref2het encoding, nonmissing is `(hom1|ref2het1) & (hom2|ref2het2)`, IBS0 is both-hom with differing ref2het, and IBS1 is exactly-one-hom among nonmissing; IBS2 is the remainder. Three popcounts per word rather than five, and 12 bytes per pair rather than 20.
- **The IBS2\* walk is fused into that loop.** The IBS0 and het-het bitmasks the PPC test needs are exactly what the count loop already derives, so they are kept in two stack buffers and consumed immediately rather than recomputed. The per-pair cursor is an absolute variant index stored alongside the counts, which is what lets the walk cross block boundaries. The jump table (for each variant, the first variant at least `--ppc-gap` away) is copied into a block-local, block-base-relative table each block; indexing the whole-variant-set array instead cost 2.3x on a 50000-variant run, since that array does not fit in L1.
- **Triangle passes.** A run that does not fit in memory degrades to a second read of the variants instead of failing; PLINK 1.9 holds the whole table and has no fallback.

2000 samples x 50000 variants over 22 chromosomes at 50 kb spacing, so the IBS2\* walk has real work to do. Minimum of 8 runs; the machine had other load at the time, so these are minima rather than means:

| | PLINK 1.9 | plink2 | |
|---|---:|---:|---|
| 16 threads | 4.92s | 2.41s | 2.0x |
| `--threads 1` | 39.77s | 30.19s | 1.3x |

(report suppressed with a PI_HAT filter, so this is scan time.)

### Verification

Field by field against PLINK 1.9.

- 200-sample, 240-variant set with 50 four-member families and a case/control phenotype: all 19900 pairs agree on RT, PHE, Z0, Z1, Z2, PI_HAT and DST to 1.9's four-decimal print precision, and IBS0/IBS1/IBS2 agree exactly. Same under `unbounded`, `nudge`, `--nonfounders` and both PI_HAT filters.
- 200-sample, 20000-variant, 5-chromosome set spanning 13 variant blocks: PPC and RATIO agree on all 19900 pairs at `--ppc-gap` 0, 2, 10 and 500.
- 2000-sample dummy fileset: all 1999000 pairs agree.
- A 12000-sample run that takes two (and, with the PPC fields, three) passes produces output identical to the single-pass run, and `--parallel 3` concatenates to the whole report. Those are what exercise the cross-block cursor.

`Tests/TEST_GENOME` covers all of the above.

### Notes

PLINK 1.x's EZ column (pedigree-based expected relatedness) is not implemented; it needs the pedigree relationship matrix, and RT already covers the .fam-derived relationship. Called out in the help text.

PPC needs every variant in genomic order, so a position-sorted .pvar/.bim is required when the ppc or ratio columns are on; the error message points at `--sort-vars` and at dropping the columns.

One numeric note worth flagging: plink2's allele frequencies are computed by multiplying by a reciprocal, so a monomorphic variant's major-allele frequency can come back as `1 - 1ulp`. PLINK 1.x's expectation accumulator tests `p > 0 && q > 0`; this one compares against half an allele copy instead, which rejects exactly the variants that carry no information. Without that, 37 of 240 monomorphic variants in my first test set were being folded into the expectations and every Z estimate was off.


chrchang:
The main question here is whether the Z0/Z1/Z2 estimates are still useful intermediate statistics for some modern IBD methods.  I have not looked at this recently.

BenjaminDEMAILLE:
My honest read, flagged as judgment rather than something I can put a usage number on:

**Not as an intermediate.** No modern IBD method I am aware of consumes method-of-moments (Z0, Z1, Z2) as an input. The segment-based callers (Refined IBD, hap-IBD, TRUFFLE, IBIS, and KING's own IBD segment mode) all go from genotypes or phased haplotypes straight to segments, and if they report genome-wide sharing proportions at all they derive them from the segments afterward. In that direction the statistic is a dead end.

**Still used as an endpoint, for one thing kinship alone cannot do.** The (Z0, Z1, Z2) triple is still the standard way to *name* a relationship, because kinship is degenerate across the two relationship types people most need to tell apart: parent-offspring and full siblings both have kinship 0.25, and are separated only by Z0 (0 vs 0.25) and Z2 (0 vs 0.25). That distinction is what pedigree-QC steps in GWAS pipelines actually key on, and the PI_HAT thresholds from the Anderson et al. QC protocol are still widely copied.

plink2 is not missing the *information* here: `--make-king-table` reports IBS0, and IBS0 near zero versus clearly positive separates PO from FS exactly as Z0 does. What it is missing is the parameterization people already know how to read, and the arithmetic to get from IBS0 to it is not something a user will derive on the spot.

So the case for `--genome` is "the conventional output format for a question plink2 can already answer", not "an input to anything modern". If that is not enough to justify a command, two smaller options that would still close the gap:

1. Add Z0/Z1/Z2/PI_HAT as optional `--make-king-table` column sets, which is where the pairwise scan already lives, and retire `--genome` with a pointer to them. The kernel in this PR would move there mostly unchanged.
2. Keep only the relationship-type call rather than the raw triple, i.e. a column that says PO/FS/HS/UN, which is what the thresholds are used for anyway.

I would lean toward (1) if you want it in plink2 at all, since it avoids a second pairwise command with its own output format, and it makes the "you already have this via KING" relationship explicit instead of leaving users to rediscover it. Happy to restructure this PR that way. If you would rather not carry the statistic at all, that is a defensible call too and I will close it.

chrchang:
Leaning toward:
* No more --genome command.  --genome's error message points to --make-king-table.
* --make-king-table gains optional RT ('rt') and EXPECTED_KINSHIP ('ekin') columns.
* Documentation includes a note about how parent/offspring pairs should be distinguishable from sibling pairs via the ibs0 column: ibs0 should be much closer to zero for parent/offspring pairs.

BenjaminDEMAILLE:
Works for me, and it is the better shape -- one relatedness table rather than two commands computing overlapping things. Plan:

1. Drop `--genome` and everything under it, including `--ppc-gap` and `Tests/TEST_GENOME`.
2. `--genome` joins the PLINK 1.x hint table, pointing at `--make-king-table`.
3. `--make-king-table` gains `rt` and `ekin` column options.
4. Help gains the note that `ibs0` separates parent/offspring from full siblings, since it is near zero for the former.

Two things worth settling before I write it, since they are cheaper to fix now than after.

**Plumbing.** `CalcKing()` currently takes `SampleIdInfo`, which has FID/IID/SID but not parental IDs or founder status. Both new columns need those, so I would pass `PedigreeIdInfo` through instead. That is the only structural change; say the word if you would rather they came in some other way.

**What `ekin` holds.** My reading is a pedigree-expected *kinship coefficient* -- 0.25 for parent/offspring and full siblings, 0.125 for half siblings, grandparent/grandchild and avuncular, 0 for unrelated -- so it sits next to the observed `kinship` column and the two are directly comparable. That differs from 1.9's `EZ`, which is an expected IBD-sharing category rather than a scalar. If you meant the latter, tell me and I will match it instead.

Related: `rt` can only say what the pedigree says. With FID and parental IDs alone, parent/offspring and full siblings are indistinguishable, which is exactly why the `ibs0` note earns its place. I will make the help say that rather than leave users to discover it.

chrchang:
"Plumbing" and "What `ekin` holds" are as I expect.

BenjaminDEMAILLE:
Closing per your note on #400.

The direction you settled on there supersedes this: no `--genome` command, its error message points at `--make-king-table`, and the pedigree information people used `--genome` for arrives as the optional `rt` and `pkin` columns instead. #400 also documents that parent/offspring pairs are separable from full siblings via `ibs0`, which is the discrimination the Z0/Z1/Z2 estimates were being used for here.

Thanks for thinking it through rather than just taking the port.

PR #384, opened 2026-09-05
Add --mgf (BIMBAM mean genotype import)

BenjaminDEMAILLE:
Follows up on #369, where you wrote that `--mgf` can be implemented in a separate PR. **Stacked on #369**: the first commit here belongs to that branch, so this diff only reads cleanly once #369 lands.

BIMBAM's mean genotype format is one of the formats PLINK 1.9 could write but never read, so a dataset exported to it could not be brought back. `--export mgf` is in #369; this is the reader.

```
--mgf <.mgf file> <.pos.txt file> ['pheno='<.pheno.txt file>]
```

The .mgf has one row per variant (ID, ALT, REF, then one ALT dosage per sample). The .pos.txt gives each variant's ID, bp coordinate and chromosome in any order, so it is read into a hash table first and the .mgf is then streamed in a single pass. Commas are accepted in place of whitespace in both, since that is what BIMBAM itself writes.

Neither file carries sample IDs, so `per0`, `per1`, ... are synthesized as `--dummy` does. `pheno=` names a file with one row per sample whose columns become PHENO1, PHENO2, ...; its row count is checked against the sample count the .mgf implies.

Dosages are imported at full precision and reduced to hardcalls by the usual `--hard-call-threshold` / `--dosage-erase-threshold` rules.

One judgment call worth flagging: `NA`, `-9`, `?`, `??` and `.` mark a missing call, and anything else that is not a number in [0, 2] is an error rather than a silent missing call. Treating an unparseable token as missing would quietly drop data on a malformed file, which seemed worse than refusing it. Easy to relax if you would rather match BIMBAM's own leniency.

`Tests/TEST_MGF` covers the `--export mgf` round trip on hardcalls and on dosages (byte-identical .mgf, .pos.txt and .pvar both ways, and the synthesized .psam), the comma-delimited and bgzipped forms, each missing-value spelling, `pheno=`, and six malformed-input cases including the line numbers the errors cite.

The first commit removes a `ChrInfo` parameter that `ExportBeagle` stopped using when the chromosome-split form was retired in #369; it was tripping `-Wunused-parameter` along with two dead locals.

### DIFF
--- a/2.0/Tests/TEST_MGF/.gitignore
+++ b/2.0/Tests/TEST_MGF/.gitignore
@@ -1,0 +1,4 @@
+plink19*
+plink2*
+tmp_*
+*.log
--- a/2.0/Tests/TEST_MGF/run_tests.sh
+++ b/2.0/Tests/TEST_MGF/run_tests.sh
@@ -1,0 +1,112 @@
+#!/bin/bash
+
+# "--mgf", the BIMBAM mean genotype reader, checked as the inverse of
+# "--export mgf".
+
+set -exo pipefail
+
+BUILD=$1
+EXTRA1=$2
+EXTRA2=$3
+
+# 1. Hardcalls.  A round trip through the text format has to reproduce the
+#    .mgf and .pos.txt byte for byte, and the .pvar apart from its header.
+plink --simulate simulate.txt --simulate-missing 0.03 --simulate-ncases 60 --simulate-ncontrols 60 --out tmp_hard > /dev/null
+$BUILD/plink2 $EXTRA1 $EXTRA2 --bfile tmp_hard --export mgf --out tmp_hard_e
+$BUILD/plink2 $EXTRA1 $EXTRA2 --mgf tmp_hard_e.mgf tmp_hard_e.pos.txt --out tmp_hard_r
+$BUILD/plink2 $EXTRA1 $EXTRA2 --pfile tmp_hard_r --export mgf --out tmp_hard_e2
+diff -q tmp_hard_e.mgf tmp_hard_e2.mgf
+diff -q tmp_hard_e.pos.txt tmp_hard_e2.pos.txt
+
+# The .pvar has to name the same variants at the same positions, with REF and
+# ALT on the same sides: the .mgf's dosages count its second column, which is
+# ALT, so a swap here would silently invert every call.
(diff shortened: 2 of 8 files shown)


chrchang:
This should optionally work with --psam/--fam, like --vcf does.

BenjaminDEMAILLE:
Done, pushed.

`--psam` (and its `--fam` spelling) now supplies real sample IDs, sex and phenotypes, through the same `RewritePsam()` path the other legacy importers use:

```
plink2 --mgf data.mgf data.pos.txt --psam real.psam --out out
```

The semantics differ slightly from `--vcf`'s, which is worth calling out: a VCF header has sample IDs to match the .psam against, and a .mgf has none, so there is nothing to match on. The file just has to have one row per sample, in the .mgf's column order. A count that disagrees with what the .mgf implies is an error naming both:

```
Error: short.psam has 3 samples, while e.mgf implies 6.
```

`--psam`/`--fam` and `pheno=` are mutually exclusive, since both supply phenotypes.

`Tests/TEST_MGF` covers both spellings (checking that `--fam` brings FID through and `--psam` does not invent one), that the genotypes round-trip unchanged when IDs come from a file, and the two error cases.

chrchang:
A .psam file is not required to contain phenotype information, and even a .fam file’s phenotype column may be all -9.  So I think it is reasonable to accept the .psam + pheno= combination, and then error out if the .psam/.fam contains any actual phenotype information.

BenjaminDEMAILLE:
Right, that was wrong of me — I conflated "has a phenotype column" with "carries phenotype data". Fixed and pushed.

`--psam`/`--fam` and `pheno=` are now accepted together, and the error fires only when the sample file actually holds a value for somebody:

```
Error: real.psam contains phenotype data, so it cannot be combined with
--mgf 'pheno='.
```

`RewritePsam()` gained an optional out-parameter reporting whether any phenotype column had a non-missing entry, which is the question that actually matters. When the answer is no, the sample IDs and sex come from the .psam/.fam and the phenotypes from the `pheno=` file, appended to the rewritten .psam as PHENO1, PHENO2, ...

Both of the cases you named are covered in `Tests/TEST_MGF`: a .psam with only `#IID`/`SEX`, and a .fam whose sixth column is entirely `-9`. Both combine cleanly, and the test checks that the IDs and sex come from the sample file while the phenotypes come from `pheno=`. A .psam that does carry phenotypes still errors.

Issue #311, opened 2025-10-30
Feature request: hamming distance and MDS in plink2


janxkoci:
Can we please have the routines for hamming distance and MDS in plink2? 

Of course, I can run both with plink 1.9, but plink2 has better output formats (tabs ftw, headers) and it's tiring to keep mentally track of which results need to be parsed differently when summarising etc. 

Thanks for this awesome tool btw, it really rocks 😎

chrchang:
Development build with 'ibs' --make-king-table column-set has been posted.

PR #385, opened 2026-09-05
Add --distance, --distance-matrix and --ibs-matrix

BenjaminDEMAILLE:
Part of #311 and #344: plink2 has `--make-rel`/`--make-king` for relatedness but no IBS or Hamming distance matrix, so there is no path from a plink2 fileset to the distance inputs that MDS and clustering tools expect.

```
--distance [{square | square0 | triangle}] [{zs | bin | bin4}] ['ibs']
           ['1-ibs'] ['allele-ct'] ['flat-missing']
--distance-matrix
--ibs-matrix
```

**Kernel.** This is the KING-robust dense scan with a smaller inner loop: the allele-count distance is `2 * IBS0 + IBS1`, which in the hom/ref2het encoding is two popcounts, plus the nonmissing count that `flat-missing` divides by. The three reports are the same number in different units (IBS is `1 - dist/(2m)`, `.mdist` is its complement), so one pass produces all of them.

**The frequency-weighted default** needs, per pair, the total weight of the variants missing in *both* samples. That term comes out of the same loop for free: the intersection of the two missing masks is the complement of the union of the called masks, which the kernel already computes. Almost every word pair intersects empty, so it costs one AND and one test per word instead of PLINK 1.9's per-pair table lookup, and the pair table is only written when the block contributed something, which matters because it is far too large to cache.

On 4000 samples x 50000 variants, best of three:

| | 1.9 | plink2 | |
|---|---:|---:|---:|
| `--distance`, 16 threads | 6.85s | 2.68s | 2.6x |
| `--distance flat-missing`, 16 threads | 6.20s | 2.14s | 2.9x |
| `--distance`, 1 thread | 10.28s | 9.58s | 1.07x |
| `--distance flat-missing`, 1 thread | 9.59s | 5.96s | 1.6x |

**Equivalence.** `Tests/TEST_DISTANCE` compares every shape and encoding against PLINK 1.9. `flat-missing` output is byte-for-byte identical, including the `.bin` matrix and each `--parallel` piece; the weighted default agrees to within 1.9's weight quantization (it rounds the per-variant weights to integers summing to just under 2^32, plink2 keeps doubles), max observed relative difference 1.6e-9, so those comparisons carry a 1e-6 tolerance.

**Two PLINK 1.9 bugs found on the way.** The test suite pins both, and will fail loudly if 1.9 changes.

1. In `calc_distance()`, the missing-observation weight loop is

   ```c
   dxx = set_allele_freqs[marker_uidx];
   if ((dxx != 0.0) && (dxx != 1.0)) {
     dxx = dxx * (1.0 - dxx) * (dxx * dxx - dxx + 1);
     ...
   }
   dist_missing_wts[marker_idx] = dxx;
   ```

   When the set allele frequency is exactly 1, `dxx` is left at 1.0 rather than set to 0, so a monomorphic variant gets weight 1 -- more than five times the largest weight any polymorphic variant can reach (0.1875, at MAF 0.5) -- and dominates the missingness correction. The comment two lines above says the opposite is intended ("a missing observation at a MAF-0 marker has no weight at all"). Minimal case, four samples, one monomorphic variant, one missing call:

   ```
   1.9      2.27433  2.27433  3.4115
   plink2   3.02196  3.02196  4.53293
   ```

   Solving for the implied total weight puts 1.9's exactly 1.0 above plink2's, which is the monomorphic variant's weight. Dropping that variant makes the two byte-identical. plink2 gives it weight 0.

2. In `distance_d_write()`, the single-precision square writer sets `fxx = 0.0` once before the loop and then reuses `fxx` for the lower-triangle values, so the diagonal write emits the value at `(i, i-1)` rather than 0. On a 120-sample fileset, 79 of the 120 diagonal entries are wrong. The double-precision path is fine. plink2 writes 0.

Happy to drop either divergence and match 1.9 exactly if you would rather keep bit-compatibility.

**Not included:** `--distance-wts`/`--exponent` variant weighting, which changes the distance accumulation itself and needs the weighted-table kernel rather than popcounts. Also `--read-dists` and the distance regressions (`--ibs-test`, `--groupdist`, `--regress-distance`), which are separate commands in the same #344 row.

**Overlap note:** the kernel here is close to the one in #382 (`--genome`), which counts nonmissing/IBS0/IBS1 from the same bitplanes. They are independent branches on master so either can land first; once one does I am happy to factor the shared loop out in a follow-up rather than leave two copies.


chrchang:
* The 32-bit quantization of PLINK 1.9 --distance's default missingness correction saved a bit of memory in its day, but I don't think it's worth carrying forward.  Just use float64 arithmetic.

Yes, this does complicate testing; it's ok if this PR's scope is reduced to just the flat-missing case.

BenjaminDEMAILLE:
Good news: that is already the case. The weights never get quantized here.

```c
const double qq = GetAlleleFreq(&(allele_freqs[allele_idx_base]), 0, cur_allele_ct);
const double cur_weight = qq * (1.0 - qq) * (qq * qq - qq + 1.0);
variant_weights[vidx] = cur_weight;
wsum += cur_weight;
```

`variant_weights`, `wmiss`, `wsum` and `wboth` are all `double` end to end. The only `uint32_t` array is `counts`, and that holds genuine integer allele/IBS counts rather than scaled weights.

The test header already says as much, and the weighted comparison against 1.9 carries a tolerance for exactly the reason you give:

```
# The default frequency-weighted mode agrees to within PLINK 1.9's weight
# quantization (it rounds the per-variant weights to integers summing to just
# under 2^32; plink2 keeps them as doubles), so those comparisons carry a
# tolerance.
```

`close plink19_w.dist plink2_w.dist 1e-6`, and the suite asserts the weighting actually changed something, so the tolerance is not hiding a no-op.

I suspect the comment on `DistanceCellVal()` is what read as carrying 1.9's arithmetic forward:

```c
// Grouped the way PLINK 1.9 groups it, so that the two programs round
// identically when the weights agree.
return (variant_ctd / u31tod(counts[...])) * rawd;
```

That is only about the *order* of the final divide and multiply, not about quantization. It is what makes `flat-missing` byte-identical to 1.9 rather than tolerance-compared -- 18 of the suite's comparisons are exact `cmp`s that depend on it.

So: happy to regroup that as `rawd * variant_ctd / denom` if you would rather the arithmetic not be shaped by 1.9 at all. The cost is that `flat-missing` becomes another tolerance comparison and I lose the exactness check, which is the strongest evidence in that suite. Your call; it is a two-line change either way.

On scope: since the weighted mode is implemented and tested, I would rather keep it than cut the PR back to `flat-missing`. Say the word if you disagree and I will drop it.

chrchang:
Nice, sorry about the false alarm.  No need to regroup that one expression.

2.0/plink2_matrix_calc.cc
```diff
@@ -4552,6 +4552,642 @@ PglErr CalcMissingMatrix(const uintptr_t* sample_include, const uint32_t* sample
...
+      // 24 bytes is enough for a %g double plus its delimiter.
+      const uintptr_t overflow_buf_size = kCompressStreamBlock + 24 * S_CAST(uintptr_t, sample_ct) + 64;
+      reterr = InitCstreamAlloc(outname, 0, output_zst, max_thread_ct, overflow_buf_size, &css, &cswritep);
+      if (unlikely(reterr)) {
+        goto WriteDistanceMatrix_ret_1;
+      }
+      // PLINK 1.9's --distance-matrix/--ibs-matrix wrote space-delimited
+      // matrices with a trailing delimiter; --distance itself is
+      // tab-delimited with none.
+      const uint32_t space_delim = (flags / kfDistanceSpaceDelim) & 1;
```
chrchang: PLINK 2.0 sticks to tab output just about everywhere it isn't explicitly producing an interchange format.  The output of --distance-matrix/--ibs-matrix doesn't rise to that bar (and on top of that, those flags were already marked as deprecated), so let's just change this bit of behavior.  Documentation will be updated accordingly.

BenjaminDEMAILLE: Done. `kfDistanceSpaceDelim` is gone rather than defaulted off, since nothing else wanted it, so both flags now write tabs with no trailing delimiter like every other plink2 table.

Useful side effect: each is now byte-identical to its long form, which is a better invariant than 1.9's spacing, so the suite pins that directly:

```
cmp plink2_dm.mdist plink2_dm_long.mdist   # --distance-matrix == --distance 1-ibs flat-missing square
cmp plink2_im.mibs  plink2_im_long.mibs    # --ibs-matrix      == --distance ibs flat-missing square
```

The 1.9 comparison stays, demoted to a value check with its separator and trailing delimiter normalized away first, plus an explicit assertion that the output has tabs, no spaces and no trailing delimiter.

Help updated: the two are now described as exactly equivalent to their long forms, with a note that PLINK 1.x wrote them space-delimited and plink2 does not.

Note the `--distance ibs`/`1-ibs` comparisons in section 2 were already exact and still are -- 1.9 only used spaces for `--distance-matrix`/`--ibs-matrix`, not for `--distance` itself.

chrchang:
oops, typed a period instead of a comma when resolving merge.  Will fix this in immediately-following commit.

chrchang:
`grep -q '\t'` and `grep -qv '\t$'` do not work as intended on Linux (though they work on my Mac).  Will replace with something that works on both macOS and Linux.

PR #386, opened 2026-09-05
Add --model and --cell (contingency-table association tests)

BenjaminDEMAILLE:
Section B of #344. `--glm` covers `--assoc`/`--linear`/`--logistic`, but not the contingency-table tests `--model` provides, and it has no exact test at all — so this is the part of the gap most often assumed to be already covered.

```
--model [{fisher | fisher-midp}] ['trend-only'] ['zs'] ['cols='<descriptor>]
--cell <thresh>
```

One 2x3 table per variant (case/control by genotype) yields five tests, one row each: GENO (2df chi-square on the full table), TREND (Cochran-Armitage), ALLELIC (1df on the collapsed allele counts), DOM and REC. `fisher` swaps the chi-square tests for exact ones.

**New statistics in `include/plink2_stats`:** `Chi22Stat`, `Chi23Stat`, `CaTrendStat`, and `Fisher23TwoSidedP`. The last is a port of PLINK 1.9's `fisher23()` — the 2-dimensional SNPHWE2-style walk, not the network algorithm — carried over with its constants, so the two agree bit-for-bit. I verified that rather than assuming it: a standalone harness ran both implementations over 54000 random 2x3 tables (small, degenerate with zero cells and zero rows, and near-balanced with six-figure counts), each with and without mid-p, and found **zero differences in the last bit**.

I kept `EXACT_TEST_BIAS` and `FISHER_EPSILON` rather than re-deriving them; both already match plink2's `kExactTestBias` (2^-83) and `kSmallEpsilon` (2^-44) exactly, so nothing new was introduced.

**Performance.** A single exact test can cost more than reading the variant it describes, and the cost varies by orders of magnitude between variants, so the per-variant tests run on worker threads with variants handed out round-robin rather than in contiguous chunks — a contiguous split leaves threads idle waiting on whichever chunk drew the expensive tables. 4000 samples x 200000 variants, best of three:

| | 1.9 | plink2 | |
|---|---:|---:|---:|
| `--model`, 16 threads | 0.54s | 0.14s | 3.9x |
| `--model fisher`, 16 threads | 2.17s | 0.26s | 8.3x |
| `--model fisher`, 1 thread | 2.17s | 1.44s | 1.5x |

**Equivalence.** `Tests/TEST_MODEL` compares both report kinds against PLINK 1.9 at three `--cell` thresholds, plus `trend-only`, the column sets, `zs` and `--threads 1`, and checks the inputs that must be rejected. Equivalence was also confirmed on 1,000,000 report rows from the benchmark fileset.

**Two deliberate differences from PLINK 1.9**, both flagged in the help text:

1. A1 is the ALT allele, as everywhere else in plink2, so DOM and REC are relative to ALT; PLINK 1.x keyed them on the minor allele, which silently swaps the two tests whenever ALT is the major allele. `--maj-ref` makes ALT the minor allele for anyone who wants the old convention, and that is how the test suite builds its comparison fileset.
2. `--model`'s `dom`/`rec`/`gen`/`trend` and the retired `--model-dom` family only select which test PLINK 1.x's *permutation* test is based on — the report itself is identical either way. Since permutation is not implemented here, they are rejected with a message saying so rather than silently ignored. Same for `perm`/`mperm=`/`set-test`.

**Not included:** the permutation framework (`--assoc perm`, `--model mperm=`, `--aperm`), which is its own row in #344 and would want a shared implementation across commands rather than one bolted onto `--model`.


chrchang:
This was one of the association tests intentionally not implemented in PLINK 2.0.  The no-covariates drawback is just too heavy for this to be worth carrying forward, given that --glm has asymptotic agreement anyway.

BenjaminDEMAILLE:
Understood, closing.

One piece is reusable independently of the decision, so flagging it before it disappears with the branch: `Fisher23TwoSidedP` in `include/plink2_stats`, a 2x3 Fisher-Freeman-Halton exact test. It is a port of PLINK 1.9's `fisher23()` with the same constants, and I checked it over 54000 random 2x3 tables (small, degenerate with zero cells and zero rows, and near-balanced with six-figure counts, each with and without mid-p) against 1.9's implementation: **zero differences in the last bit**. It sits next to the existing `Fisher22TwoSidedP` and needs nothing else from this branch.

Nothing in plink2 currently calls it, so it is only worth keeping if you expect a use. Say the word and I will send it as a standalone; otherwise it goes away with this PR and there is no need to reply.

PR #387, opened 2026-09-05
1.9: --bmerge duplicate-ID reporting (issue #140)

### LINKED ISSUE #140: Logging bug in bmerge
[See two small datasets attached](https://github.com/chrchang/plink-ng/files/4561217/Report.zip). I'm running merge-mode 6 and would thus expect the ordering of the datasets not to matter but I get different results as highlighted in bold

>PLINK v1.90b6.17 64-bit (28 Apr 2020)          www.cog-genomics.org/plink/1.9/
(C) 2005-2020 Shaun Purcell, Christopher Chang   GNU General Public License v3
Logging to PlusFirst.log.
Options in effect:
  --bfile Plus
  --bmerge Normed.bed Normed.bim Normed.fam
  --merge-mode 6
  --out PlusFirst
16340 MB RAM detected; reserving 8170 MB for main workspace.
4 people loaded from Plus.fam.
4 people to be merged from Normed.fam.
Of these, 0 are new, while 4 are present in the base dataset.
Warning: Multiple positions seen for variant 'MNV'.
Warning: Multiple chromosomes seen for variant 'PAR-X'.
30 markers loaded from Plus.bim.
**29 markers to be merged from Normed.bim**.
Of these, 0 are new, while 29 are present in the base dataset.
Warning: Variants '0monomorphic' and '00missing' have the same position.
Warning: Variants '2:103037578' and '0monomorphic' have the same position.
Warning: Variants '2:103037578:G:T' and '2:103037578' have the same position.
24 more same-position warnings: see log file.
Performing 1-pass diff (mode 6), writing results to PlusFirst.diff .
**116 overlapping calls, 84 nonmissing in both filesets.**
**76 concordant, for a concordance rate of 0.904762.**

>PLINK v1.90b6.17 64-bit (28 Apr 2020)          www.cog-genomics.org/plink/1.9/
(C) 2005-2020 Shaun Purcell, Christopher Chang   GNU General Public License v3
Logging to NormedFirst.log.
Options in effect:
  --bfile Normed
  --bmerge Plus.bed Plus.bim Plus.fam
  --merge-mode 6
  --out NormedFirst
16340 MB RAM detected; reserving 8170 MB for main workspace.
4 people loaded from Normed.fam.
4 people to be merged from Plus.fam.
Of these, 0 are new, while 4 are present in the base dataset.
Warning: Multiple positions seen for variant 'MNV'.
Warning: Multiple chromosomes seen for variant 'PAR-X'.
**27 markers loaded from Normed.bim.**
30 markers to be merged from Plus.bim.
Of these, 3 are new, while 27 are present in the base dataset.
Warning: Variants '0monomorphic' and '00missing' have the same position.
Warning: Variants '2:103037578' and '0monomorphic' have the same position.
Warning: Variants '2:103037578:G:T' and '2:103037578' have the same position.
23 more same-position warnings: see log file.
Performing 1-pass diff (mode 6), writing results to NormedFirst.diff .
**108 overlapping calls, 78 nonmissing in both filesets.**
**70 concordant, for a concordance rate of 0.897436.**

cindykrafft:
Fixes #140.

`--bmerge` keys the merge on variant ID, so several records that share an ID are one
variant in the merged fileset. The log counted *records* for the second fileset and
*distinct IDs* for the base, so the same fileset was described two different ways
depending on which side of the merge it was on, and `Of these, X are new, while Y are
present in the base dataset` could report more variants present in the base than the base
dataset has. The duplicate-ID warning was also restricted to the first fileset — the
asymmetry the comment at that site describes, and the todo it leaves — so in the
reporter's case the fileset that actually had the duplicate IDs was accepted silently
when it happened to be the second one.

On the reporter's data, `Normed.bim` was "27 markers loaded" as the base and "29 markers
to be merged" as the second fileset. On the four-variant reproduction in the new test
(`plus`: 4 records / 4 IDs, `normed`: 4 records / 3 IDs):

```
                        before                       after
base = plus     (no duplicate-ID warning)    Warning: normed.bim contains duplicate ...
                4 markers to be merged       3 markers to be merged from normed.bim.
                  from normed.bim.
                Of these, 0 are new,         Of these, 0 are new, while 3 are present
                  while 4 are present ...      ...
base = normed   3 markers loaded from        3 markers loaded from normed.bim.
                  normed.bim.
```

Cause and fix: `merge_bim_scan()` could not tell a duplicate ID *within* the fileset it is
scanning from an ID that also appears in an earlier fileset, since both are hash-table
hits. Each `Ll_bim` entry now carries the index of the last fileset it was seen in
(`last_fileset_idx`, in the same spirit as `Ll_fam::orig_order`), so the scan counts
within-fileset duplicates and returns that count; `merge_datasets()` prints the warning for
whichever fileset has them, naming it, and uses `cur_marker_ct - cur_dup_ct` for the second
fileset's variant count. The entry gains 4 bytes, which the 16-byte end-allocation rounding
usually absorbs. Genotype output is unchanged: both merge directions still produce the same
4-variant fileset.

Not addressed here, since it is a semantic question rather than a reporting one: with
`--merge-mode 6` each *record* of the second fileset is still diffed against the base, so
the overlapping-call and concordance counts still depend on merge order when one fileset
has duplicate IDs (12 calls / 0.9167 vs 9 calls / 0.8889 on the reproduction above; 116 /
0.9048 vs 108 / 0.8975 in the report). Happy to follow up on that separately if you want it
changed.

Test: `2.0/Tests/TEST_BMERGE_DUP_IDS` (registered in `2.0/Tests/run_tests.sh`, so it runs in
the PLINK2 functional tests workflow). It builds the two small filesets above with plink
1.9 and asserts the warning and both filesets' counts in each merge direction, that the
merged fileset has 4 variants either way, and that a duplicate-free pair produces no
warning. It fails on master at the first assertion and passes with this patch.

What was run, on Linux/gcc 12, with 1.9 built `make ZLIB=-lz BLASFLAGS=<local OpenBLAS>` and
plink2 from `2.0/build_dynamic`:

* `2.0/Tests/run_tests.sh`: 9/9 before (master, without the new test), 10/10 after
  (including `TEST_BMERGE_DUP_IDS`).
* The new test alone against a master build of 1.9: fails; against the patched build:
  passes.
* Build clean, no new compiler warnings.

There is no `--bmerge` unit-test target in `1.9/tests/` (it compares against a PLINK 1.07
build), so the test went in the functional-test suite that CI already runs.

Found while working through open issues in Mytochondria, a volunteer project that verifies fixes for the software behind published results (methods and harnesses: https://github.com/cindykrafft/mytochondria/tree/main/audits/plink)

---
_Generated by [Claude Code](https://claude.ai/code)_

### DIFF
--- a/1.9/plink_data.c
+++ b/1.9/plink_data.c
@@ -14501,6 +14501,10 @@ typedef struct ll_entry2_struct {
   int64_t pos;
   double cm;
   char* allele[2];
+  // index of the last fileset this variant ID was seen in; lets
+  // merge_bim_scan() tell a duplicate ID within the current fileset apart from
+  // an ID that also appears in an earlier fileset
+  uint32_t last_fileset_idx;
   char idstr[];
 } Ll_bim;
 
@@ -14745,7 +14749,7 @@ int32_t merge_sample_sortf(char* sample_sort_fname, char* sample_fids, uintptr_t
   return retval;
 }
 
-int32_t merge_bim_scan(char* bimname, uint32_t is_binary, uint32_t allow_no_variants, uintptr_t* max_marker_id_len_ptr, Ll_bim** htable_bim, uint32_t* max_bim_linelen_ptr, uint64_t* tot_marker_ct_ptr, uint32_t* cur_marker_ct_ptr, uint64_t* position_warning_ct_ptr, Ll_str** non_biallelics_ptr, uint32_t allow_extra_chroms, Chrom_info* chrom_info_ptr) {
+int32_t merge_bim_scan(char* bimname, uint32_t is_binary, uint32_t allow_no_variants, uintptr_t* max_marker_id_len_ptr, Ll_bim** htable_bim, uint32_t* max_bim_linelen_ptr, uint64_t* tot_marker_ct_ptr, uint32_t* cur_marker_ct_ptr, uint32_t* cur_dup_ct_ptr, uint64_t* position_warning_ct_ptr, Ll_str** non_biallelics_ptr, uint32_t allow_extra_chroms, uint32_t fileset_idx, Chrom_info* chrom_info_ptr) {
   unsigned char* bigstack_mark = g_bigstack_base;
   uintptr_t max_marker_id_len = *max_marker_id_len_ptr;
   uintptr_t loadbuf_size = MAXLINELEN;
@@ -14753,6 +14757,7 @@ int32_t merge_bim_scan(char* bimname, uint32_t is_binary, uint32_t allow_no_vari
   uint64_t tot_marker_ct = *tot_marker_ct_ptr;
   uint64_t position_warning_ct = *position_warning_ct_ptr;
   uint32_t cur_marker_ct = 0;
+  uint32_t cur_dup_ct = 0;
   double cm = 0.0;
   FILE* infile = nullptr;
   int32_t retval = 0;
@@ -14800,6 +14805,7 @@ int32_t merge_bim_scan(char* bimname, uint32_t is_binary, uint32_t allow_no_vari
     if (!line_idx) {
       // no variants
       *cur_marker_ct_ptr = 0;
+      *cur_dup_ct_ptr = 0;
       goto merge_bim_scan_ret_1;
     }
     line_idx--;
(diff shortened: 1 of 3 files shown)


cindykrafft:
never mind, closing this after reviewing the issue comments.

PR #389, opened 2026-09-05
--flip-scan: compare against a second fileset

BenjaminDEMAILLE:
Follow-up to #355 (`--flip-scan-ref-freq`), as mentioned there. Stacked on that branch, so review it after #355.

`--flip-scan-ref-freq` replaces the case/control split with a comparison against a reference allele frequency file. This adds `--flip-scan-ref-pfile` / `--flip-scan-ref-bfile`, which does the same comparison against a second fileset and computes its allele frequencies here, so no separate `--freq` run is needed.

Naming the fileset rather than a frequency table also lets the allele reconciliation happen inside the command, which is the part a frequency file cannot do. Variants are matched by ID, and both sides have to be biallelic with the same allele pair; a pair in the opposite order is counted, reported in the log, and its frequency turned around before the comparison, since that swap is one of the things this command exists to find rather than a reason to reject the variant. A variant that is unmatched, or that matches by ID but not by allele pair, gets `PROBLEM=NA` and is counted in a warning.

The argument shape follows `--pgen-diff`: either one prefix or all three filenames. The mode is mutually exclusive with `--flip-scan-ref-freq`.

```
--flip-scan: 130 variants matched the reference fileset by ID and allele pair
(43 with REF and ALT the other way round).
--flip-scan: 0 problem variants.
```

What this does **not** do is run the LD half of the scan across the two filesets. That needs the neighbor correlations computed on each fileset separately and then compared, which is a different report shape from the current one, so I left it out rather than guess at it. Say the word if you want it and I will add it.

Tests in `2.0/Tests/TEST_FLIP_SCAN` cover the three spellings of the argument, agreement with the frequency-file mode when the reference is the same dataset, the REF/ALT-swapped case (built with `--ref-allele` so the labels and the genotypes stay consistent), unmatched and allele-mismatched variants, a genuine frequency difference, and the column renaming under `ref-allele-based`.

### DIFF
--- a/2.0/Tests/TEST_FLIP_SCAN/run_tests.sh
+++ b/2.0/Tests/TEST_FLIP_SCAN/run_tests.sh
@@ -159,3 +159,73 @@ if $BUILD/plink2 $EXTRA1 $EXTRA2 --bfile tmp_data --flip-scan-ref-freq tmp_panel
     echo "--flip-scan-ref-freq ran without --flip-scan"
     exit 1
 fi
+
+# 8. --flip-scan-ref-pfile/--flip-scan-ref-bfile: same comparison against a
+#    second fileset, whose allele frequencies are computed here.
+$BUILD/plink2 $EXTRA1 $EXTRA2 --bfile tmp_data --maf 0.05 --make-bed --out tmp_ref
+$BUILD/plink2 $EXTRA1 $EXTRA2 --bfile tmp_ref --make-pgen --out tmp_refp
+$BUILD/plink2 $EXTRA1 $EXTRA2 --bfile tmp_data --maf 0.05 --flip-scan --flip-scan-ref-bfile tmp_ref --out plink2_rd
+head -n 1 plink2_rd.flipscan | grep -qx '#CHROM	POS	ID	REF	ALT	MAJ_FREQ	PANEL_MAJ_FREQ	PROBLEM'
+# The reference is this dataset, so it has to agree with the --flip-scan-ref-freq
+# run against this dataset's own frequency file, to the digit.
+diff -q plink2_rf.flipscan plink2_rd.flipscan
+
+# The .pgen/.pvar/.psam spelling, the three-filename spelling, and the .bed
+# spelling all name the same fileset.
+$BUILD/plink2 $EXTRA1 $EXTRA2 --bfile tmp_data --maf 0.05 --flip-scan --flip-scan-ref-pfile tmp_refp --out plink2_rdp
+diff -q plink2_rd.flipscan plink2_rdp.flipscan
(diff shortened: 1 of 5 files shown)


chrchang:
See the ref/maj-allele handling update in b83b090 , which brings basic --flip-scan in line with plink2's preexisting LD-based commands.

BenjaminDEMAILLE:
Rebased on b83b090 and brought in line with it.

`FlipScanRefDataset()` was doing exactly what you replaced: deriving "is REF the major allele" from `allele_freqs >= 0.5`. It now takes `maj_alleles` and uses `GetAlleleFreq()` the same way `FlipScanRefFreq()` does, with `ref-allele-based` clearing `maj_alleles` rather than being a separate branch.

The one thing that differs from the frequency-file path: the panel frequency comes from a second fileset with its own allele ordering, so `ref_ref_freqs[]` holds the reference's frequency for *this dataset's* REF allele, with the REF/ALT swap already undone by the variant matcher. Selecting the major allele from there is a biallelic complement, which is exact since the mode is restricted to biallelic filesets.

Also worth flagging, since it cost me a while: `2bcfbc48` (b.7.15) fixes the PLINK 1.9 monomorphic-weight bug, so `Tests/TEST_DISTANCE` now fails against any older 1.9. My local one was b.7.11 and I was chasing a phantom regression until I noticed. Building the in-repo 1.9 and putting it first on PATH is what CI does; anyone running the suites locally needs the same.

BenjaminDEMAILLE:
Thanks, that pointer caught a real bug in this PR.

`FlipScanRefDataset()` was taking the dataset's major-allele frequency with the allele count hardcoded:

```c
      const double dataset_maj_freq = GetAlleleFreq(&(allele_freqs[allele_idx_offset_base - variant_uidx]), maj_allele_idx, 2);
```

A multiallelic variant never gets a panel frequency here (both sides have to be biallelic with the same allele pair), but its MAJ_FREQ is still reported, and with the count pinned at 2 that number is wrong whenever the major allele is not REF. On a triallelic variant with REF 0.0975, ALT1 0.7475, ALT2 0.155, MAJ_FREQ came out as 0.9025, i.e. 1 - 0.0975, instead of 0.7475:

```
ID   as-submitted   after
v0   0.9025         0.7475
v1   0.5275         0.5275
v2   0.9125         0.755
v3   0.51           0.51
```

Ten of twenty rows in that test were affected, exactly the triallelic ones; 0.7475 matches what `--freq` reports for the allele.

32c40c52 takes `cur_allele_ct` from `allele_idx_offsets[]` and hoists `maj_allele_idx` next to it, so the block now reads the same as `FlipScanRefFreq()` after b83b090. The panel side keeps the REF/ALT1 complement, which is exact there since `have_panel` already implies a biallelic pair. TEST_FLIP_SCAN passes.

PR #390, opened 2026-09-05
Fix LoadSampleMissingCts() declaration parameter order

BenjaminDEMAILLE:
`plink2_data.h` declares:

```c
..., uint32_t variant_ct, uint32_t y_nosex_missing_stats, uint32_t raw_sample_ct, uint32_t max_thread_ct, ...
```

`plink2_data.cc` defines, and `plink2.cc` calls:

```c
..., uint32_t variant_ct, uint32_t raw_sample_ct, uint32_t y_nosex_missing_stats, uint32_t max_thread_ct, ...
```

Every parameter involved is `uint32_t`, so this compiles, links and behaves correctly today. The declaration is just misleading, in the way that bites the next caller: a new call written against the header's order silently passes a 0/1 flag as the sample count. This matches the declaration to the definition.

Found while reading the function for an unrelated reason; no behavior change.


chrchang:
Good catch.

PR #391, opened 2026-09-06
Tally per-sample missingness during the allele-frequency pass

BenjaminDEMAILLE:
`plink2.cc` has carried this note above the `LoadSampleMissingCts()` call for a while:

```c
// could avoid this call and make LoadAlleleAndGenoCounts() do
// double duty with --missing?
```

This does that for the case where it is cheap.

## What changed

`LoadAlleleAndGenoCounts()` takes an optional `sample_missing_hc_incr` array. When it is non-null, autosomal biallelic variants are read raw, the genotype counts the loop needs are derived from that buffer rather than by `PgrGetDCounts()`, and the missingness bitvector falls out of the same read into the usual acc1/acc4/acc8/acc32 counters.

chrX, chrY and haploid chromosomes keep going through `LoadSampleMissingCts()`, which the caller now runs over only those variants. They need the sex and hethap handling that is not free here, and they are a small share of a typical file.

Everything else keeps the old path:

- `--mind` and `--select-sid-missingness` drop samples, which changes every allele frequency, so their counts still have to be in hand *before* the frequency pass. Only the report has no such ordering constraint, and only that case is deferred.
- Multiallelic files, dosage data and `--minimac3-r2`/`--mach-r2` fall back to the single full pass.
- When the report is not requested, nothing is passed and the pass is byte-for-byte the code it was.

## Numbers

5000 samples x 1M variants, this machine, best of 5:

| | before | after | |
|---|---|---|---|
| `--missing`, autosomes only, 1 thread | 1.07s | 0.75s | 30% |
| `--missing --freq --hardy`, autosomes only, 1 thread | 1.54s | 1.19s | 22% |
| `--missing`, 5% chrX/chrY, 1 thread | 1.07s | 0.81s | 24% |
| `--missing`, autosomes only, 16 threads | 0.48s | 0.34s | 29% |
| `--freq` (fusion inactive), 1 thread | 0.50s | 0.49s | noise |
| `--freq --hardy --geno-counts` (fusion inactive), 1 thread | 0.98s | 0.97s | noise |

It is less than the full second pass because the fused path trades a count-only read for a raw read plus separate counting; the pass still disappears.

## Checking

The failure mode here is silently wrong QC numbers rather than a crash, so the change is pinned from several directions.

`Tests/TEST_MISSING_FUSION` compares the fused output against the unfused one (forced with a `--mind` threshold that removes nobody, asserting that it removes nobody) across autosome-only, mixed-chromosome and multiallelic files, with and without a sample subset, and under `--nonfounders`, `--geno`, `--y-nosex-missing-stats` and the full `scols` set. It also asserts that the separate pass disappears when it should and survives when it should not, and cross-checks the autosomal counts against PLINK 1.9's `--missing`.

Separately, I diffed every output file against a `master` build over 98 invocations spanning autosome-only, chrX-only, chrY-only, chrMT-only, mixed, multiallelic and dosage filesets, including `--loop-cats`, `--keep`, `--mind`, `trim-alts` and `--minimac3-r2-filter`: no differences.

I also checked the suite can fail, by mutating the new code three ways: dropping the chrX/chrY contribution, ignoring the sample subset when deriving the counts, and letting haploid variants be counted by both passes. The suite catches all three.

All existing `Tests/` suites pass.


BenjaminDEMAILLE:
Follow-up measurement, in case it changes how you want the gate written.

The gate here excludes dosage files, so imputed data keeps the second pass. On 3000 samples x 400k variants with `dosage-freq=0.9`, single-threaded:

```
--missing variant-only          0.42s
--missing                       0.67s   (+0.26, i.e. 38% of --missing)
```

So the same share is on the table there as for hardcalls, on the file type most large modern datasets actually are.

The clean way in looks like exposing what the dosage path already computes rather than reimplementing anything: in `GetBasicGenotypeCountsAndDosage16s()`, the non-fast path has `raw_genovec` after `ZeroTrailingNyps()`, and `raw_dosage_present` once the dosage record is parsed, which is exactly `missing_hc` and `missing_hc & ~dosage_present`. Adding two nullable out-params there (and forcing the raw path when they are requested) would cover biallelic dosage files without touching the dosage summation itself. `PgrGetDCounts()` has only the two call sites in `LoadAlleleAndGenoCountsThread()`, so the API churn is small. Multiallelic would additionally need the same treatment in `GetMultiallelicCountsAndDosage16s()`.

I stopped short of writing it because it is a pgenlib change stacked on this PR, and the several dosage encodings each need their own coverage before I would trust the output. Happy to do it if you want it, and happy to shape the API however you prefer -- or to leave it alone if you would rather the dosage path stay untouched.

chrchang:
Nit: the chrY haploid_mask bit can be assumed to be set, and under most circumstances that can be assumed for chrX as well (ab71a9d ).

BenjaminDEMAILLE:
Understood, and thanks for the cleanup. I had kept the explicit chrX/chrY tests because I did not want to rely on the mask covering them; ab71a9d makes the invariant, and the one caller-dependent exception, plain in the comment.

Rebasing my remaining branches onto it now, so they carry your version of the function rather than reintroducing mine.

chrchang:
Similarly, the chrM bit can be assumed to be set in haploid_mask (2691eb9 ).

PR #394, opened 2026-09-07
Revert "Revert "Add --flip-scan (LD-based case/control strand check) …

chrchang:
…(#355)" (#393)"

This restores the #355 branch after the premature merge.


chrchang:
This is a bit of a mess now, so I will plan on merging this within the next two hours with the phenotype-name-parameter logic unimplemented (i.e. the scan for a binary phenotype will be updated to error out if a second binary phenotype is loaded), then #389 and phenotype-name-parameter handling can build on top of that.

PR #395, opened 2026-09-07
--twolocus: optional phenotype name, one table per category

BenjaminDEMAILLE:
Follow-up to #371, and the separate PR you said this could go in. Stacked on that branch, so review it after #371.

#371 stopped `--twolocus` from guessing which phenotype to split on. This brings the split back as an explicit name:

```
--twolocus <variant ID> <variant ID> [phenotype name]
```

Same shape `--cmh` takes its phenotype name in; say the word if you would rather have a separate `--twolocus-pheno` flag.

## Behavior

Per your answer, one table per category, after the ALL table:

```
$ plink2 --bfile data --pheno ph.txt --twolocus b0v0 b0v3 CAT
ALL grp0 grp1 grp2

$ plink2 --bfile data --pheno ph.txt --twolocus b0v0 b0v3 CC
ALL CASE CONTROL
```

A binary phenotype is recoded as a two-category one the way `FstReport()` does it, so the grouping code only deals with categories and CASE/CONTROL falls out of that rather than being a second path.

- Only categories some included sample actually falls in get a table.
- Samples with no value for the named phenotype are counted in ALL alone, so the category tables partition ALL.
- A quantitative phenotype is an error (no categories), as is a name that matches nothing.

## Tests

`Tests/TEST_TWOLOCUS` gains coverage of both phenotype kinds, a check that the category counts sum back to ALL, a check that the ALL table is identical to the no-phenotype run, and both error paths.

Not done here: `--flip-scan` needs the same treatment, which I will send once #394 lands.


BenjaminDEMAILLE:
Rebased: #371 is merged, so this is now just the phenotype-name change on top of master, 177 lines rather than the whole command.

PR #398, opened 2026-09-07
--flip-scan: take an optional case/control phenotype name

BenjaminDEMAILLE:
The phenotype-name half of what you left open in ede5d76d, which made `--flip-scan` refuse to guess once more than one case/control phenotype is loaded. This gives it the way to say which one.

```
--flip-scan ['verbose'] ['zs'] ['ref-allele-based'] [cols=...] [phenotype name]
```

Same shape `--cmh` takes its name in, and `--twolocus` in #395.

## Behavior

- With exactly one case/control phenotype loaded, the name stays optional -- there is nothing to disambiguate, so the error only fires when a choice actually has to be made.
- Naming that sole phenotype gives the same report as not naming it.
- A name matching no phenotype, or naming one that is not case/control, is an error rather than a silent fallback to the scan.

```
$ plink2 --bfile d --pheno ph.txt --flip-scan
Error: --flip-scan requires a phenotype name to be specified when multiple
case/control phenotypes are defined.

$ plink2 --bfile d --pheno ph.txt --flip-scan QT
Error: --flip-scan phenotype 'QT' is not case/control.

$ plink2 --bfile d --pheno ph.txt --flip-scan NOSUCH
Error: --flip-scan phenotype 'NOSUCH' not found.
```

`FlipScan()` needed `pheno_names`/`max_pheno_name_blen` to resolve by name; that is the only signature change.

## Tests

`Tests/TEST_FLIP_SCAN` gains both error paths, a check that naming the sole phenotype matches not naming it, and a check that `CC1` and `CC2` produce *different* reports -- without that last one the argument could be parsed and then ignored, and the suite would not notice.

Categorical phenotypes are not covered here; happy to add them as a follow-up if you still want them, along the lines of #395.

### DIFF
--- a/2.0/Tests/TEST_FLIP_SCAN/run_tests.sh
+++ b/2.0/Tests/TEST_FLIP_SCAN/run_tests.sh
@@ -159,3 +159,44 @@ if $BUILD/plink2 $EXTRA1 $EXTRA2 --bfile tmp_data --flip-scan-ref-freq tmp_panel
     echo "--flip-scan-ref-freq ran without --flip-scan"
     exit 1
 fi
+
+# 9. Which case/control phenotype to split on is named explicitly once more
+#    than one is loaded.
+printf '#FID\tIID\tCC1\tCC2\tQT\n' > tmp_multi_ph.txt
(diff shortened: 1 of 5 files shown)


chrchang:
I will restore maj_alleles after merging.

PR #400, opened 2026-09-08
Add --make-king-table 'rt' and 'ekin' columns

BenjaminDEMAILLE:
Implements the `rt` and `ekin` half of the design you set out on #382:

> * No more --genome command.  --genome's error message points to --make-king-table.
> * --make-king-table gains optional RT ('rt') and EXPECTED_KINSHIP ('ekin') columns.
> * Documentation includes a note about how parent/offspring pairs should be distinguishable from sibling pairs via the ibs0 column.

All three are here. No `--genome` command is added.

**`ekin`** is the pedigree-expected kinship coefficient, on the same scale as the observed `kinship` column, from the standard recursion

```
phi(i, i) = 0.5 * [1 + phi(dad(i), mom(i))]
phi(i, j) = 0.5 * [phi(dad(i), j) + phi(mom(i), j)],  i not an ancestor of j
```

evaluated by expanding whichever of the pair sits deeper in the pedigree. That establishes the second form's precondition, since an ancestor is always strictly shallower, and makes the depth sum decrease at every step, so it terminates without a visited set. It is evaluated lazily rather than as an n x n matrix, so a cohort with no parental IDs costs nothing and a large cohort does not allocate a matrix it would not fill. A parent absent from the fileset is treated as a distinct founder, which is what drops it from the sum.

**`rt`** names only what a .psam can actually spell out: `PO`, `FS`, `HS`, `GG`, `AV`, or `UN`. Anything else with a nonzero expected kinship is `REL`. That last case is deliberate rather than lazy: several distinct configurations share an expected coefficient, and picking one would report more than the pedigree knows. A half-avuncular pair, for instance, comes out `REL` at 0.0625.

**Plumbing** is what you approved: `CalcKing()` and `CalcKingTableSubset()` take `PedigreeIdInfo` instead of `SampleIdInfo`, plus `founder_info`. Parental IDs are resolved locally rather than through `GetTriosAndFamilies()`, because that one applies trio/duo semantics and can narrow the sample set, neither of which belongs in a report with a row per pair. The subsetted path refills the pedigree once per pass, since its collapsed sample_idxs are relative to that pass's sample set.

**The ibs0 note** is in `--help make-king-table`, next to the column list, and `--genome` now gets a PLINK 1.x hint pointing at `--make-king-table`.

**Testing.** New `2.0/Tests/TEST_KING_RT`, built on a hand-made pedigree whose coefficients are known in closed form:

```
s1 x s2  -> s3, s4      s1 x s6 -> s5      s3 x s7 -> s8
```

which covers PO, FS, HS (shared father), GG, AV, the REL fallback for the half-avuncular pair s5/s8 at 0.0625, and three distinct routes to UN. It also checks that neither column appears unless asked for, that `ekin` works without `rt`, and that a fileset with no parental IDs reports nothing but UN. The genotypes are simulated and independent on purpose: these two columns describe the pedigree, so the observed KINSHIP column is irrelevant to what is under test.

Full `2.0/Tests` passes on a clean build.

**One question.** #382 as it stands still adds `--genome`. Say the word and I will close it, or if you would rather keep it open until you have settled the Z0/Z1/Z2 question, that is fine too.


chrchang:
#382 can be closed.

On second thought, `pkin` / `PEDIGREE_KINSHIP` should be a clearer name than `EXPECTED_KINSHIP` (expected based on what?).  Otherwise this looks fine.

BenjaminDEMAILLE:
Renamed in 6820e156: the modifier is `pkin`, the header `PEDIGREE_KINSHIP`, and `kfKingColEkin` and the local variables follow. Agreed that "expected" left open what it was expected from.

```
#FID1	IID1	FID2	IID2	NSNP	HETHET	IBS0	KINSHIP	RT	PEDIGREE_KINSHIP
```

`cols=+ekin` is now rejected as an unrecognized column name, and TEST_KING_RT passes with the new spelling.

Closing #382 as you suggested.

Issue #399, opened 2026-09-08
PLINK 1.9: would you consider dropping the "beta" designation?


BenjaminDEMAILLE:
This is a question about your intent for the 1.9 version scheme, not a request for code changes.

I have a formula adding PLINK 1.9 to homebrew-core under review at Homebrew/homebrew-core#301422. It builds from the `1.9/` directory of this repository, links the system zlib and OpenBLAS rather than vendored copies, and passes CI on macOS 14/15/26-arm64 and Linux x86_64/arm64. The remaining blocker is not technical. A homebrew-core maintainer wrote:

> The upstream repository refers to the version 1 branch of plink as a beta version, so I don't think we should include this in `homebrew-core`.

homebrew-core is reluctant to ship pre-release software, and the `-b.` component of `v1.9.0-b.7.15` together with the README wording is being read as exactly that.

The README currently says:

> It's technically still a beta version because there are a few rarely-used but possibly-worthwhile PLINK 1.07 commands that are still absent, but active feature development for it ended in 2016.

That reads as a deliberate statement about 1.07 feature coverage rather than a claim that the code is unstable, which is also how I would describe it after working in this tree. Ten years of production use point the same way.

So: would you consider dropping the beta designation for the 1.9 line? Concretely, tagging a `v1.9.0` (or `v1.9.1`) without the `-b.` component, and adjusting that README sentence to match, would be enough. The 1.07 gap could stay documented as a known limitation without the version string implying instability.

This is entirely your call, and I am not asking you to close the 1.07 feature gap first. If you would rather keep "beta" as an honest signal about that gap, that is a perfectly good answer. I will relay it to the homebrew-core reviewers, who can then decide knowing your intent explicitly instead of inferring it from the tag name.

One related note, offered as a heads-up rather than an argument: homebrew-core already ships `plink-ng` at 2.0.0-a.7.x, which you label alpha. I did raise that inconsistency on the PR, but I do not want to lean on it, because the likelier outcome is that both formulae get questioned rather than both accepted.

Related: #339.

chrchang:
The other condition I had in mind for reevaluating the beta designation was 1 year without a new bugfix.  There was in fact such a gap between b.7.11 and b.7.12... but then there were three more bugfix releases over the next week of Claude scrutiny.  The latter does constitute real instability, even though it was restricted to commands that few people use.

[...]

PR #402, opened 2026-09-08
1.9: fix two segfaults, three silent wrong-answer bugs, and nine UB sites

BenjaminDEMAILLE:
Following up on #399: you said the natural path to dropping the beta designation is validating the rest of 1.9's commands. This is that, run hostile.

## What was run

1.9 built with `-fsanitize=address,undefined`, then:

- **every one of the 391 flags** `--help` lists, each against seven datasets: normal simulated, 20 trio families, duplicate variant IDs, missing sex and phenotype, two samples with an all-missing and a monomorphic variant, one sample with one variant, and chrX/chrY/chrMT. **2737 runs.**
- **128 fuller invocations** with real covariate, cluster, score and set files.
- **nine input paths** crossed with eighteen commands: `--file`, `--tfile`, `--vcf`, `--data`, `--lfile`, `--ped`/`--map`, compound genotypes, `--keep-allele-order`, `--set-hh-missing`. **162 runs.**
- **44 hand-built malformed inputs**: truncated and empty `.bed`, bad magic, `.bim` with a missing column, non-numeric and negative positions, short `.fam`, duplicated samples, a 5000-character variant ID, CRLF, embedded NUL, more variants than the `.bed` holds, truncated VCF.
- **8000 randomized fuzz iterations**: mutate a `.bed`/`.bim`/`.fam` (byte flips, truncation, splices), then pick one to four flags at random from a table of 130 with adversarial numeric values (0, 1, negative, `1e-300`, `2000000000`, `4294967295`).
- **LeakSanitizer on Linux**, since macOS ASan cannot detect leaks. No leaks.

Everything above is clean now. It was not when I started.

## Two segfaults in the released binary

Both reproduce on current master.

**`--pheno` plus a `.fam` whose last line is short.** `load_fam()`'s first pass advances past the sex column without checking that a token is there. The phenotype column below is the only other thing validating line length, and `--pheno` turns that column off, so the line gets through and the second pass dereferences the null. Reproducer is a three-line `.fam` ending `F1 f1 0 0` plus any `--pheno` file.

**`--genome --filter-cases` with a constrained workspace.** In `populate_pedigree_rel_info()`, the two maternal-index chains call `is_set(founder_info, mm)` before checking the `-1` "no mother" sentinel; the paternal chains beside them check first. Reproduces on clean, unmutated data at every `--memory` setting below the default.

## Three that produce wrong answers silently

**`--hardy` and `--hwe` read uninitialized memory.** Six count arrays use `bigstack_alloc_i()` while the parallel `_allfs` arrays beside them use `bigstack_calloc_i()`. The fill loops skip haploid markers; `hardy_report()` and `enforce_hwe_threshold()` read every marker. So `--hardy` prints heap contents for chrY and chrMT, and `--hwe` filtering is nondeterministic there. UBSan surfaced it as a signed overflow inside `SNPHWE2()` that is only reachable with counts that cannot occur.

**`--neighbour` uses uninitialized indices.** `neighbor_quantiles` is zeroed right after allocation; `neighbor_qindices` beside it never is. A sample that never accumulates `neighbor_n2` neighbors leaves malloc garbage there, and the report loop uses it as a sample index.

**`--distance` on an all-monomorphic variant set writes a matrix of `nan`.** Every weight is zero, the normalizer becomes infinity, and `0 * infinity` is NaN, which the cast to `uint32_t` makes undefined. Master writes the NaN matrix without complaint. I made it stop with an error; that is a judgment call and easy to reverse if you would rather it warned and carried on.

## Two out-of-bounds writes

`--dfam` and `--linear`/`--logistic` both clear trailing words with `buf[count * i - 1]` where `count` is a per-sample word count that can be zero: for `--dfam` when no family units survive, for the GLM path when no sample has both a phenotype and every covariate. `--dfam` also then shifts a 64-bit value by 64. The assertion in `copy_quaterarr_nonempty_subset_excl()` catches the `--dfam` case in a build with assertions, which is how the released binary is built, so the symptom there is an abort. Both now give a clean error.

## Nine undefined-behavior sites

All multi-byte access at an address the standard does not permit, all fine on x86-64 and ARM64 in practice, all constructs a compiler is entitled to assume cannot happen.

The one worth singling out: `qsort_ext2()` stored a pointer at a caller-chosen stride which is routinely not a multiple of the pointer alignment (`sizeof(double) + sizeof(int32_t)` is 12), and the five `*_deref` comparators read it back the same way. That is the sorting path underneath `--assoc`, `--cluster`, `--dosage` and the flag parser, so **every** plink run tripped it. The rest: `double_cmp_decr()` on `--cluster`'s 12-byte records, `getblock32()` in MurmurHash3, `memcpyl3()`, `uint32toa_w4()`, `uint32toa_w8()`, the `--tucc` write alias, a uint32 packed after a variable-length string, a 64-bit shift by 64 in `copy_quaterarr_nonempty_subset_excl()` (genuinely wrong on any target that does not mask the shift count), and `bigstack_shrink_top()` computing its new top by a subtraction that underflows whenever a caller claims more than it reserved, which `--lasso` does deliberately.

Replacing direct accesses with fixed-size `memcpy` costs nothing; every compiler emits the same instruction.

## Regression checking

Built current master and ran both binaries over 128 invocations and seven datasets with a fixed `--seed`: **466 output files compared, and the only difference is the `--dfam` case above, where master aborts.** The full `2.0/Tests` suite also passes against the patched 1.9, which matters because several of those tests compare 2.0's output to 1.9's.

Happy to split this into smaller commits if that reviews better.

### DIFF
--- a/1.9/plink_calc.c
+++ b/1.9/plink_calc.c
@@ -7761,6 +7761,14 @@ int32_t calc_distance(pthread_t* threads, uint32_t parallel_idx, uint32_t parall
     // performance becomes less important than accuracy on 50+ million marker
     // sets.)
     // subtract marker_ct to guard against rounding-driven overflow
+    if (!(dyy > 0.0)) {
+      // Every remaining variant is monomorphic, so every weight is zero.  The
+      // division below then gives infinity, and 0 * infinity is NaN, which the
+      // cast to uint32_t makes undefined.
+      logerrprint("Error: No variant has positive weight; --distance cannot be computed.\n");
+      retval = RET_DEGENERATE_DATA;
+      goto calc_distance_ret_1;
+    }
     dyy = (4294967296.0 - ((double)((intptr_t)marker_ct))) / dyy;
     for (marker_idx = 0; marker_idx < marker_ct; marker_idx++) {
       uii = (uint32_t)(dist_missing_wts[marker_idx] * dyy + 0.5);
@@ -8411,6 +8419,11 @@ int32_t calc_cluster_neighbor(pthread_t* threads, FILE* bedfile, uintptr_t bed_o
       goto calc_cluster_neighbor_ret_NOMEM;
     }
     fill_double_zero(ulii, neighbor_quantiles);
+    // The index array has to be cleared too.  update_neighbor() only writes an
+    // entry when a distance beats the current quantile, so a sample that never
+    // accumulates neighbor_n2 neighbors leaves entries at whatever malloc
+    // returned, and the report loop then uses them as sample indices.
+    fill_uint_zero(ulii, neighbor_qindices);
   }
   fill_ulong_zero(BITCT_TO_WORDCT(initial_triangle_size), cluster_merge_prevented);
   if ((min_ppc != 0.0) || genome_main || read_genome_fname) {
--- a/1.9/plink_common.c
+++ b/1.9/plink_common.c
@@ -257,8 +257,13 @@ unsigned char* bigstack_alloc(uintptr_t size) {
 }
 
 void bigstack_shrink_top(const void* rebase, uintptr_t new_size) {
-  uintptr_t freed_bytes = ((uintptr_t)(g_bigstack_base - ((unsigned char*)rebase))) - round_up_pow2(new_size, CACHELINE);
-  g_bigstack_base -= freed_bytes;
+  // Equivalent to the previous "subtract the freed byte count" form, which
+  // resolved to the same address but computed it by way of an unsigned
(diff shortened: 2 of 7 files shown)


BenjaminDEMAILLE:
Split into four focused PRs so each can be judged on its own, as offered above: #403 (two segfaults and two out-of-bounds writes), #404 (two uninitialized count arrays), #405 (nine undefined multi-byte accesses, no behavior change), and #406 (the --distance NaN matrix, which is the only one that changes output and is a judgment call). Same commits, same verification; nothing dropped.

PR #404, opened 2026-09-08
1.9: zero two count arrays that were read uninitialized

BenjaminDEMAILLE:
Part of the validation pass for #399, split out so each piece stands alone. This one is two arrays that are read before anything writes them.

Both are the same mistake: an array allocated with the uninitialized allocator sitting immediately beside arrays that are zeroed, and read past the point the fill loop covers.

**`--hardy` and `--hwe` read uninitialized memory for haploid markers.** `calc_freqs_and_hwe()` allocates `hwe_lls`/`hwe_lhs`/`hwe_hhs` and the three case-only arrays with `bigstack_alloc_i()`, while the parallel `_allfs` arrays a few lines below use `bigstack_calloc_i()`. The fill loops skip haploid markers, but `hardy_report()` and `enforce_hwe_threshold()` read every marker.

So `--hardy` prints whatever was on the heap for chrY and chrMT, and `--hwe` can include or exclude such a variant depending on it. That last part is the one that worries me: it makes variant filtering nondeterministic rather than merely producing an odd report line.

UBSan is what surfaced it, as a signed integer overflow inside `SNPHWE2()` at a line only reachable with counts that cannot occur in practice, since a zero count returns earlier.

**`--neighbour` uses uninitialized indices.** `calc_cluster_neighbor()` zeroes `neighbor_quantiles` immediately after allocating it, and never zeroes `neighbor_qindices` beside it. `update_neighbor()` only writes an entry when a distance beats the current quantile, so a sample that never accumulates `neighbor_n2` neighbors leaves malloc garbage there, and the report loop uses it to index `sample_idx_to_uidx`. With a corrupted fileset that reaches a wild pointer; with a valid one it silently names the wrong neighbor.

Verified against a build of current master over 128 command invocations and seven datasets with a fixed `--seed`: no output differences.

### DIFF
--- a/1.9/plink_calc.c
+++ b/1.9/plink_calc.c
@@ -8411,6 +8411,11 @@ int32_t calc_cluster_neighbor(pthread_t* threads, FILE* bedfile, uintptr_t bed_o
       goto calc_cluster_neighbor_ret_NOMEM;
     }
     fill_double_zero(ulii, neighbor_quantiles);
+    // The index array has to be cleared too.  update_neighbor() only writes an
+    // entry when a distance beats the current quantile, so a sample that never
+    // accumulates neighbor_n2 neighbors leaves entries at whatever malloc
+    // returned, and the report loop then uses them as sample indices.
+    fill_uint_zero(ulii, neighbor_qindices);
   }
   fill_ulong_zero(BITCT_TO_WORDCT(initial_triangle_size), cluster_merge_prevented);
   if ((min_ppc != 0.0) || genome_main || read_genome_fname) {
--- a/1.9/plink_filter.c
+++ b/1.9/plink_filter.c
@@ -2034,18 +2034,23 @@ int32_t calc_freqs_and_hwe(FILE* bedfile, char* outname, char* outname_end, uint
   if (!hwe_needed) {
     *hwe_lls_ptr = (int32_t*)g_bigstack_base;
   } else {
-    if (bigstack_alloc_i(unfiltered_marker_ct, &hwe_lls) ||
-	bigstack_alloc_i(unfiltered_marker_ct, &hwe_lhs) ||
-	bigstack_alloc_i(unfiltered_marker_ct, &hwe_hhs)) {
+    // Zero-initialized, like the _allfs arrays below: the fill loops skip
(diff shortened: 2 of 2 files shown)


chrchang:
I think --hwe is unaffected by the uninitialized-memory issue, due to the
```
      if ((((int32_t)chrom_idx) == chrom_info_ptr->xymt_codes[MT_OFFSET]) || (is_set(chrom_info_ptr->haploid_mask, chrom_idx) && (!is_x))) {
        continue;
      }
```
branch.  But I see the problem with --hardy chrY when a binary phenotype is defined.

BenjaminDEMAILLE:
You are right on both counts, and I checked it rather than taking it on faith.

I built two copies of master whose `bigstack_alloc()` fills each block, one with `0xa5` and one with `0x00`, and compared outputs. `--hwe` is byte-identical under both on every fileset I tried, including one with chrX/chrY/chrMT: `enforce_hwe_threshold()` reaches the `continue` you quote before reading anything on those chromosomes. The original commit message claimed `--hwe` could include or exclude a variant depending on the heap, and that was wrong. Fixed in 79e75a2f, both in the message and in the code comment.

`--hardy` on chrY with a binary phenotype does show it:

```
 CHR  SNP   TEST  A1  A2                              GENO   O(HET)  E(HET)
  24   y1    ALL   0   T                             0/0/0      nan     nan
  24   y1    AFF   0   T  2779096485/2779096485/2779096485   0.3333   2.276
  24   y1  UNAFF   0   T  2779096485/2779096485/2779096485   0.3333   2.276
```

2779096485 is 0xA5A5A5A5. The ALL row is fine because it reads the `_allfs` arrays, which are already `calloc`'d.

One detail worth flagging, since it argues against the smaller patch: zeroing only the `_cases` triple fixes AFF but leaves UNAFF garbage, because `hardy_report()` reads `hwe_lls/lhs/hhs` for the UNAFF row. Both triples are needed. With both zeroed, `--hardy` output under a 0xa5 workspace is identical to a zero-filled one.

Happy to narrow this to `--hardy` only if you would rather have `hardy_report()` skip haploid chromosomes for the AFF/UNAFF rows instead of relying on zeroed counts; that is a behaviour choice, so I left it as the minimal fix matching the existing `_allfs` treatment.

PR #410, opened 2026-09-08
2.0: join BGEN import threads before freeing their decompressors

BenjaminDEMAILLE:
Fourth from the 2.0 validation pass, and the most serious so far: a heap-use-after-free.

`OxBgenToPgen()`'s exit path frees the per-thread libdeflate decompressors, and only afterwards calls `CleanupThreads()`, which is what joins the workers. So an error detected mid-import destroys objects the scan and compress threads are still using.

```
$ plink2 --bgen truncated.bgen ref-first --sample s.sample --make-pgen
==81788==ERROR: AddressSanitizer: heap-use-after-free
READ of size 4 at 0x626000006778 thread T4
    #0 deflate_decompress_default decompress_template.h:572
    #1 libdeflate_zlib_decompress_ex zlib_decompress.c:69
    #2 plink2::Bgen13DosageOrPhaseScanThread(void*) plink2_import.cc:11630
freed by thread T0 here:
    #1 plink2::OxBgenToPgen(...) plink2_import.cc:14662
```

A truncated `.bgen` is the easy way to reach it, and a partially downloaded file is exactly that. A byte-corrupted one reaches it too.

Moving `CleanupThreads()` to the top of the exit path is the whole fix; nothing else in that block needs to run before it.

Related in kind to #409, and the two together are why I have been looking specifically at what happens on error paths while workers are live. This one is worse because the objects are heap-allocated and freed rather than merely out of scope.

No output change: importing valid BGEN 1.2 and 1.3 files gives byte-identical `.pgen`/`.pvar`/`.psam` against a build of current master, and `2.0/Tests` passes.

### DIFF
--- a/2.0/plink2_import.cc
+++ b/2.0/plink2_import.cc
@@ -14654,6 +14654,10 @@ PglErr OxBgenToPgen(const char* bgenname, const char* samplename, const char* co
     // note that nomem is also possible here
   }
  OxBgenToPgen_ret_1:
+  // Join the workers before freeing anything they hold.  The decompressors
+  // below belong to the scan/compress threads, and an error path reaches here
+  // with those threads still inside libdeflate_zlib_decompress_ex().
+  CleanupThreads(&tg);
   if (common.libdeflate_decompressors) {
     for (uint32_t tidx = 0; tidx != max_thread_ct; ++tidx) {
       if (!common.libdeflate_decompressors[tidx]) {
@@ -14664,7 +14668,6 @@ PglErr OxBgenToPgen(const char* bgenname, const char* samplename, const char* co
     // common.libdeflate_decompressors = nullptr;
   }
   CleanupSpgw(&spgw, &reterr);
-  CleanupThreads(&tg);
   fclose_cond(bgenfile);
   fclose_cond(psamfile);
   CswriteCloseCond(&pvar_css, pvar_cswritep);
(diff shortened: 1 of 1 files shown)


BenjaminDEMAILLE:
Pushed a second commit to this branch. Joining the workers before freeing the decompressors fixed the heap use-after-free, but not the stack: the four thread contexts in the same function are declared inside nested blocks while `CleanupThreads()` is at the exit label, so an error path that jumps out of a block still leaves a worker writing into a scope that has ended.

```
ERROR: AddressSanitizer: stack-use-after-scope
    #0 plink2::UpdateU64IfSmaller(...) plink2_thread.cc:393
    #1 plink2::Bgen13DosageOrPhaseScanThread(void*) plink2_import.cc:11954
Address is located in stack of thread T0
    [1888, 1992) 'scan_ctx' (line 13601) <== inside this variable
```

It reproduces intermittently rather than every time, which is what a race looks like; it took four rounds of the corrupt-input sweep to catch it reliably.

All four contexts move to function scope. Two were called `scan_ctx` and two `ctx` in disjoint blocks, so they had to be renamed on the way out; the renames are mechanical and confined to the block each already lived in. Same verification as before: valid BGEN 1.2 and 1.3 imports are byte-identical to master's, and 2.0/Tests passes.

This is the same shape as #409. Given two instances now, it may be worth deciding on a rule for thread contexts rather than fixing them one at a time.

chrchang:
I think it is reasonable to move all thread-contexts up to the scope of their thread-groups when they aren't currently there.

PR #425, opened 2026-09-09
Port --gene-report from plink 1.9

BenjaminDEMAILLE:
Given an association report and a gene range file, `--gene-report` lists every (gene, variant) pair where the variant falls inside the gene, sorted by gene name. Like `--adjust-file`, it is a batch job: no main dataset is loaded, and the chromosome codes and base-pair coordinates come from the association report itself.

`LoadAndSortIntervalBed()` already did the gene range loading for `--clump-range`, and its comment even names `--gene-report` as a prospective second user. It also already had the `sorted_subset_ids` parameter that `--gene-subset` needs (previously unused, `--clump-range` passes `nullptr`). So the new code is mostly the report scan and the output.

### Flags

```
--gene-report <report file> <gene range file> ['zs'] ['0based'] ['cols='<column set descriptor>]
--gene-list-border <kbs>
--gene-subset <filename>
--gene-report-{chr,pos,id,p}-field <n...>
```

### Modernisation relative to 1.9

* The output is a tab-delimited table with a `#`-prefixed header instead of 1.9's nested human-readable blocks, so it can be loaded directly by downstream tools. Column sets are selectable with `cols=`, and `zs` compresses the output.
* Column names are searched for with `SearchHeaderLine()`, the same way `--adjust-file` and `--clump` search them, so plink 2.0 `.glm` reports (`#CHROM`/`POS`/`ID`/`P`) work without any extra flags and 1.9-style reports (`CHR`/`BP`/`SNP`/`P`) still work. The `--gene-report-*-field` flags override the search order.
* The gene range file can be 0-based/half-open (`0based`), matching `--clump-range0`.
* Overflow in the border computation is clamped rather than wrapping around.

The border is applied to the variant interval rather than to the gene ranges, so the reported gene boundaries and DIST values stay relative to the unextended gene, as in 1.9.

### Example

```
#GENE	CHROM	GENE_START	GENE_END	GENE_KB	DIST	ID	POS	P
GENEA	1	100000	410000	110.002	20	rs1	120000	0.001
GENEA	1	100000	410000	110.002	80	rs2	180000	0.5
GENEB	1	150000	250000	100.001	30	rs2	180000	0.5
GENEC	2	500000	600000	100.001	50	rs3	550000	1e-08
```

### Validation

Cross-checked against plink 1.9 on 12 randomized datasets (about 50000 rows total) and on a 1000000-variant / 20000-gene report (285219 pairs), with and without `--gene-list-border`, `--gene-subset` and `--pfilter`: same rows, same order, same gene boundaries, same kb lengths, same DIST values. The only differences are plink 2.0's chromosome-code output convention (`X` rather than `23`) and DIST precision, since 1.9 rounds DIST to 4 significant digits.

`Tests/TEST_GENE_REPORT` performs that cross-check as part of the test suite: five flag combinations diffed against plink 1.9 (each guarded against a vacuous comparison), plus checks that the P column is dropped when the input has none, that explicit field names reproduce the default-search output, that `0based` reproduces the 1-based output, and that `cols=` and `zs` are consistent with the default output. `Tests/run_tests.sh` passes in full.

Also checked under AddressSanitizer across those flag combinations and a set of degenerate inputs (empty file, empty body, empty gene list, missing/duplicate columns, non-numeric coordinates, `NA`/`INF`/junk p-values, truncated lines).


BenjaminDEMAILLE:
Found a defect in this PR while fuzzing it against 1.9: `--extract` was accepted but silently ignored, so the output was a superset of 1.9's on every `--extract` run (47 of 200 fuzz configurations differed).

Two parts to the fix:

- `GeneReport()` never received a variant filter at all. It now takes the `--extract` filename, loads the ID list into a sorted box, and skips report rows whose ID is not in it.
- The box has to be allocated *after* `LoadAndSortIntervalBed()`'s `BigstackEndReset()`. Allocated before it, the box is freed by that reset and then overwritten from the top by the `(gene, line)` match list, which grows downward from the end of the arena. The symptom was that the ID sorting last would stop matching as soon as one report row matched a gene.

1.9 documents "--extract (without 'range')" here, and the interval forms have no variant positions to work against on this code path since no dataset is loaded, so `--extract bed0`/`bed1`/`range` are now rejected with a specific message.

The 200-run differential fuzz now agrees on every configuration. `TEST_GENE_REPORT` gained two `--extract` cases; the ID list there is deliberately large enough that the arena collision above would reappear if it regressed. Full 2.0 suite passes.

PR #427, opened 2026-09-09
Port --annotate from plink 1.9

BenjaminDEMAILLE:
> **Stacked on #425** (`--gene-report`), which adds the sorted-ID-box loader this reuses. The diff against `master` shown here therefore includes that commit; reviewing #425 first, or reading only the second commit, gives the `--annotate` change on its own.

`--annotate` adds annotations to a variant-based association report, from an attribute file and/or an interval file. Like `--gene-report`, it is a batch job: no main dataset is loaded, and the chromosome codes, base-pair coordinates and variant IDs come from the report itself.

It reuses `LoadAndSortIntervalBed()` for `ranges=` and `filter=` (whose comment already anticipated `--annotate` as a user), and the sorted-ID-box loader added for `--gene-subset` for `snps=` and `subset=`. The interval-membership helpers plink 1.9 uses here, `in_setdef()` and `in_setdef_dist()`, are ported alongside as `InSetdef()` and `InSetdefDist()`.

### Flags

```
--annotate <report file> ['attrib='<file>] ['ranges='<file>] ['filter='<file>]
           ['snps='<file>] ['subset='<file>] [{NA | prune}] ['block'] ['minimal']
           ['distance'] ['zs'] ['0based']
--annotate-border <kbs>
--annotate-{chr,pos,id,p}-field <n...>
```

### Modernisation relative to 1.9

* The output is tab-delimited with a `#`-prefixed header. The report's own columns are passed through with runs of whitespace collapsed to single tabs, and DIST/SGN are ordinary columns rather than fixed-width fields. `zs` compresses the output.
* Column names are searched for with `SearchHeaderLine()`, the same way `--adjust-file` and `--clump` search them, so plink 2.0 `.glm` reports (`#CHROM`/`POS`/`ID`/`P`) work without any extra flags and 1.9-style reports (`CHR`/`BP`/`SNP`/`P`) still work. The `--annotate-*-field` flags override the search order; this replaces 1.9's `--annotate-snp-field`.
* The interval files can be 0-based/half-open (`0based`), matching `--clump-range0`.
* 1.9's `--border` is spelled `--annotate-border`, since plink 2.0 already gives `--clump-range`'s border its own flag name.

### Example

```
#CHR	BP	SNP	P	ANNOT
1	120000	rs1	0.001	GENEA(0)|coding|synonymous
1	180000	rs2	0.5	GENEA(0)|GENEB(0)|coding
2	550000	rs3	1e-8	GENEC(0)|utr
3	100	rs4	0.2	.
```

### Validation

Cross-checked against plink 1.9 on 5 randomized datasets across 14 flag combinations (70 runs, about 200000 rows), covering `ranges=`, `attrib=`, both together, `block`, `NA`, `prune`, `minimal`, `filter=`, `snps=`, `subset=`, `distance` with a border, and `--pfilter`: identical rows, identical annotations, identical DIST/SGN values. The only differences are plink 2.0's chromosome-code output convention (`X` rather than `23`) and the precision of the parenthesized distances inside ANNOT, which 1.9 rounds to 4 significant digits.

`Tests/TEST_ANNOTATE` performs that cross-check as part of the test suite (13 flag combinations diffed against plink 1.9, each guarded against a vacuous comparison), plus checks that `0based`, explicit field names, and `zs` all reproduce the default output. `Tests/run_tests.sh` passes in full.

Also checked under AddressSanitizer across those flag combinations and a set of degenerate inputs (empty report, header-only report, empty/attribute-less/blank attribute files, empty interval files, missing P column with `--pfilter`, missing files, and every rejected flag combination).


chrchang:
--annotate will probably get a backward-incompatible overhaul.  The PLINK 1.x implementation is very rarely used, and meanwhile PLINK 2.0 could use a command which bridges the divide between the VCF-style INFO column and regular tab-delimited files.

BenjaminDEMAILLE:
That makes sense, and it is a better target than what this PR does. Happy to drop the 1.x semantics.

Concretely, the divide has two directions, and plink2 currently has neither:

* **INFO out.** `--extract-if-info` / `--require-info` can filter on a key, but there is no way to materialize one as a column, so anything downstream of a `.pvar` has to re-parse `INFO` itself. Something like `--info-to-cols <key>...` writing the named keys as ordinary columns (with a configurable missing value, and `.` for flag keys) would cover it.
* **INFO in.** The reverse: read a tab-delimited file keyed by `ID`, or by `CHROM`/`POS`/`REF`/`ALT` when IDs are not unique, and fold its columns into `INFO` as `KEY=value`, so annotations produced by other tools survive a `--make-pgen` round trip and stay filterable by the existing `-if-info` flags.

If that is roughly the shape you have in mind, I would rather build that than land this. Two questions before I do:

1. Should it be one command with a direction modifier, or two separate ones? I lean towards two, since the column-name and key-name handling differs.
2. Should the interval-based half survive at all? `ranges=` here annotates a variant with the genes containing it, which is the part of 1.x `--annotate` that is not expressible as INFO plumbing. It could be folded into the same command as a synthetic key, or left out entirely.

I will hold this PR open but stop adding to it until you have had a chance to say. `--gene-report` in #425 is the same family, so tell me if that one should wait too.

PR #435, opened 2026-09-09
CI: bump the copyright end year on 1 January

BenjaminDEMAILLE:
Follow-up to #433, which fixed four copyright years that the last bulk update missed. This makes the yearly edit something nobody has to remember.

A scheduled workflow runs on 1 January, rewrites the end year, and opens a pull request. **Nothing is pushed to `master` directly**, so the change is still reviewed; it just arrives already written.

### What it rewrites

Two line forms, and only those:

```
// This file is part of PLINK 2.0, copyright (C) 2005-2026 Shaun Purcell,
  "(C) 2005-2026 S Purcell, C Chang, B Demaille      GNU General Public License v3\n";
```

The first identifies a file as PLINK's own, so the bundled third-party sources cannot match: htslib, zstd, libdeflate, SFMT and QD all name their own authors. The `LICENSE` files carry the GPL's own 2007 date rather than a copyright year, so they are out of scope too. Only the digits are rewritten, which leaves the banners' column padding alone.

### Why a script rather than a sed pipeline

`.github/scripts/bump_copyright_year.py` does the rewrite, so it behaves the same on any runner and can be run by hand (`python3 .github/scripts/bump_copyright_year.py 2027`). I first wrote this as `grep -Z | xargs -0 sed -i`, and my own local check of it silently did nothing because BSD `sed -i` takes its backup suffix differently; a job that runs unattended should not depend on that.

The workflow then re-checks **every added line** against the two expected forms and fails instead of committing if anything else changed. That guard is the point: nobody will be watching this run.

### Verification

Ran it for 2027 against the current tree:

* 139 files, 141 lines changed; every changed line matches an expected form.
* The bundled htslib headers keep their `2013-2015` and `2012-2014` dates.
* The result compiles clean, and `plink2` prints `(C) 2005-2027 S Purcell, C Chang, B Demaille      GNU General Public License v3` with its alignment intact.
* A second run changes nothing, so a manual re-trigger is harmless.

Two things worth knowing about GitHub's scheduler, both noted in the file: it disables scheduled workflows in a repository with no commit activity for 60 days, and it only runs them from the default branch. `workflow_dispatch` is there for both cases.


BenjaminDEMAILLE:
The red check here is not this PR. `can_install_pgenlib_package` is the only workflow that runs on it (the others are path-filtered and this branch only touches `.github/`), and that job is currently broken by Google's apt repository serving an index whose hash does not match its `Release` file, which makes `setup-r`'s `apt-get update` exit non-zero:

```
E: Failed to fetch https://dl.google.com/linux/chrome-stable/deb/dists/stable/main/binary-amd64/Packages.gz  Hash Sum mismatch
##[error]Failed to get R release: Failed to get R 4.6.1: Failed to install R
```

It started at roughly 17:25 UTC today; every run of that workflow before then passed, on this branch and others. #436 removes the Chrome source from that job, and the two runs three minutes apart make the point:

```
17:43  failure  ci/copyright-year-bump      (without the fix)
17:40  success  ci/pgenlibr-apt-sources     (with it)
```

So this will go green once #436 lands, or once Google's index is consistent again. Nothing to change here.

BenjaminDEMAILLE:
Rebased onto master. The red check was the Google Chrome apt source's hash mismatch breaking `apt-get update` and so the R install, which is what #436 fixed; this branch predated that merge and did not have it.

chrchang:
Looks good.

One helpful refinement: make this submit separate PRs for 1.9 and 2.0.  The 1.9 workflow would be immediately useful: I forgot to update the 1.9 copyright year in code comments when I released the first 1.9 update of the year last week.  (And, conversely, I'd like to leave the 1.9 copyright year un-updated if I've made changes to 2.0 but not 1.9 since the beginning of the year.)

BenjaminDEMAILLE:
Done, and your second point turned out to change the trigger too.

**Separate pull requests.** A matrix job runs the rewrite scoped to `1.9/` and to `2.0/` and opens one PR each, titled `1.9: update copyright end year to YYYY` and likewise for 2.0. Only those two trees carry PLINK copyright lines (41 and 98 files), so the split is clean; there is now a second guard that fails the job if the rewrite reaches outside the tree that job is responsible for, alongside the existing one that fails if it rewrites a line of the wrong shape.

**Only bump a tree that has changed.** Each job skips unless that directory has a commit in the current year, so a year in which you touch 2.0 and not 1.9 leaves 1.9's lines alone.

That second condition makes 1 January the wrong day to run: nothing has happened yet, so both jobs would skip, and the bump would then have to wait until the following year. The useful moment is the year's first change to a tree, which is exactly the case you hit last week. So the schedule is now weekly instead, and does nothing on the weeks where both trees are already current. `workflow_dispatch` still lets you force it.

If you would rather it fired on push to master instead of weekly, that is a one-line change; I went with the schedule to keep it off the critical path of every merge.

PR #437, opened 2026-09-09
Add --info-to-cols

BenjaminDEMAILLE:
First half of the INFO/tab-delimited bridge you described on #427.

This is the direction VCF-derived data usually has to leave plink2 in. `--extract-if-info` and `--require-info` can filter on an INFO key, but nothing could get one *out*, so anything downstream of a `.pvar` has to re-parse the INFO column itself.

```
--info-to-cols {<key>[,<key>...] | 'all'} ['zs']
```

writes `<output prefix>.vinfo`:

```
#CHROM	POS	ID	REF	ALT	DP	AF	DB	VT
1	100	rs1	A	G	30	0.25	1	SNP
1	200	rs2	C	T	12	NA	0	SNP
1	300	rs3	G	A,C	NA	0.1,0.2	1	NA
1	400	rs4	T	A	NA	NA	0	NA
```

### Decisions worth flagging

* **The key list is comma-separated.** INFO keys cannot contain commas, so this stays unambiguous against the `zs` modifier; a space-separated list could not be told apart from one. `all` takes every key the `##INFO` lines declare.
* **`Number=0` means Flag**, so that column is 0 or 1. Any other key is its value, or `NA` where the variant does not carry it.
* **A key present without a value that the header did not declare a Flag is reported as `1`.** That is the only reading available when the header does not say otherwise; keys absent from the header entirely still work, reported by presence.

It reuses `PvarInfoReload()`, so the INFO column is streamed from the `.pvar` rather than held in memory, and `InfoReloadIsNeeded()` now covers this command.

### Testing

`TEST_INFO_TO_COLS` checks the output **against the source VCF**, not against another plink2 run: `check_vinfo.py` parses the VCF itself, works out what every cell should be from the `##INFO` declarations, and asserts. It also covers `all` including a key no variant carries, an undeclared key, Zstd round-tripping, reading the VCF directly against reading the `.pgen`, variant filters applying, and the error when the fileset has no INFO column. `2.0/Tests` passes in full.

### The other direction

The reverse, folding a tab-delimited file's columns into `INFO` as `KEY=value` so annotations survive a `--make-pgen` round trip and stay filterable by the existing `-if-info` flags, is the natural companion but touches the `.pvar` writer, so I have kept it out of this PR. My two questions from #427 still stand, and the second one matters more for that half: whether it should be one command with a direction modifier or two, and whether the interval-based half of 1.x `--annotate` survives at all.


chrchang:
Hmm, this doesn't have the kind of speed advantage over `bcftools query` that I'd look for in such a command.  The utils/vcf_subset/ program is almost 3x as fast at scanning a VCF in my testing.

This has a narrow but highly relevant job which is applicable to ordinary VCF files as well as .pvar files; it should be extremely efficient for that job.

BenjaminDEMAILLE:
You were right: reusing PvarInfoReload() after a full load meant reading the file twice. When no filters are present, --info-to-cols now reads the .pvar/VCF once, tokenizes each line only up to INFO, jumps to the end of the line with Rawmemchr, and reads INFO in one vector pass that stops once all requested keys are found. On a 2M-variant, 200-sample VCF it takes 0.36s plain and 0.75s bgzipped, against 2.52s and 3.18s for bcftools query and 0.28s and 0.72s for vcf_subset --no-samples. The output matches the old path, which is still used when filters are given.

PR #439, opened 2026-09-09
Add --write-var-ranges

BenjaminDEMAILLE:
Divides the variants that pass your filters into a given number of equal-size blocks and writes each block's first and last variant ID, which is what makes `--snps` usable for splitting a job across machines.

```
--write-var-ranges <block ct> ['zs']
```

Output is `<output prefix>.var.ranges`, with 1.9's `FIRST`/`LAST` columns under a `#`-prefixed header.

Block sizes come from a running division rather than a fixed width, so they differ by at most one variant when the count does not divide evenly. That detail is what makes the boundaries match 1.9's exactly rather than approximately.

### Verification

Against PLINK 1.9 at block counts 1, 2, 3, 7, 10, 97 and one block per variant: the rows are byte-identical every time, including both extremes.

`TEST_WRITE_VAR_RANGES` runs that comparison, and also covers Zstd output round-tripping and the error when the block count exceeds the variant count. `2.0/Tests` passes in full.


chrchang:
This should have --write-snplist's duplicate-ID check, and 'allow-dups' workaround.

BenjaminDEMAILLE:
Added, in a second commit: `--write-var-ranges` now runs `CheckIdUniqueness()` unless the new `allow-dups` modifier is given, with the same error text as `--write-snplist` adapted to this flag. The output is fed back in with `--snps`, so a duplicate ID makes the ranges ambiguous the same way it makes a snplist ambiguous.

`TEST_WRITE_VAR_RANGES` gains the duplicate case: a .bim with one ID repeated, checked to be rejected, then accepted under `allow-dups` with the expected block count.

Rebased onto master (through 3f62faca); the misc flag takes bit 52, since `--make-perm-pheno` claimed 51 upstream.

PR #440, opened 2026-09-09
2.0: port --make-perm-pheno

BenjaminDEMAILLE:
PLINK 1.9's `--make-perm-pheno` writes a file of permuted copies of a phenotype, for use as the input of an externally-scripted permutation test. This ports it to PLINK 2.0.

```
--make-perm-pheno <ct> [pheno name] [zs]
```

Differences from PLINK 1.9, all of which follow PLINK 2.0 conventions:

- the phenotype may be named, so the flag is usable when more than one is loaded (1.9 errors out);
- the output is an ordinary `.pphe` with a `#FID`/`IID` header rather than a headerless file;
- `zs` compresses the output.

Fisher-Yates with `RandU32()`, which is unbiased, and one shuffle buffer regardless of the permutation count.

## Testing

`Tests/TEST_MAKE_PERM_PHENO`. The values cannot be compared against PLINK 1.9, since the two draw from different random streams, so the test checks the invariants instead:

- every column is a rearrangement of the input values (case/control and quantitative);
- a sample whose phenotype is missing stays missing in every column;
- the columns are distinct from each other;
- the same `--seed` reproduces the file, a different seed changes it;
- `zs` round-trips through `--zst-decompress`;
- naming the phenotype is required when more than one is loaded, and naming it works.

### DIFF
--- a/2.0/Tests/TEST_MAKE_PERM_PHENO/.gitignore
+++ b/2.0/Tests/TEST_MAKE_PERM_PHENO/.gitignore
@@ -1,0 +1,4 @@
+plink19*
+plink2*
+tmp_*
+*.log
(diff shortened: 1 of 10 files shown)


chrchang:
This may eventually be worth extending to categorical phenotypes, but no need to do that before implementation of e.g. multinomial logistic regression.

PR #441, opened 2026-09-09
2.0: port --list-23-indels

BenjaminDEMAILLE:
PLINK 1.9's `--list-23-indels` writes the subset of `--write-snplist`'s output whose variants carry 23andMe-style indel calls, so that they can be handled separately from the SNPs (`--23file` import already points at them: "N variants with indel calls present"). This ports it to PLINK 2.0.

```
--list-23-indels ['zs']
```

A variant qualifies when at least one allele code is `D` or `I` and every allele code is `D`, `I` or missing, which is what PLINK 1.9's two-allele test amounts to. Two things follow from being in PLINK 2.0 rather than 1.9:

- multiallelic variants are handled, and a `D`/`I` variant that also has an ACGT allele is correctly not listed;
- both missing-allele spellings are recognized: plink2's `.`, and the opposite call that PLINK 1.9's `--23file` importer writes for a monomorphic indel.

Like `--write-snplist`, it is permitted on a split-chromosome dataset.

## Testing

Added to `Tests/TEST_23FILE`, whose fixture already has `DD`/`DI`/`II` calls:

- against PLINK 1.9 on the same .bed, with a non-emptiness check so the comparison isn't vacuous;
- the same list from plink2's own `--23file` import, where the monomorphic indel's other allele is `.`;
- `zs` round-trips through `--zst-decompress`;
- a hand-written expectation on the cases PLINK 1.9 cannot represent: a multiallelic `D`/`I`/`A` variant, `DEL`/`INS` spelled out, a plain SNP, and a `D`/`G` variant. Only the two genuine indel calls are listed.


chrchang:
This should have --write-snplist's duplicate-ID check, and 'allow-dups' workaround.

BenjaminDEMAILLE:
Added, in a second commit: same duplicate-ID check and `allow-dups` modifier as `--write-snplist`.

One decision worth flagging: the check covers every variant that passed the filters, not just the indel subset that gets written. The .indel list is fed back in with `--extract`/`--exclude`, so what matters is whether a written ID matches more than one variant in the dataset, which a non-indel duplicate does too.

`TEST_23FILE` gains that case: a fileset where the duplicate is one of the indels, plus a duplicate outside the subset, checked to be rejected and then accepted under `allow-dups`.

Rebased onto master (through 3f62faca).

BenjaminDEMAILLE:
The ubuntu CI failure here is not this port: it is `TEST_META_ANALYSIS`'s weighted-Z comparison, which this branch does not touch (its only test change is `TEST_23FILE`). The message is in the uploaded log rather than the console, since the suite redirects each test's stdout:

```
WEIGHTED_Z differs on causal_snps_49: 8.961 vs 8.97163
```

That test is mine, and the check was badly chosen: the fixture was regenerated from an unseeded `--simulate` on every run, and the 0.1% band did not allow for PLINK 1.9's approximation of the inverse normal CDF in the tail, where a |Z| ~ 9 variant's p-value lives. #480 seeds the fixture, gives that one column a band with the reason written down, and adds a recomputation of plink2's weighted Z from the study files that pins it to 1e-5 without going through anyone's approximation.

Re-running the job here should be green on its own, since the fixture is a fresh roll each time; the underlying fix is #480.

PR #442, opened 2026-09-09
2.0: port --tucc

BenjaminDEMAILLE:
PLINK 1.9's `--tucc` builds a pseudo-case/pseudo-control dataset out of the trios in a fileset: for each child with both parents present, one sample carries the child's genotypes and one carries the alleles the parents did not transmit, so a TDT-equivalent test can be run with any case/control method. This ports it to PLINK 2.0.

```
--tucc ['vzs']
```

Output is `<prefix>.tucc.pgen` + `.pvar` + `.psam` (PLINK 1.9's `write-bed` output; `--make-bed` on the result gives the .bed form, and `--export ped` the .ped form, so the two PLINK 1.9 output modes are both reachable). Sample IDs get PLINK 1.9's `_T`/`_U` suffix, and the pseudo-case is a case and the pseudo-control a control.

## Implementation note

Each parent transmits exactly one allele, so the untransmitted ALT count is `(paternal + maternal) - child` no matter which parent transmitted what: the pair is determined by the three ALT counts alone. That replaces PLINK 1.9's 64-entry lookup table, and it also decides Mendel consistency, since the reachable child ALT counts are the contiguous range `[(dad == 2) + (mom == 2), (dad != 0) + (mom != 0)]`. A trio that falls outside it, or has any genotype missing, yields two missing genotypes, as in PLINK 1.9.

Haploid, chrMT and multiallelic variants have no well-defined untransmitted genotype and are excluded, with a count reported. PLINK 1.9 does the same for the first two and cannot represent the third.

## Testing

New `Tests/TEST_TUCC`, on a 40-family fixture whose parental genotypes are drawn from a per-variant allele frequency and whose children's alleles are drawn from the parents, with injected Mendel errors and missing calls. Two families are deliberately incomplete (one child has a single parent, one is a founder) and must not appear in the output.

- 22800 genotypes checked against the answer the fixture generator derived from the genotypes it drew, 19320 of them nonmissing, so the comparison isn't against another PLINK run;
- the whole fileset compared against PLINK 1.9's `--tucc write-bed` through .ped, with heterozygote allele order normalized since A1/A2 assignment differs;
- the .psam contents, including the exclusion of the two incomplete families;
- chrX/chrMT/multiallelic exclusion, checked against the variant list the fixture wrote;
- `vzs` compresses the .pvar and leaves the .pgen and .psam byte-identical;
- a dataset with no trios warns instead of erroring, and writes nothing;
- an MT-only dataset errors out.


chrchang:
Untransmitted alleles should still be defined for multiallelic variants.

BenjaminDEMAILLE:
Implemented, in a second commit.

The untransmitted alleles are well-defined for a multiallelic variant: pick a Mendel-consistent assignment of the child's two alleles to the two parents, and each parent's other allele is untransmitted. Where more than one assignment exists it leaves the same unordered pair behind, so the result is the multiset difference (paternal + maternal) - child, which is the identity the biallelic ALT-count arithmetic already relies on.

The multiallelic path reads with `PgrGetM`, works in allele codes rather than ALT counts, and writes the pair back out through `SpgwAppendMultiallelicSparse`; the .pvar now emits every ALT, and the .pgen declares the written allele counts. Biallelic variants keep the existing fast path untouched. chrX is in `haploid_mask`, so the multiallelic code only ever sees autosomes and chrXY, and needs no sex-dependent recoding.

`TEST_TUCC` gains 60 three-allele variants, checked as allele pairs through an exported VCF rather than as ALT counts, which cannot express them. The checker fails if no genotype pairs two distinct ALT alleles, so it cannot pass without real multiallelic support (it sees 665 such genotypes on the committed seed). The PLINK 1.9 comparison now runs on the biallelic subset, since 1.9 cannot read the multiallelic variants at all.

Clean under `-fsanitize=address,undefined`, and rebased onto master (through 3f62faca).

chrchang:
I will set kfMiscTuccVzs to (1LLU << 56) after merge.

PR #445, opened 2026-09-09
Pad the flattened string lists so the sanitizer build runs to completion

BenjaminDEMAILLE:
**Correction: you're right, and the original claim in this PR was wrong.** These reads cannot fault. See the discussion below; the PR now stands or falls on sanitizer usability alone, not memory safety.

`AllocAndFlattenEx()` (plink2_cmdline.cc) and `AllocAndFlattenCommaDelimEx()` (plink2.cc) `malloc()` exactly the number of bytes their flattened, double-null-terminated list needs. Consumers walk that list with `strnul()`, whose vectorized `Rawmemchr()` reads a whole aligned vector at a time and so touches up to `kBytesPerVec - 1` bytes past the final terminator. Because the loads are vector-aligned they never straddle a page boundary, so they cannot fault — but AddressSanitizer flags them, and these two buffers are the only exact-size `malloc()`s in the test suite that get scanned this way (everything else is bigstack). The result is that an ASan build aborts in `Tests/TEST_PGEN_FREQ` before reaching anything else.

This pads both allocations with `kBytesPerVec` zeroed bytes so the sanitizer build runs to completion. The alternative would be `__attribute__((no_sanitize("address")))`, but that would have to go on nine functions in plink2_string.cc rather than two allocation sites.

Your call whether that's worth carrying — closing this is a perfectly reasonable outcome. I'm flagging it only because #446, which is a real bug, only turned up once the suite could run all the way through under ASan.

## The two reports

`--keep`, via `AllocAndFlattenEx()` (fires in `Tests/TEST_PGEN_FREQ`):

```
==31515==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6020000007b0
READ of size 16 at 0x6020000007b0 thread T0
    #0 plink2::Rawmemchr(void const*, int) plink2_string.cc:32
    #1 plink2::LoadSampleIds(...) plink2_common.cc:1855
    #2 plink2::KeepOrRemove(...) plink2_filter.cc:1239
    #3 plink2::Plink2Core(...) plink2.cc:1792

0x6020000007bb is located 0 bytes after 11-byte region [0x6020000007b0,0x6020000007bb)
allocated by thread T0 here:
    #1 plink2::AllocAndFlattenEx(...) plink2_cmdline.cc:2832
    #2 main plink2.cc:8483
```

`--clump`, via `AllocAndFlattenCommaDelimEx()` (fires in `Tests/TEST_PHASED_VCF`):

```
==44560==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x604000001bb0
READ of size 16 at 0x604000001bb0 thread T0
    #0 plink2::Rawmemchr(void const*, int) plink2_string.cc:54
    #1 plink2::ClumpReports(...) plink2_ld.cc:8468
    #2 plink2::Plink2Core(...) plink2.cc:3253

0x604000001bb4 is located 0 bytes after 36-byte region [0x604000001b90,0x604000001bb4)
allocated by thread T0 here:
    #1 plink2::AllocAndFlattenCommaDelimEx(...) plink2.cc:3456
    #2 main plink2.cc:5382
```

Both faulting loads are at `0x...7b0` and `0x...bb0`, i.e. 16-byte aligned, which is exactly what the `RoundDownPow2()` in `Rawmemchr()` guarantees.

## Reproducing

`2.0/build_san` is a copy of `build_dynamic`'s Makefile with

```
BASEFLAGS += -fsanitize=address,undefined -fno-sanitize=alignment -fno-omit-frame-pointer -g
LINKFLAGS += -fsanitize=address,undefined -fno-sanitize=alignment
```

and `-O1` in place of `-O2`/`-O3`. `-fno-sanitize=alignment` is needed because plink2 intentionally does unaligned loads in several places. Happy to drop that directory from this PR if you'd rather not carry it.

## Testing

With this change, all 29 test directories in `2.0/Tests` pass under `-fsanitize=address,undefined` with no diagnostics. (`TEST_DISTANCE` fails in my environment for an unrelated reason: the `plink` on my PATH is a Homebrew build of b.7.11, which predates the 7 Sep 2026 `bin4` diagonal bugfix that the test now expects.) The suite also passes normally on a plain `build_dynamic` build.

### DIFF
--- a/2.0/build_san/Makefile
+++ b/2.0/build_san/Makefile
@@ -1,0 +1,233 @@
+# Linux/macOS Makefile for PLINK 2.0.
+#
+# Compilation options (leave blank after "=" to disable, put "= 1" to enable):
+#   Do not use AVX2 instructions: NO_AVX2
+#   Do not use SSE4.2 instructions: NO_SSE42
+#   Print clear error message if SSE42/AVX2 needed but missing: CPU_CHECK
+#   Do not link to LAPACK: NO_LAPACK
+#   Use only -O2 optimization for zstd (may be necessary for gcc 4.x): ZSTD_O2
+#   Statically link zlib: STATIC_ZLIB
+#   Statically link zstd: STATIC_ZSTD
+#   Link to MKL with 64-bit indexes (dynamically): DYNAMIC_MKL
+#     (this also requires MKLROOT and MKL_IOMP5_DIR to be defined, and
+#     LD_LIBRARY_PATH to include the appropriate directories)
+#   Link statically to AOCL with 64-bit indexes: STATIC_AOCL
+#     (this supports older OSes than dynamic linking)
+#   32-bit binary (also sets STATIC_ZLIB and ZSTD_O2):
+#     FORCE_32BIT (warning: you may need to add a zconf.h symlink to make that
+#     work)
+#   Debug symbols: change DEBUG to -g (still incompatible with -flto?)
+NO_AVX2 = 1
+NO_SSE42 = 1
+CPU_CHECK = 1
+NO_LAPACK =
+ZSTD_O2 = 1
+STATIC_ZLIB =
+STATIC_ZSTD = 1
+DYNAMIC_MKL =
+MKLROOT = /opt/intel/oneapi/mkl/2025.1
+MKL_IOMP5_DIR = /opt/intel/oneapi/compiler/2025.1/lib
+STATIC_AOCL =
+AOCLROOT = /home/admin/aocl/5.1.0/gcc
+FORCE_32BIT =
+# Giving up on -flto default for now (Mar 2026) due to the high frequency of
+# gcc 10.2 crashes.  Will try again after next upgrade of my own build
+# machines.
+# DEBUG = -flto
+DEBUG =
+CC ?= gcc
+CXX ?= g++
+
+BASEFLAGS= -DZSTD_MULTITHREAD -ffp-contract=off
+# ***** end configuration *****
+
+BASEFLAGS += ${DEBUG}
+
+include ../Makefile.src
+
+LINKFLAGS=-lm -lpthread ${DEBUG}
+ZLIB=
+ARCH32=
+CPUCHECK_FLAGS = ${DEBUG}
+
+ifdef FORCE_32BIT
+  # this should work on Debian 10 with the right packages installed.
+  STATIC_ZLIB = 1
+  ZSTD_O2 = 1
+  BASEFLAGS += -msse -mfpmath=sse
+  ARCH32 = -m32 -march=i686
+  CXXFLAGS = -std=c++17
+else
+  ifdef NO_AVX2
+    ifndef NO_SSE42
+      BASEFLAGS += -msse4.2
+      ifdef CPU_CHECK
+        BASEFLAGS += -DCPU_CHECK_SSE42
+        CPUCHECK_FLAGS = ${DEBUG} -O2 -DCPU_CHECK_SSE42 ${CXXWARN2}
+      endif
+    endif
+  else
+    BASEFLAGS += -mavx2 -mbmi -mbmi2 -mfma -mlzcnt
+    ifdef CPU_CHECK
+      BASEFLAGS += -DCPU_CHECK_AVX2
+      CPUCHECK_FLAGS = ${DEBUG} -O2 -DCPU_CHECK_AVX2 ${CXXWARN2}
+    endif
+  endif
+  # plink2_base tries to use C++ <execution> header when compiled with C++20 or
+  # newer; this typically requires something like -ltbb to be added to
+  # LINKFLAGS.
+  CXXFLAGS = -std=c++17
+endif
+BASEFLAGS += ${ARCH32}
+BASEFLAGS += -fsanitize=address,undefined -fno-sanitize=alignment -fno-omit-frame-pointer -g
+LINKFLAGS += -fsanitize=address,undefined -fno-sanitize=alignment
+
+CFLAGS=-O1 -std=gnu99
+# zstd appears to be seriously targeted at -O3; see 26 Jul 2016 entry at
+# cbloom.com/rants.html
+ifdef ZSTD_O2
+  ZCFLAGS=-O1 -std=gnu99
+else
+  ZCFLAGS=-O3 -std=gnu99
+endif
+# this actually needs to be named "CXXFLAGS"
+CXXFLAGS += -O1
+
+############################
+# PLINK2 Linux build update:
+
+# Old BLASFLAGS=-llapack -lblas -lcblas -latlas fails on Ubuntu 22.04/24.04 (lcblas/Atlas gone).
+
+# Fixed: BLASFLAGS=-llapacke -llapack -lopenblas
+
+# Install: libopenblas-dev liblapack-dev liblapacke-dev zlib1g-dev
+
+# macOS build unaffected (still uses Accelerate)
+
+# Works cleanly in Docker / GitHub Actions
+
+## Old:
+# BLASFLAGS=-llapack -lblas -lcblas -latlas
+
+## Update:
+BLASFLAGS=-llapacke -llapack -lopenblas
+
+###########################
+
+ifdef STATIC_ZLIB
+  BASEFLAGS += -DSTATIC_ZLIB
+  LINKFLAGS += -L. ../../zlib-${ZLIB_VER}/libz.a
+else
+  LINKFLAGS += -lz
+endif
+
+SKIP_STATIC_ZSTD =
+ifndef STATIC_ZSTD
+  BASEFLAGS += -DIGNORE_BUNDLED_ZSTD
+  ZCSRC2 =
+  ZSSRC2 =
+  SKIP_STATIC_ZSTD = echo
+  OBJ = $(CSRC:.c=.o) $(CCSRC:.cc=.o)
+  OBJL = $(notdir $(OBJ))
+  LINKFLAGS += -lzstd
+endif
+
+LIBTOOL=ar
+STATIC=rcs
+UNAME := $(shell uname)
+ifeq ($(UNAME), Darwin)
+  ifdef FORCE_32BIT
+    $(error 32-bit OS X builds are not supported)
+  endif
+  ifdef DYNAMIC_MKL
+    $(error MKL is not currently supported on OS X)
+  endif
+  ifdef STATIC_AOCL
+    $(error AOCL is not currently supported on OS X)
+  endif
+  BLASFLAGS=-framework Accelerate
+  CFLAGS += -I/usr/local/include
+  CXXFLAGS += -I/usr/local/include -stdlib=libc++ -DACCELERATE_NEW_LAPACK
+  LINKFLAGS += -L/usr/local/lib
+  LIBTOOL=libtool
+  STATIC=-static
+else
+  ifdef DYNAMIC_MKL
+    ifdef NO_LAPACK
+      $(error DYNAMIC_MKL and NO_LAPACK conflict)
+    endif
+    ifdef FORCE_32BIT
+      $(error DYNAMIC_MKL + FORCE_32BIT not supported)
+    endif
+    ifdef STATIC_AOCL
+      $(error DYNAMIC_MKL + STATIC_AOCL does not make sense)
+    endif
+    BASEFLAGS += -DDYNAMIC_MKL -DLAPACK_ILP64 -I${MKLROOT}/include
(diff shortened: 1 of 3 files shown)


chrchang:
Hmm.  64-bit Rawmemchr() is supposed to perform vector-aligned loads:

```
  const VecI8* ss_viter = R_CAST(const VecI8*, RoundDownPow2(starting_addr, kBytesPerVec));
```

So, while it can read `kBytesPerVec - 1` bytes past the terminator, that should only be possible when that read wouldn't cross a page boundary; am I misunderstanding something here?

BenjaminDEMAILLE:
You're not misunderstanding anything; I was wrong. `RoundDownPow2()` makes every load vector-aligned, an aligned `kBytesPerVec` load cannot straddle a page boundary since the page size is a multiple of it, and the loop only ever advances by whole vectors. So the vector containing the terminator is always fully inside the same page as the terminator, and the over-read cannot fault. My "faults if the allocation ends on a page boundary" claim has no case behind it.

The two reports agree with you rather than with me, which I should have noticed: the faulting reads are at `0x6020000007b0` and `0x604000001bb0`, both 16-byte aligned, i.e. exactly the addresses `RoundDownPow2()` produces.

So there is no memory-safety bug here, and closing this is a perfectly reasonable outcome. What is left is narrower: ASan doesn't model the alignment trick, so it aborts on these loads, and because bigstack is one big allocation these two exact-size `malloc()` buffers are the only ones in the whole test suite that get scanned this way. Without padding them an ASan build dies in `Tests/TEST_PGEN_FREQ` and never reaches the rest.

I've retitled the PR and pushed a commit correcting the code comments and the rationale, so nothing in it claims unsafety any more. Three ways to go, in the order I'd rank them:

1. Close it, and anyone wanting to run ASan patches locally.
2. Keep the padding as-is: two allocation sites, `kBytesPerVec` bytes each, no behavior change.
3. `__attribute__((no_sanitize("address")))` on the scanners instead, which leaves memory exactly as it is — but that means nine functions in plink2_string.cc (`Rawmemchr`, `rawmemchr2`, `rawmemchr3`, `strchrnul3`, `FirstPrecharFar`, `NextTokenMultFar`, `TokenLexK0`, `Memrchr`, `LastSpaceOrEoln`) rather than two allocation sites, and it would suppress genuine out-of-bounds detection inside them.

Whichever you prefer. The reason I'd like *some* form of it is #446: that one is a real out-of-bounds write, and it only turned up after the suite could run all the way through under the sanitizer.

chrchang:
How about rounding the allocation sizes up to a multiple of kBytesPerVec?

BenjaminDEMAILLE:
Done — pushed, both sites now use `RoundUpPow2(tot_blen, kBytesPerVec)`. Tidier, and it costs nothing when the total is already a multiple. All 29 test directories still pass under `-fsanitize=address,undefined` with no diagnostics.

One caveat worth stating explicitly, since it applies to the fixed-padding version I had before as well. Rounding the *size* up silences ASan exactly when the *address* is already vector-aligned, which is where `pgl_malloc` comes in: it is a plain `malloc()`, which guarantees 16 bytes of alignment on 64-bit. So:

- SSE2 and ARM builds have `kBytesPerVec == 16`, alignment matches, and rounding up is airtight.
- AVX2 builds have `kBytesPerVec == 32`. A `malloc()` result at 16 mod 32 leaves the last aligned 32-byte vector hanging 16 bytes past the end even after rounding, so ASan would still flag it.

And on that same AVX2 case the leading `RoundDownPow2(starting_addr, kBytesPerVec)` steps 16 bytes *before* the buffer, which is an equally valid ASan complaint that neither rounding nor fixed padding addresses.

If you want it clean on AVX2 too, the shape that covers both ends is `vecaligned_malloc()` with the rounded size, since that aligns to `kBytesPerVec` by construction — at the cost of having to match every `free()` on these buffers with `aligned_free()`. I have not pushed that: it is a wider change, and I can't measure it here (arm64 Mac, so `kBytesPerVec` is 16 and the AVX2 case is reasoning rather than an observed failure). Happy to do it if you'd rather have it, or to leave this as is for the 16-byte-vector builds.

BenjaminDEMAILLE:
One more concrete data point for the usability argument, from running the 2.0 test suite under an ASan + UBSan build of current master:

```
+ .../plink2 --bfile tmp_data --output-chr 26 --keep subset.id --geno-counts cols=-provref --out plink2_subset
==85789==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6020000007b0
READ of size 16 at 0x6020000007b0 thread T0
    #0 plink2::Rawmemchr(void const*, int) plink2_string.cc:32
    #1 plink2::LoadSampleIds(...) plink2_common.cc:1855
    #2 plink2::KeepOrRemove(...) plink2_filter.cc:1239
0x6020000007bb is located 0 bytes after 11-byte region [0x6020000007b0,0x6020000007bb)
allocated by thread T0 here:
    #1 plink2::AllocAndFlattenEx(...) plink2_cmdline.cc:2832
```

ASan aborts the process, so the run stops at that point and the rest of the suite never executes. With the rounding in this PR applied the same build gets past it.

The 11-byte allocation is the `--keep subset.id` filename; the read is vector-aligned as you said, so nothing is actually wrong at runtime. This is only about being able to run the suite under the sanitizer to look for real defects elsewhere.

BenjaminDEMAILLE:
Result with this PR's rounding applied, same ASan + UBSan build, running the full 2.0 test suite (`Tests/run_tests.sh`, with a current 1.9 on PATH):

```
All tests passed.
```

1854 lines of log, 0 sanitizer reports. Without the rounding the same build stops after 170 lines, at the `--keep` above. So the practical effect of this change is the difference between the sanitizer build being unusable past the first `--keep`/`--extract`-style flag and it covering the whole suite.

chrchang:
Hmm.  My experience is that `malloc()` returns 16-byte aligned addresses on macOS but may return only 8-byte-aligned addresses on 64-bit Linux.  So I don't think this solution is reliable even in the non-AVX2 64-bit case.

However, your original solution should be?  And we may as well use `kBytesPerVec - 1` bytes of padding instead of `kBytesPerVec`.  (Could also round down to a word-multiple afterward.)

BenjaminDEMAILLE:
You are right, and I should have caught that: rounding the size up only helps when the base is already vector-aligned. With an 8-aligned base and kBytesPerVec 16, `RoundUpPow2(tot_blen, 16)` can still end 8 bytes short of the last vector `Rawmemchr()` touches, so the version you reviewed was not actually reliable on 64-bit Linux.

Pushed the padding form instead, `kBytesPerVec - 1` as you suggested. That bound holds whatever `malloc()` returns: the scan starts at `RoundDownPow2(addr, kBytesPerVec)` and reads whole vectors, so the furthest byte past the terminator is `kBytesPerVec - 1`, independent of the base address.

Verified on the ASan + UBSan build: the full 2.0 test suite runs to `All tests passed.` (1837 log lines, 0 sanitizer reports). Without the padding the same build stops after 170 lines at the first `--keep`.

On rounding down to a word multiple afterward: I left it as the plain `+ kBytesPerVec - 1`, since the saving is at most 7 bytes on one allocation per flag and the expression stays obviously correct. Happy to add it if you would rather the arithmetic match what the tail actually reads.

chrchang:
Sounds good, don't think this function is called enough for round-down to matter.

One other thing worth trying, not necessarily in this PR: there is a -DNO_UNALIGNED compile flag that is supposed to remove the unaligned memory accesses.  Of course, we shouldn't *only* run ASan on the -DNO_UNALIGNED configuration, since that would remove coverage of `#ifndef NO_UNALIGNED` code that executes in plink2 production binaries.

BenjaminDEMAILLE:
Tried  under ASan + UBSan, on top of this PR's padding.

Result: the full 2.0 suite passes there too, 1837 log lines, 0 sanitizer reports. So it does not surface anything the default configuration misses on this suite, and it does not replace it either, for the reason you gave: the  paths are what production binaries run, and building with the flag removes them from coverage.

The useful shape looks like running both configurations rather than choosing between them, so I have kept my local harness on the default build and will run  alongside it when I next sweep for memory bugs. If you would rather that pairing live in CI I am happy to send it, but it doubles the sanitizer job's runtime for what is currently no extra findings, so I would rather wait until it earns its place.

Nothing changed in this PR from that; it is still just the  padding.

BenjaminDEMAILLE:
(Backticks got eaten by my shell in the comment above; reposting it readable.)

Tried `-DNO_UNALIGNED` under ASan + UBSan, on top of this PR's padding.

Result: the full 2.0 suite passes there too, 1837 log lines, 0 sanitizer reports. So it does not surface anything the default configuration misses on this suite, and it does not replace it either, for the reason you gave: the `#ifndef NO_UNALIGNED` paths are what production binaries run, and building with the flag removes them from coverage.

The useful shape looks like running both configurations rather than choosing between them, so I have kept my local harness on the default build and will run `NO_UNALIGNED` alongside it when I next sweep for memory bugs. If you would rather that pairing live in CI I am happy to send it, but it doubles the sanitizer job's runtime for what is currently no extra findings, so I would rather wait until it earns its place.

Nothing changed in this PR from that; it is still just the `kBytesPerVec - 1` padding.

chrchang:
To clarify: was `-DNO_UNALIGNED` run with `-fno-sanitize=alignment` removed from the compile/link flags?

BenjaminDEMAILLE:
No, it was not, and that invalidates what I reported. I built `-DNO_UNALIGNED` on top of my existing sanitizer Makefile, which still carried `-fno-sanitize=alignment`, so the alignment check was off and "0 reports" said nothing about alignment. Sorry for the noise.

Rebuilt with `-DNO_UNALIGNED` and `-fsanitize=address,undefined` with **no** `-fno-sanitize=alignment`. The suite still reaches `All tests passed.`, but it is no longer clean: UBSan reports one misaligned load.

```
+ plink2 --bcf tmp_data8.bcf dosage=HDS --out tmp_data10
../plink2_import.cc:5217:23: runtime error: load of misaligned address 0x107e000245ca
    for type 'const float *', which requires 4 byte alignment
```

`-DNO_UNALIGNED` does not remove it, because the access is not one of the unaligned-load primitives that flag controls. It comes from casting a byte pointer into the BCF record buffer to `const float*`:

```c
// plink2_import.cc:5385 and :7011
dosage_main = R_CAST(const float*, dosage_main_raw);
```

BCF FORMAT payloads carry no 4-byte alignment guarantee, so that pointer can land on any offset, and everything downstream inherits it.

The suite only reaches one of the three sites. Driving a BCF with GP dosages through the same path reaches the other two, and those are ordinary indexed loads rather than memcpy:

```
+ plink2 --bcf db.bcf dosage=GP --make-pgen --out dg
../plink2_import.cc:5217:23: ... 'const float *'
../plink2_import.cc:5170:28: ... 'const float'      // cur_gp_start[0]
../plink2_import.cc:5174:28: ... 'const float'      // cur_gp_start[1]
```

On x86-64 and ARM64 these are harmless, which is why nothing has ever gone wrong. But `NO_UNALIGNED` exists for targets where they are not, and this path is outside its reach, so a build made for such a target would still fault here.

The fix looks mechanical: have `ParseBcfBiallelicGp()` and `ParseBcfBiallelicDosage()` take `const unsigned char*` and `memcpy` each float out, which is already what `ParseBcfBiallelicDosage()` does for its first read. Happy to send that as its own PR; it is unrelated to the padding in this one, so I would rather not fold it in here.

So the useful answer to your question is: run both, and run the `NO_UNALIGNED` one *with* the alignment check, since that combination is the only one that says anything about alignment. It is not a substitute for the default build, for the reason you gave.

BenjaminDEMAILLE:
Sent the BCF part as #458, since it is unrelated to the padding here. It also answers the practical half of your question: the alignment check does find something that `-DNO_UNALIGNED` alone does not remove, so the two are worth running together rather than either one on its own.

BenjaminDEMAILLE:
Answering the `-DNO_UNALIGNED` question from this thread: no, it had not been run with `-fno-sanitize=alignment` removed. `build_san/Makefile` passes that flag unconditionally, so `-DNO_UNALIGNED` alone compiles the aligned-access paths and then turns off the only check that says anything about them.

Posted as #482: a `NO_UNALIGNED=1` switch that compiles `-DNO_UNALIGNED` and keeps the alignment check, as a second configuration rather than a replacement. It found two pre-existing `--pmerge` bugs, both in the same-position scan arenas, both on `TEST_PMERGE`'s `--merge-mode nm-match` case: `RescanOnePosRecord` and `SamePosPvarRecord` are variable-length records packed back to back, so every record after the first is written through a misaligned pointer. Fixes are in the same PR.

PR #450, opened 2026-09-10
2.0: port --homozyg group

BenjaminDEMAILLE:
`--homozyg`'s help currently ends with "the 'group'/'group-verbose' pool reports, 'consensus-match', 'extend', --homozyg-match and --pool-size are not implemented yet". This implements `group`, `consensus-match`, `--homozyg-match` and `--pool-size`, writing `<prefix>.hom.overlap`. `group-verbose` (a separate per-pool output format) and `extend` (which changes the scan, not the pooling) stay unimplemented.

## Pool discovery

A pool is a maximal set of mutually overlapping runs. For an interval graph those are exactly the sets of runs covering some run's first variant, so the sweep emits the active set at each run start where at least one member ends before the next start; anything else is a subset of a later set. That replaces PLINK 1.9's back-to-front heap walk with a forward sweep, and finds the same pools.

Pools are numbered by decreasing size, then by position, as PLINK 1.9 does; members are ordered by allelic-match group and then by sample.

## Matching

Two runs match when, over the variants where both samples are homozygous, at most `1 - --homozyg-match` (default 0.95) carry *different* homozygous genotypes. The comparison region is the pair's own overlap, or the pool's consensus region under `consensus-match` — which is what PLINK 1.9's `compute_allelic_match_matrix()` does, with `roh_slot_uncached` selecting between the two. Groups are then formed greedily: the unassigned run matching the most others becomes a group's reference (`*` in the GRP column), everything still unassigned that matches it joins, repeat.

## Agreement with PLINK 1.9

Note that PLINK 1.9's `--homozyg group` aborts on an assertion for every input I tried, so the reference outputs below come from a build with #448 applied.

On a 60-sample fixture with planted 600-variant runs (22 pools, 419 report rows), the two agree **row for row** — pool IDs, member order, KB, NSNP, NSIM, GRP with its reference marker, and the CON/UNION rows:

| run | rows | mismatches |
|---|---|---|
| default | 419 | 0 |
| `--homozyg-match 0.8` | 419 | 0 |
| `--pool-size 12` | 419 | 0 |
| `--pool-size 16` | 325 | 0 |
| `--pool-size 19` | 176 | 0 |
| second fixture, chr1+chr2, case/control | 104 | 0 |

Two settings diverge, and in both PLINK 1.9's report contradicts itself. Group 1 is formed before anything has been assigned, so its size must equal its reference's NSIM plus one. Counting pools where that fails:

| run | PLINK 1.9 | this PR |
|---|---|---|
| `--homozyg-match 0.95` | 0 of 22 | 0 of 22 |
| `--homozyg-match 0.8` | 0 of 22 | 0 of 22 |
| `--homozyg-match 0.99` | **5 of 22** | 0 of 22 |
| `consensus-match` | **1 of 22** | 0 of 22 |

At 0.99 the NSIM values agree exactly between the two programs; only the grouping differs. PLINK 1.9 caches match verdicts across pools in a matrix indexed by ROH *slot*, and slots are recycled as runs are flushed — while the flush only clears the matrix row, not the column. So the bits driving the grouping can disagree with the counts that produced NSIM, which is what the table shows. This port keeps no cross-pool cache, so it cannot get into that state.

Under `consensus-match` there is a second effect: `is_allelic_match()` advances to the 64-variant block containing the region start and compares whole words, so the region compared is rounded outward to block boundaries. In the default mode that is harmless, since the extra positions fall outside one member's run and are padded as missing; under `consensus-match` both members usually have real data there, so the comparison silently covers more than the consensus region. This port compares exactly the documented region.

I have not tried to reproduce either behaviour. Say the word if you'd rather have bit-compatibility than self-consistency here.

## Testing

`Tests/TEST_HOMOZYG` gains structural checks, since PLINK 1.9 cannot produce a reference on CI. `check_pool.py` verifies, for every pool: the consensus region is contained in each member's run and the union contains it, the CON/UNION member counts match the row count, pool IDs increase while sizes do not, group numbers are 1..n with exactly one reference each, NSIM is below the pool size, each member's run appears in the `.hom`, and group 1's size equals its reference's NSIM plus one. Run on the default report, on `consensus-match`, and on `--homozyg-match 1`. Plus: the header, the pool count, that `--pool-size` drops smaller pools, and that a `--pool-size` nothing reaches leaves no report.

The rest of `2.0/Tests` passes, and `TEST_HOMOZYG` also passes under `-fsanitize=address,undefined` with no diagnostics. (`TEST_DISTANCE` fails in my environment only because the `plink` on my PATH is a Homebrew build of b.7.11, which predates the 7 Sep 2026 `bin4` diagonal bugfix; unrelated.)

Performance note: genotypes are loaded per pool rather than through PLINK 1.9's lookahead buffer, so heavily overlapping pools reread variants. That was the simple thing to get right first; happy to add the sharing pass if it matters.


BenjaminDEMAILLE:
Cross-validation update. Two bugs in 1.9's `--homozyg group` came out of comparing this port against it, now sent as #451 (group members stolen by later, lower-NSIM groups) and #452 (`consensus-match` comparing a block-aligned window instead of the consensus segment).

With those two applied on top of #448, 1.9 and this port now produce identical `.hom.overlap` output in all four modes (`consensus-match` and `--homozyg-match` 0.8 / 0.95 / 0.99), on three datasets, and across `--homozyg-snp`, `--homozyg-kb`, `--homozyg-window-snp`, `--homozyg-window-het`, `--homozyg-density` and `--homozyg-gap` variations: 419 rows, 0 differences each.

So the answer to the bit-compatibility question in the description is that this port is bit-compatible with a fixed 1.9, and deliberately not with the current one.

BenjaminDEMAILLE:
Fuzz + sanitizer results for this port.

384 randomly generated .bed/.bim/.fam triples (2-12 samples, 1-400 variants, 1-3 chromosomes, homozygous runs broken by random het and missing calls), each run through `--homozyg group` and `--homozyg group consensus-match` under an AddressSanitizer + UBSan build:

- 0 sanitizer reports
- 0 differences against 1.9 with #448, #451 and #452 applied, comparing every `.hom.overlap` field (KB compared with a 1e-5 relative tolerance, since 2.0 prints 6 significant digits and 1.9 prints 3 decimals; that is the same convention the existing 2.0 `.hom` writer uses)

Also clean under the sanitizer across `--homozyg-match`, `--homozyg-snp`, `--homozyg-window-snp`, `--homozyg-kb`, `--homozyg-het`, `--homozyg-window-het`, `--homozyg-density` and `--homozyg-gap` variations on the larger 60-sample dataset.

PR #453, opened 2026-09-10
Add --genome (method-of-moments IBD report)

BenjaminDEMAILLE:
Ports PLINK 1.x's `--genome` to 2.0: for each pair of samples, the method-of-moments estimates of the probability of sharing 0, 1 and 2 alleles IBD, plus `PI_HAT = Z2 + Z1/2`.

The help text says plainly that these estimates assume a homogeneous, randomly-mating population and that `--make-king{-table}` is the better choice for relatedness filtering; this is here for compatibility with existing pipelines that parse a `.genome` file.

### What is included

- `--genome`, with `zs`, `unbounded`, `nudge`, and a `cols=` column set (`maybefid`, `fid`, `id`, `maybesid`, `sid`, `rt`, `z`, `pihat`, `phe`, `dst`, `nsnp`, `ibs`, `homhom`, `hethet`). The default set reproduces 1.9's default report.
- `--genome-min-pi-hat` / `--genome-max-pi-hat`, replacing 1.x's `--min` / `--max`, which were too generic a spelling for 2.0.
- `--ppc-gap` and the PPC / RATIO columns (the IBS2\* test).
- `--parallel`, `--threads`, `--nonfounders` and `--read-freq` support.

The `RT` column comes from the .fam pedigree, using the same relationship-type logic as `--make-king-table`'s `rt` column.

### What is not included

1.x's `EZ` column (pedigree-based expected relatedness). `--make-king-table`'s `pkin` column already reports the pedigree kinship coefficient on a well-defined scale, and `RT` covers the categorical part.

### Testing

`Tests/TEST_GENOME` checks against PLINK 1.9 on a 200-sample, 600-variant simulated fileset with a 50-family pedigree (father, mother, two children) spread over three chromosomes, so both the RT column and the `--ppc-gap` window have something to work on. 1.9 prints four decimals, so the floating-point columns are compared with a 1e-4 tolerance while RT and PHE must match exactly.

Ten scenarios:

1. defaults
2. `unbounded`
3. `nudge`
4. `--nonfounders`
5. `--min 0.2` and `--max 0.15` against `--genome-min-pi-hat` / `--genome-max-pi-hat`
6. `--ppc-gap 0`, `20`, `2000`
7. IBS counts against 1.9's `--genome full`, including the invariants `NSNP == IBS0 + IBS1 + IBS2` and `NSNP == HOMHOM + HETHET + IBS1` (19900 pairs)
8. `--threads 1` reproduces the multithreaded report byte for byte
9. the three `--parallel i 3` chunks concatenate to the whole report
10. `zs` round-trips through `--zst-decompress`


chrchang:
"Existing pipelines that parse a `.genome` file" are supported by continued maintenance of PLINK 1.9.  That type of thing is explicitly a non-goal for 2.0, as discussed in #382 .

BenjaminDEMAILLE:
You are right, and this PR should not have been opened. We settled exactly this in #382: no `--genome` command, its error message points at `--make-king-table`, and the pedigree information people used it for arrives as the `rt` and `pkin` columns, which is what #400 delivered. I closed #382 myself on that basis.

What happened is that the branch predated that discussion and was never deleted; I found it unsubmitted this week and pushed it without checking why it had been left alone. Closing.

I will re-read #344 before sending any more 1.9 ports, and ask there rather than guess when the answer is not already on record.

PR #455, opened 2026-09-10
Add --neighbour (nearest-neighbour IBS report)

BenjaminDEMAILLE:
Ports PLINK 1.x's `--neighbour`: for every sample, its n1th- through n2th-closest neighbours by identity-by-state, with a Z-score for each rank. The Z-score is taken over that rank's values across all samples, so a large negative Z marks a sample sitting much further from its neighbours than the rest of the cohort does, which is what the report is for.

### Where the numbers come from

The IBS value is the one `--distance ibs flat-missing` writes to `.mibs`: the proportion of shared alleles over the variants where both samples are called. That is what PLINK 1.x's `--neighbour` reports, and the test checks the column against the matrix cell for the named pair.

`--neighbour` rides on `--distance`'s existing pairwise pass rather than adding one, so `--distance ibs flat-missing --neighbour 1 5` costs a single pass. A `--distance` that leaves the default frequency-weighted rescaling in place would need a different accumulation, so that combination is rejected with a message that says which modifier to add. `--parallel` is rejected too, since a top-n list needs every pair.

### One deliberate difference from 1.x

When every sample shares the same value at a rank (two samples, or a duplicated fileset), the Z-score is undefined. PLINK 1.x prints a `nan` or a `0` from `0/0`; this writes `NA`.

### Testing

`Tests/TEST_NEIGHBOUR`, against PLINK 1.9 on a 120-sample, 600-variant simulated fileset:

1. `--neighbour 1 1`
2. `--neighbour 1 5`
3. `--neighbour 3 7`, a range that does not start at 1
4. `--neighbour 4 4`, a single rank above 1
5. `--neighbour 3 7` is a slice of `--neighbour 1 7`
6. the `IBS` column equals the `.mibs` cell for the named pair
7. `zs` round-trips through `--zst-decompress`
8. `cols=` trims the report, and a full `cols=` reproduces the default one byte for byte
9. `--parallel` is rejected
10. `--neighbour` with a default-rescaling `--distance` is rejected, and with `--distance ibs flat-missing` produces the same report

Beyond the checked-in test, 666 runs over 150 randomly generated filesets (2-30 samples, 1-120 variants, mixed sexes, chrX/Y/MT, all-missing and monomorphic variants) across five `(n1, n2)` pairs: the only rows that differ are the two-sample cases where the Z-score is undefined, plus 85 rows out of ~30000 where two candidates share a sample's IBS exactly and the two programs' rescalings round the tie differently.

### Note on comparing against 1.9

`--neighbour <n1> <n2>` with n1 > 1 is broken in PLINK 1.9: the report rows carry rank 1 and up while being labelled n1 and up. That is #454, sent separately, and it has to be applied before 1.9 and this port can be compared for n1 > 1.

### DIFF
--- a/2.0/Tests/TEST_NEIGHBOUR/.gitignore
+++ b/2.0/Tests/TEST_NEIGHBOUR/.gitignore
@@ -1,0 +1,3 @@
+plink19*
+plink2_*
+tmp_*
--- a/2.0/Tests/TEST_NEIGHBOUR/compare.awk
+++ b/2.0/Tests/TEST_NEIGHBOUR/compare.awk
@@ -1,0 +1,43 @@
+# Compares a PLINK 1.9 .nearest report against a plink2 one.
+#
+# PLINK 1.9 prints four significant digits, so IBS and Z are compared with a
+# tolerance.  When two candidates share a sample's IBS value exactly, which one
+# is named is down to the last bits of two different rescalings, so a row whose
+# numbers agree but whose neighbour differs is counted as a tie rather than a
+# mismatch; the test caps how many of those it will accept.
+function abs(x) { return (x < 0)? -x : x }
+function close_enough(a, b) { return abs(a - b) <= 1e-4 + 1e-4 * abs(a) }
+FNR == NR {
+    if (FNR > 1) {
+        # FID IID NN MIN_DST Z FID2 IID2
+        k = $1 "\t" $2 "\t" $3;
+        dst[k] = $4; z[k] = $5; nbr[k] = $6 "\t" $7;
+        ++n1;
+    }
+    next
+}
+/^#/ { next }
+{
+    # FID IID NN IBS Z FID2 IID2
+    ++n2;
+    k = $1 "\t" $2 "\t" $3;
+    if (!(k in dst)) { print "row missing from the 1.9 report: " k; exit 1 }
+    if (!close_enough(dst[k], $4)) {
+        print "IBS mismatch on " k ": " dst[k] " vs " $4; exit 1
+    }
+    if (z[k] == "nan" || z[k] == "-nan" || $5 == "NA") {
+        # Undefined Z: 1.9 prints a NaN or a zero from 0/0, plink2 prints NA.
+        # Both sides have to see it as undefined.
+        if (!(($5 == "NA") && (z[k] == "nan" || z[k] == "-nan" || z[k] + 0 == 0))) {
+            print "Z definedness mismatch on " k ": " z[k] " vs " $5; exit 1
+        }
+    } else if (abs(z[k] - $5) > 2e-3) {
+        print "Z mismatch on " k ": " z[k] " vs " $5; exit 1
+    }
+    if (nbr[k] != $6 "\t" $7) { ++ties }
+}
+END {
+    if (n1 != n2) { print "row count mismatch: " n1 " vs " n2; exit 1 }
+    if (ties > n2 / 20) { print ties " tied rows out of " n2 ", too many to be ties"; exit 1 }
+    print n2 " rows matched (" ties+0 " ties)"
+}
--- a/2.0/Tests/TEST_NEIGHBOUR/run_tests.sh
+++ b/2.0/Tests/TEST_NEIGHBOUR/run_tests.sh
@@ -1,0 +1,89 @@
+#!/bin/bash
+
+# --neighbour, checked against PLINK 1.9's .nearest report.
+
+set -exo pipefail
+
+plink --simulate simulate.txt --simulate-missing 0.03 --simulate-ncases 60 --simulate-ncontrols 60 --out tmp_data > /dev/null
+
+compare() {
+    awk -f compare.awk "$1" "$2"
+}
+
+# 1. The single nearest neighbour.
+plink --bfile tmp_data --neighbour 1 1 --out plink19_1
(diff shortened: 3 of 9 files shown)


BenjaminDEMAILLE:
Flagging this against your note on #453 and your answer on #311, since I wrote it before re-reading either.

`--neighbour` is in the same module #311 covers, and your position there was that MDS is low priority because it is straightforward in R or Python once the distances exist. The same applies here with more force: this report is a top-n sort of a matrix `--distance ibs flat-missing` already writes, plus a per-rank Z-score. Three lines of R on the .mibs file.

So the honest case for it is thin, and if it reads as a 1.9 port for its own sake, close it and I will not argue. The thing I would not want lost either way is #454: `--neighbour <n1> <n2>` with n1 > 1 has been printing rank-1 values under rank-n1 labels in 1.9, and that fix stands on its own.

If you do want it, the one argument I would make is that the Z-score is the point rather than the neighbour list, and computing it correctly means being careful about which rank's distribution you standardize against, which is exactly what 1.9 got wrong.

BenjaminDEMAILLE:
CI was red and is now fixed. The cause was self-inflicted and worth stating plainly: `Tests/TEST_NEIGHBOUR` compared `--neighbour 3 7` and `--neighbour 4 4` against PLINK 1.9, and released 1.9 is wrong for any first rank above 1 (#454). So correct output was being checked against a wrong reference.

Those two ranges now check against plink2's own `--neighbour 1 7` instead: the n1..n2 report has to be a slice of the 1..n2 one, which is the same requirement without depending on 1.9 carrying the fix. The n1 = 1 comparisons against 1.9 stay, since 1.9 is right there.

Verified by running the suite against the released homebrew 1.9, i.e. the same binary CI has.

This does not change my read from the previous comment: the case for the command itself is thin, and I would not argue if you close it.

chrchang:
Current practice is to look at outliers in principal-component space, rather than using IBS for this.  So `--neighbour` should not survive as-is.  Section 3.4 of https://academic.oup.com/bioinformatics/article/36/16/4449/5838185 outlines a similar PCA-based calculation that may be worth implementing (possibly with minor modifications) under the same name.

Meanwhile, while we're revisiting `--pca`, I now think a `--grm-maf` (alias: `--grm-min-af`) flag should be added with the following behavior:
* If specified, `--grm-maf` behaves like `--maf` applied just to `--pca` / GRM-construction.  Since the GRM calculation supports multiallelic variants, `--grm-maf` should accept the same second parameter `--maf` does.  Since the calculation becomes wildly unstable for MAF much smaller than the inverse-square-root of the sample size, there could be a `yes-really` modifier which could be required if the given threshold is less than a quarter of the inverse-square-root of the sample size. 
* If `--grm-maf` is not specified, and --pca / GRM-construction encounters a variant with MAF less than a quarter of the inverse-square-root of the sample size, PLINK 2 now errors out, with a message reporting the minimum MAF remaining in the dataset and the `--grm-maf` options.

`--homozyg-maf` can be modified to have similar behavior (i.e. not strictly required if all variants have MAF >= 0.05).

BenjaminDEMAILLE:
Agreed, and that is the right reason to drop it: not that the report is redundant, but that IBS is the wrong space to look for outliers in. Closing this.

I read section 3.4 of Privé et al. before answering so I am not guessing at what you have in mind. The statistic there is a modified probabilistic local outlier factor on the PC scores: for sample j with K nearest neighbours j_1..j_K in PC space,

```
pd_j   = (1/K) * sum_k d(j, j_k)^2
stat_j = pd_j / ((1/K) * sum_k pd_{j_k})
```

so a sample scores high when it sits far from its own neighbours *and* those neighbours are themselves tightly clustered, which is what separates a genuine outlier from a point in a sparse region. The paper is explicit that the threshold is left to visual inspection rather than being automatic, so the command's job is to emit the statistic and the neighbour distances, not to make a call.

Three things I would rather settle before writing it than guess at:

1. **Where the PCs come from.** Computing them inside `--neighbour` duplicates `--pca`, and the natural thing is for `--neighbour` to imply or attach to `--pca` and consume its scores, so that `--pca 10 --neighbour` is one pass. Alternatively it could read a `.eigenvec` back. I lean towards the first.
2. **How many PCs, and K.** The paper uses the PCs the user chose to keep and a K the user sets. A default K of 5 or 10 and "all computed PCs" would be my starting point, but you may have a view.
3. **Whether the name should be reused.** Reusing `--neighbour` means a 1.9 command line silently computes something different. Given that 1.9 stays available, and that the old output is not what anyone should be using, I think that is fine, but it is your call and it is cheap to pick a new name instead.

Say which way you want each of those and I will write it as a fresh PR against `--pca`. Meanwhile I will start on `--grm-maf`, which you spelled out completely.

Worth keeping regardless of what happens here: #454, the 1.9 fix for `--neighbour <n1> <n2>` printing rank-1 values under rank-n1 labels. That is a defect in the released program independent of whether 2.0 ever grows the command.

BenjaminDEMAILLE:
Posted the replacement as #483: `--neighbour` now reports the section 3.4 statistic from the paper linked here (a simplified local outlier factor over the PC scores), on a fresh branch so this PR's IBS branch stays intact.

It requires `--pca` and runs inside `CalcPca()` while the scores are still in memory, K defaults to 5, and it applies no threshold, since the paper picks one by inspecting a histogram. The `--grm-maf` half of your comment here went in separately as #459.

PR #456, opened 2026-09-10
Add --tail-pheno

BenjaminDEMAILLE:
`--tail-pheno` from PLINK 1.x, which 2.0 has no equivalent for.

### `--tail-pheno <Lt> [Hbt]`

Downcodes a quantitative phenotype to case/control: above `Hbt` is a case, at or below `Lt` is a control, in between is set to missing. With `Hbt` unspecified it equals `Lt`, so nothing becomes missing.

PLINK 1.x had a single phenotype. With several loaded, every quantitative one is downcoded; phenotypes that are already case/control or categorical are left alone. The case/control bitvector is narrower than the double array it replaces and lives in the same allocation, so the conversion is in place.

### `--must-have-sex`

This started as a second transform and became a rejection on review, since `--remove-nosex` covers the usual intent:

```
Error: --must-have-sex is not implemented.  --remove-nosex drops the same
samples outright, which covers the usual intent; contact us if you need their
genotypes kept with only the phenotypes blanked.
```

The two are not the same operation, which is why the message invites a report rather than closing the door: `--remove-nosex` drops the samples, so they stop contributing to allele frequencies and leave a written fileset, while `--must-have-sex` kept them with their phenotypes blanked. That difference is the only thing worth hearing about.

This also corrects my own audit in #344, which listed `--must-have-sex` under "covered by a rename or modifier" pointing at `--remove-nosex`. The gap was real, just narrower than that entry implied.

### Testing

`Tests/TEST_PHENO_UTILS`, against PLINK 1.9 on a 200-sample `--dummy` fileset with a quantitative phenotype spanning both bounds:

1. `--tail-pheno -1 1.5`, compared value by value
2. `--tail-pheno 0`, and nothing becomes missing
3. `--tail-pheno` leaves an already-binary phenotype byte-identical
4. missing, non-numeric and inverted bounds are rejected, as is `--must-have-sex`

### DIFF
--- a/2.0/Tests/TEST_PHENO_UTILS/.gitignore
+++ b/2.0/Tests/TEST_PHENO_UTILS/.gitignore
@@ -1,0 +1,3 @@
+plink19*
+plink2_*
+tmp_*
--- a/2.0/Tests/TEST_PHENO_UTILS/run_tests.sh
+++ b/2.0/Tests/TEST_PHENO_UTILS/run_tests.sh
@@ -1,0 +1,57 @@
+#!/bin/bash
+
+# --tail-pheno and --must-have-sex, checked against PLINK 1.9.
+
+set -exo pipefail
+
+plink --dummy 200 400 0.03 --out tmp_data > /dev/null
+
+# A quantitative phenotype, and a sex column with a third of the samples unknown.
+awk 'BEGIN{OFS=" "; print "#FID", "IID", "QT"} {print $1, $2, ($NR % 17) - 6 + (NR % 5) * 0.25}' tmp_data.fam > tmp_qt.txt
+awk 'BEGIN{OFS=" "} {$5 = (NR % 3 == 0)? 0 : (1 + (NR % 2)); print}' tmp_data.fam > tmp_sexed.fam
+mv tmp_sexed.fam tmp_data.fam
+
+pheno_col() {
+    # <psam or fam> <column> -> "IID value", with 1.9's -9 normalized to NA
+    awk -v c="$2" '!/^#/ { v = $c; if (v == "-9") { v = "NA" }; print $2, v }' "$1"
+}
+
+# 1. --tail-pheno with both bounds.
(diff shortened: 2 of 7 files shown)


BenjaminDEMAILLE:
Correction to my own audit before you spend time on this. In #344 I listed `--must-have-sex` under "covered by a rename or modifier", pointing at `--remove-nosex`. That was wrong, and I should have caught it while writing this.

They are not the same operation:

- `--remove-nosex` drops the samples. They leave the fileset, and they stop contributing to allele frequencies, missingness rates and everything else computed from the sample set.
- `--must-have-sex` keeps them, with their phenotypes set to missing. They still contribute to allele frequencies and still appear in a written fileset.

For `--make-bed` the two produce different filesets, and for anything frequency-based they produce different numbers. So the gap is real, but it is narrower than the audit entry implied, and it is worth deciding whether it is worth a flag at all given `--remove-nosex` covers the more common intent.

`--tail-pheno` is unaffected by this; there is no 2.0 equivalent.

Happy to split this into two PRs, drop the `--must-have-sex` half, or leave it as is.

chrchang:
As you note, `--tail-pheno` should be the more useful of the two.  `--must-have-sex` can fail for now with a pointer to `--remove-nosex` and a "Contact us if that doesn't cover your use case" note.

BenjaminDEMAILLE:
Done. `--must-have-sex` now errors out:

```
Error: --must-have-sex is not implemented.  --remove-nosex drops the same
samples outright, which covers the usual intent; contact us if you need their
genotypes kept with only the phenotypes blanked.
```

I put the "contact us" on the one thing `--remove-nosex` genuinely does not do, rather than on the flag as a whole, so anyone who reads it knows what would actually be worth reporting.

`PhenoMustHaveSex()` and its transform flag are gone, so this is `--tail-pheno` plus a rejection. Retitled accordingly. `Tests/TEST_PHENO_UTILS` loses the two comparisons and gains a check that the flag is rejected; its fixture no longer needs a sex column. Suite still passes.

PR #458, opened 2026-09-10
Address BCF FORMAT float payloads as bytes

BenjaminDEMAILLE:
Follow-up to the `-DNO_UNALIGNED` question on #445, sent separately since it has nothing to do with the padding there.

### The problem

The BCF dosage parsers take the record pointer as `const float*`:

```c
// plink2_import.cc:5385 and :7011
dosage_main = R_CAST(const float*, dosage_main_raw);
```

BCF FORMAT payloads carry no alignment guarantee, so that pointer can land on any offset. Forming it is undefined behavior, and the loads in `ParseBcfBiallelicGp()` are ordinary indexed float loads, which fault on a target that does not permit unaligned access.

`-DNO_UNALIGNED` does not cover this. It removes the unaligned-access primitives, not an indexed load through a misaligned pointer of a type with an alignment requirement, which is exactly why it is worth checking with the alignment sanitizer rather than assuming.

### How it shows up

Built with `-DNO_UNALIGNED -fsanitize=address,undefined` and, unlike my earlier runs on #445, **without** `-fno-sanitize=alignment`:

```
+ plink2 --bcf db.bcf dosage=GP --make-pgen --out dg
../plink2_import.cc:5217:23: runtime error: load of misaligned address ... for type 'const float *'
../plink2_import.cc:5170:28: runtime error: load of misaligned address ... for type 'const float'
../plink2_import.cc:5174:28: runtime error: load of misaligned address ... for type 'const float'
```

Only the first is reached by `Tests/`, via `--bcf ... dosage=HDS`; the other two need a BCF carrying GP dosages.

### The change

`ParseBcfBiallelicGp()`, `ParseBcfBiallelicDosage()` and `ParseBcfBiallelicHds()` take `const unsigned char*` and copy each float out with `memcpy`, which is what `ParseBcfBiallelicDosage()` already did for its first read. The two casts become plain assignments, and the one piece of pointer arithmetic becomes byte arithmetic.

### Verification

- The same BCF imported with `dosage=GP`, `dosage=DS` and `dosage=HDS` gives byte-identical `.pgen` and `.pvar` before and after.
- Under the sanitizer build above, the BCF GP import goes from 4 reports to 0, with an identical `.pgen`.
- `Tests/run_tests.sh` passes on the ordinary build (1859 lines, `All tests passed.`).

I was not able to run the whole suite under `-DNO_UNALIGNED` plus the alignment check *on this branch alone*, because it stops at the first `--keep` on the unrelated issue #445 addresses. With #445 applied on top locally it gets through, and the `plink2_import.cc:5217` report is gone.

### DIFF
--- a/2.0/plink2_import.cc
+++ b/2.0/plink2_import.cc
@@ -5161,17 +5161,29 @@ BcfParseErr BcfScanGt(const BcfImportContext* bicp, const unsigned char* gt_star
 
 // Dosage parsing is messy enough that we stick as closely as possible to the
 // VCF logic, instead of trying to vectorize.
-BoolErr ParseBcfBiallelicGp(const float* cur_gp_start, uint32_t is_haploid, double import_dosage_certainty, DosageParseResult* dpr_ptr, double* alt_dosage_ptr) {
+// BCF FORMAT payloads are not aligned within the record, so the float vectors
+// are addressed as bytes and each value is copied out.  Casting the record
+// pointer to const float* and indexing it is undefined behavior, and on a
+// target that does not permit unaligned loads it faults; -DNO_UNALIGNED does
+// not help, since the access is an ordinary indexed load rather than one of
+// the primitives that flag controls.
+HEADER_INLINE float BcfFloatAt(const unsigned char* base, uintptr_t idx) {
+  float fval;
+  memcpy(&fval, &(base[idx * sizeof(float)]), sizeof(float));
+  return fval;
+}
+
+BoolErr ParseBcfBiallelicGp(const unsigned char* cur_gp_start, uint32_t is_haploid, double import_dosage_certainty, DosageParseResult* dpr_ptr, double* alt_dosage_ptr) {
   // See ParseVcfBiallelicGp().
   // P(0/0), P(0/1), P(1/1), etc.
   // assumes dpr initialized to kDosageParseOk
   // assumes *gp_iter is not missing
   // returns 1 if missing OR parsing error.  error: dpr still kDosageParseOk
-  const float prob_0altf = cur_gp_start[0];
+  const float prob_0altf = BcfFloatAt(cur_gp_start, 0);
   if (unlikely((prob_0altf < S_CAST(float, 0.0)) || (prob_0altf > S_CAST(float, 1.0)))) {
     return 1;
   }
-  const float prob_1altf = cur_gp_start[1];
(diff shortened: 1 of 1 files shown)


BenjaminDEMAILLE:
Follow-up now that this is in: I swept the rest of the class rather than leaving it at the one instance, since `plink2_import.cc` has 29 casts of a record pointer to an alignment-requiring type and the question was which of them actually receive a misaligned pointer.

Same build as before, `-DNO_UNALIGNED -fsanitize=address,undefined` with the alignment check on. Everything below reports 0:

- **BCF QUAL/FILTER/INFO readers**, which is where most of those 29 casts live. Twelve fixtures whose QUAL and INFO integers span the int8, int16 and int32 encodings plus float QUAL, each imported, exported to BCF and re-imported. Confirmed the values actually round-trip rather than being dropped, so the readers were exercised: `QUAL=1.15904e+06`, `FILTER=q10`, `INFO=NS=30;BIG=30200829;AF=0.0472;STR=abc6;FLG` all survive.
- **BGEN**, which I expected to be the next most likely: export and re-import at 1.1, 1.2 and 1.3, plain and with forced dosages.
- Import: `--import-dosage`, `--haps`, `--gen`, `--bfile`, plain `--vcf`, plain `--bcf`, `--pmerge`.
- Export and analysis: `haps`, `hapslegend`, `oxford`, `A-transpose`, `vcf` with HDS, `--make-bed`, `--validate`, `--pgen-diff`, `--make-king-table`, `--glm`.

And the full `Tests/run_tests.sh` under that configuration was clean apart from the report this PR fixed.

So the float dosage path looks like the only one that receives a misaligned pointer in practice; the other casts are on payloads that happen to land aligned. Not a proof, since a reader only faults on inputs that reach it, but the obvious candidates are covered.

BenjaminDEMAILLE:
Confirmed on merged master (278e9df0), which now carries both this and #445: built with `-DNO_UNALIGNED -fsanitize=address,undefined` and the alignment check enabled, `Tests/run_tests.sh` reaches `All tests passed.` with **zero** sanitizer reports across 1859 log lines.

That is the configuration you asked about on #445, and it is now a clean baseline rather than one that stops at the first `--keep` or trips on the BCF dosage path. Worth running alongside the default build rather than instead of it, for the `#ifndef NO_UNALIGNED` coverage reason you gave.

PR #459, opened 2026-09-10
Add --grm-maf, and require it only when it matters

BenjaminDEMAILLE:
Implements the `--grm-maf` design from your comment on #455, and the `--homozyg-maf` half of it.

### `--grm-maf` (alias `--grm-min-af`)

A frequency floor for `--pca` and GRM construction alone; every other command's variant set is untouched. It takes the same mode parameter `--maf` does and goes through the same `EnforceFreqConstraints()` path, so multiallelic handling comes along rather than being reimplemented.

A threshold below a quarter of the inverse square root of the sample size requires `yes-really`:

```
Error: --grm-maf 0.001 is below 0.0125, a quarter of the inverse square root of
the sample size, where --pca/GRM construction becomes unreliable. Add
'yes-really' to --grm-maf if that is what you want.
```

Without the flag, a variant below that point is an error rather than a silently unstable answer, and the message reports the lowest frequency actually present:

```
Error: --pca/GRM construction is unreliable at very low minor allele
frequencies, and the lowest remaining here is 0, under 0.0125 (a quarter of the
inverse square root of the 400 samples). Use --grm-maf <freq> to set a floor
for this calculation alone, or "--grm-maf <freq> yes-really" to proceed anyway.
```

Writing this turned up a second bug, fixed here on review: `--maf`, `--max-maf`, `--mac` and `--max-mac` all scanned their threshold with `ScantokDouble()`, which only succeeds when the number ends the token. That made each one's `':'` branch unreachable, so the bcftools `freq:mode` notation their help documents was rejected outright: `--maf 0.05:minor` gave `Invalid --maf mode '0.05:minor'`. All four now use `ScanadvDouble()`, which is what those branches were written for, and `--grm-maf` takes the notation too.

A monomorphic variant is rejected whatever the threshold, with its own message rather than the instability one, since there is no minor allele to standardize by and no `yes-really` should reach it:

```
Error: --pca/GRM construction cannot use monomorphic variants, and 6 of the 600
being scanned carry no minor allele. Exclude them, with any positive --grm-maf
threshold or with --mac 1.
```

That check runs after the `--grm-maf` filter rather than before it, so `--grm-maf 0 yes-really` hits it as well. "No minor allele" is read as a frequency below half an allele copy rather than exactly zero: plink2 computes frequencies by multiplying by a reciprocal, so a monomorphic variant's nonmajor frequency can come back as an ulp rather than 0, and a singleton sits two allele copies above that line.

### `--homozyg-min-af`

Same idea from the other side. It was required unconditionally, decided at parse time before any frequency was known. It is now required only when the data still holds a variant below 0.05, so an already-common variant set needs no floor. The check moves to the point where the frequencies exist, and names the lowest one.

### Testing

`Tests/TEST_GRM_MAF`, on a 400-sample simulated fileset half of whose variants sit below the instability point:

1. the monomorphic error and that `yes-really` does not bypass it, then the instability error behind `--mac 1`, each checked for its own wording
2. a common-only fileset needing no flag
3. `--grm-maf 0.05` running, and reporting what it dropped
4. `--grm-min-af` being the same flag
5. `--freq` still seeing all 1000 variants while `--pca` sees 500
6. **the result matching `--maf 0.05` applied globally**, which is what "applied just to --pca" has to mean
7. `yes-really` required below the floor, and sufficient
8. the mode parameter, plus four malformed arguments rejected
9. `--make-rel` and `--make-grm-list` gated identically
10. a threshold that removes everything erroring rather than crashing

`Tests/TEST_HOMOZYG`'s assertion that `--homozyg` fails without `--homozyg-min-af` becomes two: it succeeds on the all-common fixture and matches an explicit floor of 0, and still fails on a fixture carrying rare variants. That test is what caught the behaviour change, which is the point of it.

Full suite passes.


chrchang:
Oops, yes, `--maf` parser should be based on `ScanadvDouble` rather than `ScantokDouble`, feel free to fix that in this PR.

In the meantime, the no-`--grm-maf` error message should be different when there's a MAF-0 variant in the dataset, since even `yes-really` shouldn't accept that.

BenjaminDEMAILLE:
Both done.

**The freq:mode parsers.** Switched `--maf`, `--max-maf`, `--mac` and `--max-mac` to `ScanadvDouble()`. That makes each one's existing `':'` branch reachable for the first time, so the notation the help documents now parses:

```
$ plink2 --bfile r --maf 0.05:minor --write-snplist
1600 variants remaining after main filters.      # was: Invalid --maf mode '0.05:minor'
```

`--grm-maf` takes it too, and the space-separated form and the malformed cases behave as before: `--maf 0.05abc` is still rejected, `--maf minor` still means the 0.01 default with a mode, bare `--maf` still means 0.01.

**Monomorphic variants.** You are right that `yes-really` should not reach those, and it no longer can. The scan counts them separately, in whatever variant set the calculation will actually see, and they get their own message:

```
Error: --pca/GRM construction cannot use monomorphic variants, and 6 of the 600
being scanned carry no minor allele. Exclude them, with any positive --grm-maf
threshold or with --mac 1.
```

Because that check runs *after* the `--grm-maf` filter rather than before it, `--grm-maf 0 yes-really` hits it as well, which is the case that mattered.

One implementation note: "no minor allele" is read as a frequency below half an allele copy rather than exactly zero. plink2 computes frequencies by multiplying by a reciprocal, so a monomorphic variant's nonmajor frequency can come back as an ulp rather than 0, and an exact comparison would miss it. A singleton is two allele copies above that line, so nothing real is caught by it.

`Tests/TEST_GRM_MAF`'s first step splits into the two errors: the simulated 0.0005-0.01 fixture reliably contains monomorphic variants, so it asserts that message and that `yes-really` does not bypass it, then re-runs behind `--mac 1` to reach the instability error. Full suite passes.

PR #460, opened 2026-09-10
Add --fast-epistasis

BenjaminDEMAILLE:
On #344 I said I would not finish `boost` and `joint-effects` without your word, and I have not. But the default test has been finished and verified for a while, and holding it back only makes the decision more expensive for you than a PR you can close in one click. **If epistasis is not wanted in 2.0, close this and I will not send the rest.**

Unlike the `--genome` PR I sent by mistake this morning, this is not a format-compatibility port: `--fast-epistasis` and `--epistasis` are section A of #344, a capability with no substitute in 2.0, and an interaction scan is not something a user reconstructs from other output.

**Update, after your answer:** `boost` is now here too, in a second commit. `joint-effects` is not, and I will not send it, since you said the other approaches have not held up. The quantitative-phenotype `--epistasis` with covariates is a separate piece of work and will be a separate PR.

### What this is

PLINK 1.07's allele-based test on the 3x3 joint genotype table, with the variance and empty-cell corrections from Ueki M, Cordell HJ (2012) *Improved statistics for genome-wide interaction analysis*. `no-ueki` drops the corrections for the original statistic, `case-only` uses the cases alone and writes `.epi.co`, and `--gap` then skips same-chromosome pairs close enough for LD rather than interaction to explain them.

`boost` replaces that test with the two-stage one of Wan X et al. (2010) *BOOST: A fast approach to detecting gene-gene interactions in genome-wide case-control studies*: every pair is scored with the Kirkwood superposition approximation, which is closed-form, and only the pairs clearing `--epi1` are then fit by iterative proportional fitting of the homogeneous association model. So `--epi1` decides which pairs are tested at all rather than only which are printed, and its default is 5e-6 rather than 1e-4. An empty genotype row or column costs two degrees of freedom instead of dropping the pair, so df is 4, 2 or 1 and the report gains a DF column; a second empty row (or column) leaves nothing to test, and those pairs are dropped from the report and from the summary's N_TOT. As in 1.9, `boost` rejects `case-only` and `no-ueki`.

Still not included: `joint-effects` and `--je-cellmin`, and the `set-by-set`/`set-by-all` modifiers, which need variant sets. All are rejected with a message saying so rather than silently ignored.

### Implementation

The per-pair 3x3 counts come from three bitvectors per variant per group (hom-REF, het, hom-ALT), so a cell count is one `PopcountWordsIntersect`. Variants are held in two blocks, rows and columns; the column block is sized to hold the whole range when it fits, and when it does not the row block drops to a single row, because the report has to be row-major for concatenated `--parallel` jobs to reproduce a single run.

Which group must be polymorphic follows 1.9 rather than intuition: with the Ueki correction the +4.5 adjustment absorbs empty cells, so only variants monomorphic across *every* phenotyped sample are dropped; with `no-ueki` each group has to be polymorphic on its own. Getting this wrong is what first made the two disagree on which pairs exist.

`boost` reuses all of that and adds the two stages. 1.9 caches the marginal probability vectors when the row variant has no missing calls; this recomputes them from the pair's own 3x3 counts instead, which is arithmetically the same thing (the cached and recomputed values are ratios of the same integers) and a good deal less state to get wrong. The screening stage is cheap enough that it does not show up next to the popcounts.

### One deliberate difference

When a pair's statistic is undefined (an empty allele-table cell under `no-ueki`, a table `boost` cannot fit), 1.9 writes a report row with `STAT=nan`, but only when `--epi1` is permissive enough, and its own `.summary` excludes that pair from `N_TOT`. This omits the row, which is what the summary already says happened. It is a small internal inconsistency in 1.9; say if you would rather match it.

### Testing

`Tests/TEST_EPISTASIS`, against PLINK 1.9 on a 300-sample, 600-variant simulated fileset over three chromosomes at 400 kb spacing:

- the default test and `no-ueki`, all 179700 pairs
- `boost`, all 179700 pairs: identical STAT and DF on every one
- `boost` at four `--epi1` thresholds including its default, which change which pairs are fit rather than only which are printed: identical report sets and identical `.summary` each time
- a second fileset of rare variants with missing calls, where empty genotype rows and columns pull the degrees of freedom down and throw some pairs out: the `.summary` matches there too, `N_TOT` included, which is what checks that both programs throw out the same pairs
- `case-only` with and without the corrections
- `--gap` at 0, 900 and 2000 kb, which take the pair count from 179700 to 178509 to 177330, matching 1.9 at each
- the `.summary`'s `N_SIG`, `N_TOT` and `BEST_CHISQ` across all 600 variants
- `--epi1` filters the main report without changing the summary's `N_TOT`
- `--threads 1` reproduces the default run, `--parallel` chunks concatenate to it, `zs` round-trips, `nop` drops the p-value column
- the unimplemented tests and the flag dependencies are rejected

STAT is compared at 1.9's four-significant-digit print precision rather than a fixed tolerance. The exception is a statistic that is mathematically zero: it is rounding noise in both programs, and on these degenerate tables it is common, so those rows are compared against an absolute noise floor instead. That also decides membership when `--epi1 1` puts `boost`'s screening threshold at zero and the sign of the noise decides which side of it a pair falls on.

Beyond the checked-in test, 150 fuzz runs over randomly generated filesets across all four mode combinations agreed except for two rows where a tie in `BEST_CHISQ` named a different variant.


chrchang:
* Looking at recent reviews of the subject, the methods most worth preserving are regular --epistasis for quantitative phenotypes, and `--fast-epistasis boost` for case/control phenotypes.  (The other --fast-epistasis approaches have not held up as well.)
* Efficient support for covariates should be added to quantitative-phenotype --epistasis, in a similar manner to how it was done for --glm (which is much faster than its PLINK 1.9 counterpart).
* The logistic-regression side of --epistasis (also with added covariate support) might be worth implementing later.
* If some GPU kernels are added later, this is one of the commands that should be affected.

BenjaminDEMAILLE:
Understood, thanks. `boost` is in this PR now, in a second commit, verified against 1.9 over all 179700 pairs of the test fileset (identical STAT and DF), at four `--epi1` thresholds including its default, and on a second fileset of rare variants with missing calls where empty genotype rows and columns pull the degrees of freedom down and throw pairs out. `joint-effects` I will leave alone.

Quantitative-phenotype `--epistasis` with covariates is what I am building next, as a separate PR. Since 1.9 has no covariate support there, only the no-covariate case can be checked against it; the covariate case I will check against an independent oracle, and against the identity that a covariate uncorrelated with everything must leave the interaction statistic where it was.

The logistic side and GPU kernels I will leave until you say so.

BenjaminDEMAILLE:
One more commit here: the block sizing capped `max_slot_ct` at `variant_ct`, which made the branch that holds the whole column range in one block unreachable, so the row block always fell back to a single row and every row reread every column variant. Output is unchanged and the suite still matches 1.9 pair for pair.

Quantitative `--epistasis` with covariates is now up as #468, which sits on top of this branch.

chrchang:
One clarification: I don't think the default (without `boost`) `--fast-epistasis` method is worth keeping.  So it's reasonable to rename the flag to `--epistasis-boost`, and the command-line parser could error out on `--fast-epistasis` without `boost` while treating `--fast-epistasis` with `boost` identically to `--epistasis-boost`.

`set-by-set` and `set-by-all` are probably worth keeping, though they don't need to be implemented in the initial PRs.

Reviewing the BOOST paper, it notes that the method could support covariates by using regular logistic regression in the testing stage; so we should plan to add that logic in a later PR.

BenjaminDEMAILLE:
Done, in a fourth commit.

`--epistasis-boost` is now the flag. The parser rejects `--fast-epistasis` without `boost` and, with it, routes to exactly the same code, modifiers included, so a 1.9 command line for the surviving test keeps working.

Rejecting the default test left the 1.07 statistic, the Ueki and Cordell corrections, `case-only` and `--gap` unreachable, so they are gone rather than dead: `FepiCountsToStats()`, the group-by-group monomorphic-variant pass it needed (boost only needs the whole phenotyped set to be polymorphic), the `.epi.co` output path, and the branches that made the DF column and the chi-square thresholds conditional. `case-only`, `no-ueki` and `joint-effects` now report that they belonged to a retired test; `set-by-set` and `set-by-all` keep reporting that they need variant sets, so they are ready to be filled in when the set code lands.

The test suite drives `--epistasis-boost` against 1.9's `--fast-epistasis boost` over the same 179700 pairs as before, plus the rare-variant fileset where the degrees of freedom drop, and adds the synonym, the retirement rejections and the loss of `--gap`. Rebased onto master, clean build, suite green.

On covariates: the paper's conclusion puts the logistic regression after the two stages rather than inside the testing stage.

> If the covariate is discrete or can be discretized, our method can be directly extended to handle it. If not, logistic regression can be used in the postprocessing step to adjust the covariate.

> In the postprocessing step, the computational burden of logistic regression is affordable because the number of selected interactions is limited.

That reading is convenient for us, since it leaves the bitwise screen and the IPF fit untouched: the covariate pass only has to run over the pairs that already cleared --epi1, so it can reuse --glm's logistic machinery without the per-pair cost mattering. The equivalence the paper relies on for that is the one in Methods, that the homogeneous log-linear model is the equivalent form of the main-effect logistic model and the saturated one of the full model, so the 4-df statistic is the same quantity either way and the no-covariate case stays checkable against 1.9. Discrete covariates have the cheaper route of stratifying the 2x3x3 table instead, at two more degrees of freedom per level. I will do this in its own PR, after the quantitative-phenotype one.

#468 sits on top of this branch and is rebased onto the rename.

BenjaminDEMAILLE:
Follow-ups to your two notes here, all stacked on this branch:

- **#474** covariates for `--epistasis-boost`, as the postprocessing step the paper describes.
- **#475** `--set`/`--set-names`/`--set-collapse-all`/`--write-set`, since 2.0 had no way to load a variant set.
- **#476** `set-by-set` and `set-by-all`, on top of #475. All three modes match 1.9 pair for pair.

#475 has two open questions in its description (whether `--subset` and `--make-set` should come along) that are worth settling before the rest, since everything else in that PR follows from the format and the storage.

chrchang:
Looks like I garbled the merge; will fix now.

chrchang:
Test currently fails when initial `plink --simulate` seed is 1789604545:

```
+ awk -f compare.awk plink19.epi.cc plink2.epi.cc
STAT differs on common_131|common_406: 0.655775 vs 0.655774
pair missing from the plink2 report: common_190|common_230
```

BenjaminDEMAILLE:
Thanks for the seed. Neither message came from plink2. For common_131 x common_406, PLINK 1.9 computes 0.65577450000006 and plink2 computes 0.65577449999999, which straddle the rounding boundary and print as 0.655775 and 0.655774; compare.awk only allowed for rounding on one side. The "pair missing" line was a side effect: awk runs END after the first exit, and END scanned a report read only up to the failing row. #487 fixes both. It passes on your seed and on 200 other random seeds; the old script failed on 3 of those 201.

PR #474, opened 2026-09-12
Add covariate support to --epistasis-boost

BenjaminDEMAILLE:
Stacked on #460 and #468; the covariate commit is the last one. This is the "later PR" from your note on #460.

## What the paper says

Its conclusion, not the Methods:

> If the covariate is discrete or can be discretized, our method can be directly extended to handle it. If not, logistic regression can be used in the postprocessing step to adjust the covariate.

> In the postprocessing step, the computational burden of logistic regression is affordable because the number of selected interactions is limited.

So the regression goes *after* both stages rather than replacing the testing stage. The bitwise screen and the IPF fit are untouched, and only the pairs that clear `--epi1` are refit, on genotype dummies for both variants, their interaction, and the covariates. `STAT`/`DF`/`P` come from that fit. At the default `--epi1` that is a handful of pairs, so the command keeps its speed.

## One thing that is not a free extension

The obvious design, main-effect dummies plus product dummies, is not always full rank: a genotype combination with no samples leaves its product term with nothing to estimate. That is not a corner case, 42% of the pairs of the 600-variant, 300-sample test fileset have such a cell, and the first version of this dropped 28% of its pairs on singular fits.

The full model is instead saturated over the *occupied* cells, which is what the log-linear saturated model is too.

The reported degrees of freedom are the table's rather than the fit's, per review: a sampling zero costs the full model a parameter but not the table a degree of freedom, so `DF` is `(rows - 1) * (columns - 1)` over the genotype levels that are present. A pair with an empty interior cell is therefore tested at 4, as on the unadjusted side, and only an empty row or column, which drops a level outright, gives the existing 2 and 1.

## Also

- `BEST_CHISQ` and `N_SIG` stay on the screening statistic, since they cover pairs that were never refit.
- A sample missing a covariate is dropped from the whole scan, not just the refit, so the screen and the refit see the same samples, as `--glm` does.
- A constant covariate is dropped with a warning; with none left, the screen-and-fit path runs unchanged.
- A pair whose fit does not converge is left out of the report and counted in a warning.
- Categorical covariates are rejected for now, pointing at `--split-cat-pheno`.
- `LogisticRegressionD()` is exposed in the header for this; nothing else about `--glm` changes.

## Verification

Two independent checks.

Forcing the refit with no covariate at all must reproduce the log-linear statistic, which is the correspondence the paper's Methods rely on. It does, to five significant digits over the 127804 pairs where the degrees of freedom agree, and no pair is dropped.

With covariates there is no 1.9 to compare against, so the new `TEST_EPISTASIS` section checks `STAT` and `DF` against `oracle_covar.py`, which fits the same two models under a different coding (last level and last cell as references, where plink2 takes the first of each). 190 of 190 pairs agree, `DF` exactly and `STAT` to 4e-6. Standard library only, no numpy.

The section also pins the behaviour: the adjusted `DF` equals the unadjusted one on every pair, and a refit of one of the pairs the unadjusted scan gives fewer than 4 comes back at the same 2, the constant-covariate path reproduces the unadjusted report byte for byte, a sample missing a covariate leaves the same `N_TOT` as `--remove`ing it, and a categorical covariate is rejected.

Clean build, `TEST_EPISTASIS` and `TEST_EPISTASIS_QT` green.


chrchang:
* I will update the LnPToChisq() function in include/plink2_stats to take a second df parameter.
* When there is a "sampling zero" but no row/column of the contingency table is empty (the df=3 logistic regression case you mention), we should just use df=4 in the p-value computation.

BenjaminDEMAILLE:
Done: the refit now reports the table's degrees of freedom rather than the fit's. A sampling zero still costs the full model a parameter, since that cell has nothing to estimate, so the model stays saturated over the occupied cells and `STAT` is unchanged; `DF` is now `(rows - 1) * (columns - 1)` over the genotype levels that are present. So the df=3 case becomes 4, and only an empty row or column, which drops a level outright, gives 2 or 1.

That makes the adjusted degrees of freedom equal the unadjusted ones on every pair, which is what TEST_EPISTASIS now pins, along with a refit of one of the pairs the unadjusted scan gives fewer than 4 (it comes back at 2). `oracle_covar.py` follows the same rule, and its 190 pairs still agree on `STAT`.

On `LnPToChisq()`: the only thing on my side that wants it is `FepiChisqThresh()` in plink2_ld.cc, which currently bisects on `ChisqToLnP()` to turn `--epi1`/`--epi2` into chi-square thresholds for df 4, 2 and 1. I will drop it for `LnPToChisq(ln_alpha, df)` once your change lands, in #460 where it was added.

chrchang:
LnPToChisq() is now updated.

BenjaminDEMAILLE:
Thanks. Rebased onto master (through 3f62faca), which is what this needed: #460 landed upstream as 09947e0b + 8f8160f6, so the four `--fast-epistasis`/`--epistasis-boost` commits this branch was carrying are gone from it, and only the covariate work remains.

No change was needed for the new `LnPToChisq()` signature: this code goes the other way, through `ChisqToLnP(stat, df)`, which already took the degrees of freedom. `TEST_EPISTASIS` and `TEST_EPISTASIS_QT` both pass on the rebase.

#468 is rebased the same way and is now just the quantitative-phenotype commits on top of master.

PR #475, opened 2026-09-12
Add --set/--make-set and the rest of PLINK 1.x's set-manipulation flags

BenjaminDEMAILLE:
Groundwork for `--epistasis-boost set-by-set`/`set-by-all`, which you said are worth keeping on #460. 2.0 has no way to load a variant set today, so they need this first; it is a separate PR because nothing here is epistasis-specific and several other 1.x commands want the same thing.

Now covers the whole set-manipulation group, per your review: `--set`, `--make-set` (+ `--make-set-border`, `--make-set-collapse-group`), `--set-names`, `--subset`, `--set-collapse-all`, `--complement-sets`, `--make-set-complement-all`, `--make-set-complement-group`, `--write-set`, `--set-table`, `--gene` and `--gene-all`.

## Design

A `--set` file is 1.x's: a stream of whitespace-delimited tokens, each set being its name, then its variant IDs, then `END`. An ID that is not in the filtered variant set is ignored and a set may be empty, as in 1.x.

`--make-set` reuses `LoadIntervalBed()`, the existing port of 1.9's `load_range_list()` that already backs `--extract range` and `--clump-range`. It regains the column-5 group label that 1.9 read and the 2.0 port dropped, since `--make-set-collapse-group` builds its sets from that column; every other caller passes `collapse_group = 0` and is unaffected. The ranges come back as variant uidx intervals, so `DefineSets()` only has to move them into the filtered variant space, which it does by filling the same member bitvector the `--set` path fills. Overlapping ranges then merge for free, instead of needing 1.9's sort-and-merge pass.

Storage uses the setdef encoding `plink2_set.h` already documents for `LoadAndSortIntervalBed()`. `SaveSetdef()` picks whichever of the two forms is smaller for the set at hand: eight bytes per run of consecutive members against sixteen bytes plus a bit per variant between the first and last member. `--complement-sets` inverts the member bitvector before it is saved, so it needs no separate representation.

`--subset` and `--set-names` become one sorted, deduplicated list, so both select sets by binary search; `--make-set` matches that list against the set ID in column 4 even when the sets themselves are column-5 groups, as in 1.9.

`--set-table` walks the setdefs with a per-set cursor rather than unpacking them all, so its memory is proportional to the set count rather than set count times variant count.

Setdefs index the filtered variant space, as in 1.x, so `DefineSets()` runs after every variant filter has been applied and builds its own variant-ID hash table there rather than borrowing the one the early filters use. The table is freed before the setdefs are written, so it costs nothing once loading is done. The consequence is stated in the help and tested: a filter that drops a variant drops it from every set too.

## Differences from 1.9

Both are on the `--set` + `--complement-sets` path, and both are 1.9 bugs rather than choices:

- 1.9 applies the `C_` name prefix its help text promises only to `--make-set` sets. Its own `--gene` lookup then cannot match a complemented `--set` set, since that lookup keeps only `C_`-prefixed names. 2.0 applies the prefix on both paths.
- 1.9's `save_set_bitfield()` starts its first complement range at index 1 rather than 0, so it drops the first variant from every complement of a set that does not contain it. For example, with a set whose only member is `sv_5`, 1.9's `--write-set` output for the complement starts at `sv_1`.
- 1.9's `--gene` with `--complement-sets` intersects its exclusion mask at every set, including ones `--gene`/`--subset` skipped, so a single skipped set makes it keep every variant.

All three are fixed for 1.9 in #478.

## --gene/--gene-all

Now here too. They are variant filters, so `GeneFilter()` runs in the filter stage: it loads the sets over the current variant set, unions the ones it was asked for, clears everything else out of `variant_include`, and throws those setdefs away; `DefineSets()` rebuilds them later over the variants that survived, exactly as 1.9 rereads its `--set` file after the same pre-pass. Reading the file twice is the price of keeping setdefs indexed over the filtered variant space.

`--complement-sets` needs no special case, since the inversion is already in the setdefs: a variant survives iff it is outside at least one named set. `--gene` takes the names `--write-set` reports, so a complemented set needs its `C_` prefix.

## Verification

`TEST_SETS` diffs 2.0's `--write-set` and `--set-table` output directly against 1.9's, since the formats are the same, for `--set`, `--subset`, `--set-names`, `--make-set`, `--make-set-border`, `--make-set-collapse-group`, `--complement-sets`, `--make-set-complement-all`, `--make-set-complement-group` and `--set-collapse-all`. They agree in every case; the one exception is `--set` + `--complement-sets`, where the test checks 2.0 against a complement computed in the script, for the two reasons above.

`--gene`/`--gene-all` are checked the same way, by diffing the surviving variant list against 1.9's for `--set` and `--make-set`, plus a check that the sets the later commands see cover only the variants `--gene` kept; their complement case is checked against expected output instead, for the `--gene` defect above.

It also covers the setdef forms (a scattered set, an adjacent pair carrying an unknown ID, an empty set, a fifteen-variant run), `--set-collapse-all`'s deduplicated union, that `--exclude` takes variants out of the sets, `zs` for both reports, and the rejections: an unmatched `END`, a set with no `END`, a range line missing its set ID, an unknown chromosome code, `--make-set-collapse-group` on a four-column file, selections that match nothing, an empty `--subset` file, and the flag dependencies and conflicts.

Clean build, suite green, and `TEST_SETS` also clean under ASan+UBSan.


chrchang:
Generally looks fine.
* We may as well port the rest of PLINK 1.x's set-manipulation flags.
* Can you clarify what you mean by "--write-set deliberately writes the same format it reads, rather than 1.9's --write-set output"?  Formats look the same to me, but I may be missing something.

BenjaminDEMAILLE:
> Can you clarify what you mean by "--write-set deliberately writes the same format it reads, rather than 1.9's --write-set output"?

Nothing to clarify: that sentence is wrong, and I have dropped it from the description (and from the comment at the top of TEST_SETS). 1.9's `write_set()` list branch writes the set name, one member ID per line, `END`, blank line, which is exactly what this PR writes:

```
$ plink --bfile tmp_data --set sets.txt --write-set --out p19
$ od -c p19.set
0000000    g   e   n   e   A  \n   s   v   _   0  \n   s   v   _   1  \n
0000020    s   v   _   2   0  \n   E   N   D  \n  \n   g   e   n   e   B
0000040   \n   s   v   _   2  \n   s   v   _   3  \n   E   N   D  \n  \n
0000060    g   e   n   e   C  \n   E   N   D  \n  \n
```

I had conflated `--write-set` with `--set-table`/`--write-set-r2`, whose output really is a different shape. Since the formats do match, TEST_SETS will now diff 2.0's `.set` against 1.9's directly instead of only round-tripping.

Porting the rest of the set-manipulation flags now: `--subset`, `--make-set` with `--make-set-border`/`--make-set-collapse-group`, `--complement-sets`/`--make-set-complement-all`/`--make-set-complement-group`, and `--set-table`. `--gene`/`--gene-all` are variant filters rather than set manipulation, and they need the set loader to run in the filter stage rather than at command time, so I plan to leave those for a separate PR unless you want them folded in here.

BenjaminDEMAILLE:
Pushed: `--make-set` (+ `--make-set-border`, `--make-set-collapse-group`), `--subset`, `--complement-sets`, `--make-set-complement-all`, `--make-set-complement-group` and `--set-table`. Description updated with the design; the short version is that `--make-set` reuses `LoadIntervalBed()`, which gets back the column-5 group label that 1.9's `load_range_list()` read (other callers pass `collapse_group = 0`), and that the ranges only need moving into the filtered variant space, so the same member bitvector the `--set` path fills does the rest.

Since `--write-set` writes 1.9's format, TEST_SETS now diffs against 1.9 directly for `--set`, `--subset`, `--set-names`, `--make-set`, `--make-set-border`, `--make-set-collapse-group`, `--complement-sets`, `--make-set-complement-all`, `--make-set-complement-group` and `--set-collapse-all`, for both `--write-set` and `--set-table`. All agree, with one exception, which is a 1.9 bug:

`save_set_bitfield()`'s complement branch starts its first range at `bit_idx + 1` rather than `bit_idx`, so the first variant is dropped from every complement of a set that does not contain it:

```
$ printf 'solo\nsv_5\nEND\n' > solo.txt
$ plink --bfile tmp_data --set solo.txt --complement-sets --write-set --out b19
$ head -4 b19.set
solo
sv_1      # sv_0 is missing
sv_2
sv_3
```

Related: on that same path 1.9 does not apply the `C_` prefix its help text promises (only `--make-set` names get it), which also means its `--gene` lookup, which keeps only `C_`-prefixed names when complementing, can never match a complemented `--set` set. 2.0 prefixes on both paths and keeps the first variant, so that one case is checked against a complement computed in the test script rather than against 1.9. I can send a 1.9 fix for both as a separate PR if you want it.

`--gene`/`--gene-all` are the one part I left out: they are variant filters, so they need the set loader to run in the filter stage rather than at command time. Say the word and they go in next.

chrchang:
Yes, you are welcome to post a fix to plink 1.9 --complement-sets .

BenjaminDEMAILLE:
1.9 fix posted as #478. It grew to four defects once --gene could match complemented --set sets at all: the dropped first variant, the missing 'C_' prefix, --gene + --complement-sets keeping everything as soon as one set was skipped, and a two-bit-short sentinel buffer. Non-complement output is byte-identical to master's throughout.

--gene/--gene-all for 2.0 are next, in this PR.

BenjaminDEMAILLE:
--gene/--gene-all are in as well, so the whole 1.x set group is now covered.

They are variant filters, so `GeneFilter()` runs in the filter stage: it loads the sets over the current variant set, unions the ones named, clears the rest out of `variant_include`, and discards those setdefs; `DefineSets()` rebuilds them later over what survived, the same shape as 1.9's pre-pass and reread. `--complement-sets` needs no special case, since the inversion is already in the setdefs.

TEST_SETS diffs the surviving variant list against 1.9's for `--gene`/`--gene-all` over both `--set` and `--make-set`, and checks that the sets the later commands see cover only what `--gene` kept. The complement case is checked against expected output rather than 1.9, for the third defect in #478.

PR #476, opened 2026-09-12
Add --epistasis-boost set-by-set and set-by-all

BenjaminDEMAILLE:
The two set modes from your note on #460. Stacked on #460 (the rename) and #475 (`--set`); the last commit is the only new work.

## What changed in the scan

It was a triangle over one contiguous index range. It is now a rectangle of a row list against a column list, and the triangle is the special case: without a set mode both lists are every variant and the column sweep still starts at the row block, so the default path does exactly the work it did before. One set scans the pairs inside it (still a triangle, just smaller), two sets scan every ordered pair across them, and `set-by-all` scans one set against every variant. A variant's pair with itself is dropped, as in 1.9.

`--parallel` splits the row list: the mirrored triangle partition when the scan is triangular, an even split when it is not, since every row then carries the same number of pairs.

Set membership is over the variant space plink2 hands `CalcEpi()`, while its own indexes are over the polymorphic autosomal subset of that, so the membership test goes back through the raw variant index. A variant filter therefore takes its variants out of the sets as well; the suite pins that.

## Verification

1.9 has all three modes, so this is checked against it rather than against an identity:

| mode | pairs | result |
|---|---|---|
| `set-by-set`, one 40-variant set | 780 | STAT and DF match 1.9 |
| `set-by-set`, 40 against 30 | 1200 | STAT and DF match 1.9 |
| `set-by-all`, 40 against 600 | 23960 | STAT and DF match 1.9 |

23960 rather than 24000 is the 40 self-pairs, which 1.9 also drops. The `--parallel` chunks concatenate to the single-job report in both modes, and 1.9's first chunk comes out the same size as ours.

Also rejected, with tests: a set mode with no `--set`, both modes together, `set-by-all` with more than one set, and `set-by-set` with more than two.

Clean build, `TEST_EPISTASIS` and `TEST_SETS` green.


BenjaminDEMAILLE:
Rebased. #460 landed upstream (09947e0b + 8f8160f6), so the four `--fast-epistasis`/`--epistasis-boost` commits this branch was carrying are gone from it, and the only new commit here is `set-by-set`/`set-by-all`.

It still stacks on #475, since it needs `VariantSets`/`DefineSets`: the three commits below the top one are that PR's, and the diff shown here will shrink to one commit once #475 is merged. `CalcEpi()` takes a `const VariantSets*`, and `--set`/`--make-set` is now required when `set-by-set` or `set-by-all` is given.

Both `TEST_EPISTASIS` and `TEST_SETS` pass on the rebase.

PR #479, opened 2026-09-14
Add ldsc: a standalone LD Score regression implementation

BenjaminDEMAILLE:
Implements LD Score regression as a separate binary, per the same argument as mtag (#379) rather than as a plink2 flag: it is a method that operates on summary statistics rather than on the genotype data plink2 has open. It links the same `include/` libraries plink2 does and builds the same way; nothing in plink2 changes. The whole diff is one new directory.

`--ld-score` (#367) computes the LD Scores; this does the regression on top of them, and the conversion step that comes before.

## Why

The reference implementation (bulik/ldsc) requires **Python 2.7**, end of life since January 2020, and loads everything through pandas. It is still the standard way to get these numbers, so people are running an EOL interpreter to do it. `mtag` already had to grow an internal copy of the univariate and bivariate regressions to get Sigma; this is the full thing, usable on its own.

## What it does

```
ldsc --munge raw_gwas.txt.gz --merge-alleles w_hm3.snplist --out trait
ldsc --h2 trait.sumstats --ref-ld-chr eur_w_ld_chr/ --w-ld-chr eur_w_ld_chr/ --out trait_h2
ldsc --rg t1.sumstats,t2.sumstats --ref-ld-chr eur_w_ld_chr/ --w-ld-chr eur_w_ld_chr/ --out rg
ldsc --h2 trait.sumstats --ref-ld baseline --w-ld weights --overlap-annot --frqfile-chr 1000G. --out part
```

- **`--munge`**: the conversion and quality control `munge_sumstats.py` does. Columns are detected case-insensitively with the same aliases; filters in the same order (missing values, the `--merge-alleles` variant list, INFO, minor allele frequency, p outside (0, 1], variants that are not strand-unambiguous SNPs, duplicated IDs, sample size below the 90th percentile over 1.5); the p-value and the signed statistic become a signed Z, and the signed column's median is checked against its no-effect value. Also `--N`/`--N-cas`/`--N-con`, the column-naming flags, `--signed-sumstats <column>,<null>`, `--ignore`, `--info-list`, `--info-min`/`--maf-min`/`--n-min`/`--nstudy-min`, `--a1-inc`, `--no-alleles`, `--keep-maf`, and `--daner`.
- **`--h2`**: SNP-heritability (Bulik-Sullivan et al., *Nat Genet* 47:291-295, 2015), with the free or constrained intercept, the two-step estimator, `--chisq-max`, `--n-blocks`, and `--samp-prev`/`--pop-prev` for the liability scale.
- **`--rg`**: genetic correlation (*Nat Genet* 47:1236-1241, 2015) against one or more other traits, with `--intercept-h2`/`--intercept-gencov`/`--no-intercept`.
- **Partitioned LD Scores**: any LD Score file with several columns, or a comma-separated list of filesets, gives a stratified regression; `--overlap-annot` with the `.annot` files corrects the proportions and enrichments for annotations that overlap, and adds the enrichment standard error and p-value.

Input handling worth flagging: `--ref-ld`/`--w-ld` read both plink2's `--ld-score` output and ldsc's own `<prefix>.l2.ldscore[.gz]`, `--ref-ld-chr` takes the `@` convention, M comes from the `.l2.M_5_50` files (or `--M`), and the LD Scores are sorted by position and deduplicated before merging, since that order is what the jackknife blocks are cut on. Estimates also go to `<prefix>.h2`, `<prefix>.rg` or `<prefix>.results` at full precision, which the reference implementation only prints rounded into its log.

## Fidelity

The point was to reproduce the reference implementation's numbers, not merely its method: the same regression weights, two iterations of IRWLS, the same two-step estimator, the same block jackknife, the same merge order.

Verified against the reference implementation itself, driven as an oracle, on the same inputs:

- `--h2` with a free intercept and the two-step estimator; with the intercept constrained to 1 and to 1.06; with a custom two-step cutoff; with `--chisq-max` and a non-default block count; and the liability-scale conversion. Agreement to 9 significant digits, which is all it prints.
- `--rg` with free intercepts, under `--no-intercept`, and with a constrained genetic covariance intercept alongside two-step heritabilities. Same agreement, for both heritabilities, the genetic covariance and its intercept, and rg with its standard error, z and p.
- The partitioned regression on a three-annotation fixture, and `--overlap-annot` on an overlapping one: agreement to 12 significant digits for the total, the intercept, and every per-annotation coefficient, standard error, proportion and enrichment; to 7 digits for the enrichment p-value, where the two use different incomplete-beta routines.
- `--munge` on five configurations (a Beta/SE file with nonstandard column names and deliberate defects, case and control count columns with an odds ratio, `--merge-alleles`, `--daner`, `--a1-inc`): **byte-identical** output, including the reference's number formatting.

Three places where this deliberately differs, all documented in the code:

1. Where the reference errors out on a constrained intercept combined with the two-step estimator, this logs that the two-step is inapplicable and runs the constrained regression. That combination is reachable from the reference's own defaults, since it sets `--two-step 30` whenever `--intercept-h2` is absent, so `--intercept-gencov` alone fails there.
2. For a `--merge-alleles` variant the input does not supply, the reference writes empty fields; this writes `NA`, which is what its own readers treat as missing.
3. The reference pairs `.annot` rows with `.frq` rows by position, assuming the files are aligned; this matches on the variant ID, which agrees wherever that assumption holds.

## Testing

`2.0/utils/ldsc/tests/run_tests.sh <binary dir>` — 14 tests, about a second. It checks against `tests/oracle.py` and `tests/munge_oracle.py`, independent implementations of the same estimators and the same quality control, sharing no code with the C++ and deliberately numpy-free since the test runners have no numpy. It also requires the estimates to land within 4 SE of the parameters the fixtures were simulated under, the category proportions to sum to 1, `--overlap-annot` to change the answer rather than pass through, and a munged file to feed straight back into `--h2`, so an implementation that merely agrees with itself cannot pass.

Clean under `-fsanitize=address,undefined` on every path, and warning-free with the Makefile's `-Wall -Wextra -Wshadow`.

## Not implemented

- `--h2-cts` (cell-type-specific).
- `--print-delete-vals` / `--print-cov`.
- Computing the LD Scores themselves, which is what `--ld-score` is for.

Partitioned output assumes non-overlapping annotations unless `--overlap-annot` is given, and says so in the `.results` file.

## Notes

`include/plink2_stats.cc` (with `plink2_float` and `plink2_highprec`) is compiled in for the t-distribution tail the enrichment test needs, and the Makefile moves to `-std=c++17` for it, matching plink2's own.

One thing found along the way and worth knowing: `ScanadvDouble` is documented as approximate in the last couple of bits, so `--munge` parses with `strtod` instead. Any parsed value that gets written back out as text at fixed decimals needs that; the regression readers still use `ScanadvDouble`, where the difference cannot reach the estimates.

This is easy to decline; nothing in plink2 changes.


chrchang:
I will try to integrate https://github.com/kolemannix/ffc.h tomorrow for a fast accuracy-preserving alternative to strtod.

PR #481, opened 2026-09-14
Add --ld-score-annot: partitioned LD Scores, and the variant counts

BenjaminDEMAILLE:
`--ld-score` (#367) computes each variant's LD Score as the sum of unbiased r^2 over a window. Stratified LD Score regression needs that sum taken once per annotation, weighting each window variant by its value there, which is what ldsc's `--l2 --annot` does and what plink2 had no way to produce; without it the partitioned regression can only run on ldsc's shipped reference files.

```
plink2 --pfile data --ld-score --ld-score-annot baseline.annot \
       --ld-score-window-cm 1 --out scores
```

The annotation file is the `.annot` layout: a variant ID column, the position columns if present, and one column per annotation. Every variant being scored needs a row, since a missing one has no defined contribution; values are usually 0 or 1, but any finite weight works. One L2 column is written per annotation, named after it with an `L2` suffix, which is what reads these files keys on.

The counts behind the scores go to `<prefix>.ldscore.M` (all variants) and `<prefix>.ldscore.M_5_50` (minor allele frequency above 0.05), one value per annotation. LD Score regression needs them to turn a coefficient into a heritability, and they cost nothing here: the allele frequency falls out of the aggregates the r^2 path already computes as each variant streams past.

## Implementation

- **The unpartitioned path is untouched.** With no annotation file the inner loop keeps its scalar accumulator, and the output is byte-identical to master's on the same data. A single all-ones annotation reproduces it exactly as well, which the test checks.
- Annotations are held as floats, one row per raw variant, and the inner loop skips zero weights, which is most of them for the sparse annotations people use. For the baseline model's scale (about 100 annotations over 1.2M variants) that is around 0.5 GB of the workspace.
- The annotation count is capped at 512.
- `LdScoreInfo` gains an owned filename, so it now has a `CleanupLdScore()`, called next to `CleanupVcor()`.
- Reading the file needs the variant-ID hash table, so `LdScore()` takes `max_variant_id_slen` now.

## Verification

`TEST_LD_SCORE_ANNOT` compares against a brute-force computation from the `--export A` genotype matrix, which shares no code with plink2's r^2 path and applies the same unbiased estimator by hand. It agrees to plink2's printed precision: worst relative difference 4.9e-6 over 600 values. The test also checks

- the `.M` and `.M_5_50` counts against allele frequencies computed from the same export,
- that an all-ones annotation reproduces the unpartitioned score byte for byte, and that its `.M` is the variant count,
- that a narrower window covers a subset of the variants (compared through `nobsi`; the scores themselves are not monotone in the window, since the bias correction makes some contributions negative),
- that an incomplete annotation file, or one with no ID column, is rejected rather than silently producing a partial answer.

The whole `2.0/Tests` suite passes, and `TEST_LD_SCORE_ANNOT` is clean under `-fsanitize=address,undefined`.

## Why here rather than in the ldsc utility

The regression side of LD Score regression is #479, a standalone binary in the mould of mtag, because it works on summary statistics. This half is the opposite case: it needs the genotype data plink2 already has open, and the windowing and r^2 machinery already exist in `plink2_ld.cc`. With the two together, the stratified pipeline runs without ldsc's Python or its precomputed files.


chrchang:
Hmm, I get "Error: --ld-score-annot cannot be used with duplicate variant IDs." when trying to execute the new test on my Mac.

chrchang:
Uninitialized-variable bug addressed by 7184e87 .
- It's reasonable to modify PopulateIdHtableMt() in a future commit to remove this footgun.
- In the meantime, when duplicate IDs are not allowed in the hash table at all, AllocAndPopulateNondupHtableMt() should be used.  (This should be distinguished from the case where duplicate IDs are allowed as long as we never look them up, which is handled by AllocAndPopulateHtableMt() with htable_dup_base_ptr=nullptr.)

BenjaminDEMAILLE:
Thanks. I audited my open branches for the same pattern; one was a real bug.

`--lgen` (#357) built its .map table with `PopulateIdHtableMt()` and looked it up with `VariantIdDupflagHtableFind()`, which signals a duplicate by setting bit 31 of the returned index. Neither lookup site checked that bit, so a .map with two identical variant IDs sent a ~2^31 index into `allele_codes[2 * variant_idx]`, in the main .lgen loop and in the `--reference` pass. It now uses `AllocAndPopulateNondupHtableMt()` and errors out, with a message mirroring `--ld-score-annot`'s, since every .lgen record has to resolve to exactly one variant. Pushed to that branch with a duplicate-ID fixture in TEST_LGEN.

The other three new call sites are the duplicate-tolerant case rather than the footgun: `--set`/`--make-set` (#475), `--epistasis-boost` set mode (#476) and `--vc-test` (#401) all pass a `htable_dup_base` and look up with `VariantIdDupHtableFind()`, which walks the whole duplicate list, which is what PLINK 1.x's set loaders do with a repeated ID.

Separately, all of my open PRs are rebased on 7184e876 as of today, so they should be reviewable again.

PR #482, opened 2026-09-15
Add a NO_UNALIGNED sanitizer configuration, and fix the two --pmerge alignment bugs it finds

BenjaminDEMAILLE:
Follow-up to #445, where the suggestion was to try `-DNO_UNALIGNED` as a second sanitizer configuration.

To answer the question asked there first: no, `-DNO_UNALIGNED` had not been run with `-fno-sanitize=alignment` removed. `build_san/Makefile` passes that flag unconditionally, so adding `-DNO_UNALIGNED` on its own compiles the aligned-access paths and then switches off the one check that would say anything about them.

### The configuration

`make NO_UNALIGNED=1` compiles `-DNO_UNALIGNED` and leaves the alignment check on; the default build is untouched and keeps `-fno-sanitize=alignment`, since its code paths use unaligned vector loads on purpose. As noted in #445, this is a second configuration rather than a replacement: `#ifndef NO_UNALIGNED` is what production binaries run, so it still needs its own coverage.

### What it found

Two pre-existing bugs, both in `--pmerge`'s same-position scan arenas, both hit by `TEST_PMERGE`'s `--merge-mode nm-match` case:

```
../plink2_merge.cc:2746:21: runtime error: member access within misaligned address
  0x10808012578e for type 'RescanOnePosRecord', which requires 4 byte alignment

../plink2_merge.cc:6896:53: runtime error: member access within misaligned address
  0x10800028752a for type 'SamePosPvarRecord', which requires 8 byte alignment
```

Both arenas pack variable-length records back to back, each starting immediately after the previous record's last null terminator, so every record after the first lands wherever the strings happened to end. Writing `rec_blen` or `secondary_key` through such a pointer is undefined behavior, and would fault on a target that needs aligned loads (which is exactly the target `NO_UNALIGNED` exists for).

* `RescanOnePosRecord`: round the record length up to `alignof(RescanOnePosRecord)`. Nothing reads the padding, since `rec_blen` is only the stride to the next record, and the NOMEM check already runs on the rounded-up value.
* `SamePosPvarRecord`: round the write iterator up before each record is placed, and give the buffer `alignof(SamePosPvarRecord)` slack bytes per record instead of 1 to pay for it.

### Testing

The full 2.0 suite passes under `make NO_UNALIGNED=1` and under the default sanitizer build (`UBSAN_OPTIONS=print_stacktrace=1:halt_on_error=1`), on macOS/arm64. Without the second commit, both configurations abort on `TEST_PMERGE`; with it, neither reports anything.


chrchang:
I will replace the uses of alignof() to make this work under more compilation settings.

PR #483, opened 2026-09-15
Add --neighbour: the PC-space outlier statistic, reusing the retired name

BenjaminDEMAILLE:
Follows the guidance in #455: PLINK 1.x's IBS-based `--neighbour` should not survive as-is, and the name is worth reusing for the PCA-based calculation in section 3.4 of [Privé et al. (2020)](https://academic.oup.com/bioinformatics/article/36/16/4449/5838185).

### The statistic

A simplified local outlier factor over the PC scores. For each sample the report gives:

| Column | Meaning |
| --- | --- |
| `DIST_SELF` | mean Euclidean distance to its K nearest neighbours in PC space |
| `DIST_NN` | mean of those neighbours' own `DIST_SELF` values |
| `STAT` | `sqrt(DIST_SELF / DIST_NN)`, or `NA` when `DIST_NN` is zero |

The square root is from the paper, and is what stops a sample sitting beside a dense cluster from scoring like one alone in empty space. K defaults to 5, matching the `prob_dist()` default the paper's own code uses. All of the available PCs are used.

The one deliberate simplification: no robust Mahalanobis option. `bigutilsr`'s `prob_dist()` has a `robMaha` switch, and it can be added later if anyone wants it.

### Where the PCs come from

Either source:

```
plink2 --bfile x --pca 10 --neighbour
plink2 --bfile x --read-eigvec earlier.eigenvec --neighbour 12
```

With `--pca`, the calculation runs inside `CalcPca()` while `eigvecs_smaj` still holds exactly the scores just written to `.eigenvec`.

`--read-eigvec` loads coordinates from an earlier run, so a different K does not require recomputing the GRM. It accepts both PLINK 2.0 `.eigenvec` files (with a `#FID`/`#IID` header line) and PLINK 1.9's (headerless, FID then IID), and treats every column past the ID columns as a principal component. Samples in the file that are not in the current fileset are ignored, so one `.eigenvec` can be replayed against a `--keep` subset; a sample with no entry is an error, as are a duplicate entry, a ragged line and a non-numeric value. The two sources are mutually exclusive, and `--read-eigvec` without `--neighbour` is rejected. Coordinates come back with the precision they were printed at, so a replayed report can differ from an in-memory run in the last digit or two.

The "needs `--pca` or `--read-eigvec`" check lives after the parse loop rather than inside it, since `--neighbour` is parsed before both.

### No threshold

The paper picks a cutoff by inspecting a histogram of the statistic, which is not something to guess at for the user, so the report stops at the numbers. `--remove` does the rest. That also keeps this a report, like 1.x's `--neighbour` was.

### Cost

The nearest-neighbour search runs over a k-d tree, as nabor/libnabo does for the R-side implementations: median split on the widest-spread dimension, leaves of 16 scanned directly, three-way partitioning in the median selection so duplicate samples cannot make the build quadratic. The query carries the usual incremental bound (the per-dimension offsets from the query point to the current cell, not just the distance to the current splitting plane), and a leaf scan stops accumulating a candidate's distance as soon as it passes the current K-th best. Build is single-threaded; queries are threaded over samples, with per-thread scratch taken from bigstack.

Against the direct scan, on dummy filesets with synthetic PC scores, K=5, 16 threads, min of two runs, both binaries reading the same `.eigenvec` so only the search is timed:

| samples | PCs | direct scan | k-d tree |
|--------:|----:|------------:|---------:|
| 100k | 2 | 1.60 s | 0.05 s |
| 100k | 5 | 1.63 s | 0.11 s |
| 100k | 10 | 2.05 s | 0.87 s |
| 100k | 20 | 3.83 s | 4.43 s |
| 100k | 40 | 7.73 s | 8.39 s |
| 500k | 10 | 80.6 s | 16.2 s |
| 1M | 10 | 457 s | 49.6 s |

The reports are byte-identical in every row. Past roughly 20 PCs a k-d tree stops pruning usefully, which is a property of the method rather than of this implementation; the partial-distance cutoff is what holds that case to within 15% of a straight scan instead of several times worse.

Neighbour sets are tie-broken by sample index, so the report does not depend on the order the tree visits points in, and a fileset with duplicate samples gives the same answer a direct scan does.

### Testing

`TEST_NEIGHBOUR` checks the report against a direct recomputation from the `.eigenvec` it came from, at two values of K and two PC counts; that the four planted outliers rank top-4; `cols=`, `zs`, and the `--neighbor` spelling; the `--read-eigvec` round trip at two values of K, a 1.9-format file, and a `--keep` subset; and seven error paths (no PC source, both sources, `--read-eigvec` without `--neighbour`, K ≥ sample count, a missing sample, a duplicate sample, a ragged or non-numeric file). The outliers each get their own block of homozygous variants rather than sharing one: a tight group of identical samples reads as a small cluster, not as outliers, which is what a *local* factor is for. The oracle is pure Python, since the CI runners have no numpy.

The k-d tree was also checked against a NumPy brute-force reference over 14 configurations (up to 2000 samples; 1 to 20 PCs; K from 1 to sample_ct - 1; duplicated points, wholly identical points, constant dimensions, coordinate scales spanning 1e6 to 1e-6, pre-sorted input); every `DIST_SELF` and `DIST_NN` agrees to the six significant digits the report prints.

Full 2.0 suite passes on a clean build, and `TEST_NEIGHBOUR` is clean under `build_san` (ASan + UBSan).


chrchang:
- This should optionally take preexisting --pca output (ideally from either PLINK 1.9 or 2.0).
- ...at which point it should be clear that we want to do better than O(sample_ct^2 * pc_ct).  Doesn't need to be in this PR, but if not, the PR should be written with the assumption that the brute-force algorithm will be replaced immediately.

chrchang:
Roughly speaking, there's almost no point in implementing this at all if the implementation isn't at least as efficient as the R nabor package utilized by bigutilsr.

BenjaminDEMAILLE:
Both points are addressed, and the branch is rebased on current master.

### The search runs over a k-d tree now

`CalcNeighbour()` builds one tree over the PC scores and queries it from the worker threads: median split on the widest-spread dimension, leaves of 16 scanned directly, three-way partitioning in the median selection so duplicate samples (or a dimension that is constant over part of the data) cannot make the build quadratic. The query carries the standard incremental bound, the per-dimension offsets from the query point to the current cell rather than just the distance to the current splitting plane, and a leaf scan stops accumulating a candidate's distance as soon as it passes the current K-th best.

Timings against the direct scan the first version used, on dummy filesets with synthetic PC scores (5 clusters plus noise), K=5, 16 threads, min of two runs, both binaries reading the same .eigenvec so only the search is being timed:

| samples | PCs | direct scan | k-d tree |
|--------:|----:|------------:|---------:|
| 100k | 2 | 1.60 s | 0.05 s |
| 100k | 5 | 1.63 s | 0.11 s |
| 100k | 10 | 2.05 s | 0.87 s |
| 100k | 20 | 3.83 s | 4.43 s |
| 100k | 40 | 7.73 s | 8.39 s |
| 500k | 10 | 80.6 s | 16.2 s |
| 1M | 10 | 457 s | 49.6 s |

Every row's two reports are byte-identical.

Two things worth flagging. First, past roughly 20 PCs a k-d tree stops pruning usefully, which is a property of the method rather than of this implementation; the partial-distance cutoff is what keeps that case within 15% of a straight scan instead of several times worse, so there is no PC count at which this is a meaningful regression. Second, neighbour sets are tie-broken by sample index, so the report does not depend on the order the tree visits points in, and a fileset with duplicate samples gives the same answer the scan did.

### --read-eigvec

`--neighbour` now takes its coordinates from either source:

```
plink2 --bfile x --pca 10 --neighbour
plink2 --bfile x --read-eigvec earlier.eigenvec --neighbour 12
```

`--read-eigvec` accepts both PLINK 2.0 `.eigenvec` files (with a `#FID`/`#IID` header line) and PLINK 1.9's (headerless, FID then IID), and treats every column past the ID columns as a principal component. Samples in the file that are not in the current fileset are ignored, so one `.eigenvec` can be replayed against a `--keep` subset; a sample with no entry is an error, as is a duplicate entry, a ragged line, or a non-numeric value. The two sources are mutually exclusive, and `--read-eigvec` without `--neighbour` is rejected.

The coordinates come back with the precision they were printed at, so a replayed report differs from an in-memory run in the last digit or two; the test compares against a recomputation from the same text rather than byte for byte.

### Testing

`2.0/Tests/TEST_NEIGHBOUR` gains the `--read-eigvec` cases: the round trip at two different K, the 1.9-format file, the `--keep` subset, and the five error paths above.

Separately, against a NumPy brute-force reference over 14 configurations (up to 2000 samples; 1 to 20 PCs; K from 1 to sample_ct - 1; duplicated points, wholly identical points, constant dimensions, coordinate scales spanning 1e6 to 1e-6, pre-sorted input), every `DIST_SELF` and `DIST_NN` agrees to within the six significant digits the report prints.

PR #488, opened 2026-09-18
--glm: take missing calls out of the precomputed covariate dot products

BenjaminDEMAILLE:
Linear `--glm` has a fast path that reuses the covariate dot products computed once for the whole sample and only fills in the genotype row per variant. A variant with even one missing call was excluded from it, so the generic path copied the covariates for the nonmissing samples and called `MultiplySelfTranspose()` on the copy, at O(sample_ct * predictor_ct^2) per variant. On genotyped data nearly every variant has a missing call, so that was the usual path.

The nonmissing-sample dot products are the full-sample ones minus the missing samples' contributions, so the fast path now takes those variants too: it subtracts the missing samples from copies of `xtx_image` and `xt_y_image`, inverts the intercept-and-covariate block per variant, and lets the genotype row skip missing calls the way it already skips zeroes. That is O(missing_ct * predictor_ct^2 + predictor_ct^3).

### Speed

20000 samples, 50000 variants, 10 covariates, 1% of calls missing, M4 Max, best of three:

| | before | after |
|---|---|---|
| 1 thread | 19.4s | 1.6s |
| 16 threads | 1.66s | 0.26s |

The same data with no missing calls is unchanged (0.88s and 0.20s before and after), as is a no-covariate run. The gain grows with sample_ct, since the per-variant term it removes is the one proportional to it.

### Output

Unchanged, byte for byte, on that 50000-variant run and on a 30-command differential against the current build: every genotype coding, `hide-covar` on and off, `--tests`, `--parameters`, `interaction`, `qt-residualize`, `--covar-variance-standardize`, `--vif`/`--max-corr`, chrX/chrY/MT, `--xchr-model 1`, dosages, multiallelic variants, the no-covariate case, 1 versus 16 threads, and `zs`. stderr included.

### Where it does not apply

- More than half the calls missing: subtracting then costs more than copying the rest, and gives up precision.
- A covariate that is constant over the variant's nonmissing samples: its variance there is zero, and a subtraction leaves rounding noise instead, which the correlation check reads as a large correlation rather than an undefined one. Such variants go to the generic path, so their `VIF_INFINITE` rows are unchanged. The check for this is O(missing_ct * covar_ct).
- `--parameters`, since the precomputed matrices are laid out for the full predictor list. Happy to extend this later if you want it.
- chrX, dosages, multiallelic variants and interactions, as before.

### Tests

`2.0/Tests/TEST_GLM_MISSING` checks each variant's rows against an oracle run that extracts that variant alone and removes its missing-genotype samples up front, so the oracle regression sees no missing calls and takes the path this one is modelled on. It covers both linear drivers (a three-phenotype batch and a single phenotype), all six genotype codings, covariate rows as well as `hide-covar`, and the constant-covariate fallback. It takes about 3 seconds.

I also ran the suite under `build_san` (ASan + UBSan), and the whole of `2.0/Tests` after a clean build.

Two notes on the comparator, both in its header comment: it allows half of the last printed place on each side, since both sides are rounded; and it treats `VIF_INFINITE` and `RANK_DEFICIENT` as equivalent, because a genotype column with no variance (a recessive coding of a variant with no hom-ALT call) is named the first way by the path that handles missing calls and the second way by the one the oracle takes. Tell me if you would rather those agreed, and which one you prefer.

One more thing, separate from this PR: with no covariates, `--glm genotypic` (and `hethom`) rejects variants that have missing calls more often than it should. The genotype and domdev dot products are written to rows 1 and 2 of `xtx_inv`, but the missing-call branch calls `CheckMaxCorrAndVif()`, which reads the lower triangle, where the geno x domdev entry is still `xtx_image`'s zero. On 40 dummy variants, `--glm allow-no-covars genotypic` gives 49 error rows out of 120, against 9 for the oracle; mirroring those two rows into the lower triangle, which this PR already does for the covariate case, makes the two agree row for row. I left it out of here since it changes results, and will send it separately.


chrchang:
I will lower the missing-rate upper bound for this from 1/2 to 1/3 since, when the two approaches are close in speed, we should be biased against the subtract-missing approach due to its lower precision.  But otherwise this looks great.

